PR #15, opened 2020-09-24
Added WebRTC demo

blagojce95:
This PR contains a demo for WebRTC as a backend for p2p communication. It is only tested locally, we plan to test it using a google cloud instance as a signaling server. This solution should also be able to connect peers which are behind NATs.

### DIFF
--- a/WebRTC_backend/textchat/client.js
+++ b/WebRTC_backend/textchat/client.js
@@ -1,0 +1,210 @@
+//our username 
+var name; 
+var connectedUser; 
+
+//connecting to our signaling server 
+var conn = new WebSocket('ws://52.200.7.179:55535');
+
+console.log("Connecting to: " + "55535");
+
+conn.onopen = function () { 
+   console.log("Connected to the signaling server");
+};
+ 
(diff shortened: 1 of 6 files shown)


WebRTC_backend/textchat/client.js
```diff
@@ -0,0 +1,210 @@
+//our username 
+var name; 
+var connectedUser; 
+
+//connecting to our signaling server 
+var conn = new WebSocket('ws://52.200.7.179:55535');
```
mmilenkoski, line 6: We should change this line to `var conn = new WebSocket('ws://localhost:9090');` for the local version

martinjaggi:
thanks! as this one is 'pure' WebRTC, how does it relate to the peerjs ones (which also use WebRTC as backend) #16 and #17

martinjaggi:
we can close this and/or add it to the /experimental/  folder

mmilenkoski:
Closing this because it was moved to `/experiments` with 73b133cc0314c536c83d947674ec89b716baa1b7.

PR #16, opened 2020-09-27
Add peerjs backend prototype

mmilenkoski:
This is a simple example of p2p communication with [peerJS](https://github.com/peers/peerjs). The benefit of peerJS is that it is much simpler to work with than pure WebRTC and they provide a free hosted server for development. The only requirement Is to choose unique usernames so that they don't collide with others using the public server. Additionally, they provide the server code as another [library](https://github.com/peers/peerjs-server). In this way, we can host our own server without having to implement it from scratch, and also customize it to our needs. With the publically hosted server, we need a way to share the ids between users. However, if we host our own server, there is an option to get a list of all connected peers from the server which will eliminate the need to communicate the ids. 

To test the example, open the file in two tabs, choose unique usernames, register on both tabs, and then you can send messages from one to the other.

I also tested the example with a locally hosted server (the commented line in the code) and it works fine.

The next step is to try to host our own server on the cloud.

Let me know if you have any questions or suggestions.

### DIFF
--- a/peerjs-backend/README.md
+++ b/peerjs-backend/README.md
@@ -1,0 +1,53 @@
+To use the file `index.html`, open it in two browser tabs, register with different usernames and then you can send messages by typing the id of the receiver and the message 
+and clicking Send.
+
+
+There are several options for the backend communication. The simplest thing is to use the public server without a STUN/TURN server. To use this option, use the following line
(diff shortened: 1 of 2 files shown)


martinjaggi:
thanks! let's merge this already as discussed. can you add a readme?

PR #17, opened 2020-10-07
Distributed training (PyTorch) proof-of-concept with Flask and PeerJS communication

giorgiosav:
This builds on PR #16 with a very basic proof-of-concept of 2 peers training a model and communicating.

It works like this: a model is trained locally with Pytorch and periodically serves its parameters (as JSON) over localhost using a Flask app. The peer on the same host machine where the model is being trained makes periodic requests for the parameters and sends them to another peer using PeerJS.
The flow of data in one direction looks like this:
```
|Pytorch| ---> |localhost| ---> |local peer| ---> (internet) ---> |remote peer|
```

(Disclaimers) The code is not very elaborate/stable, it is just meant as a demonstration and I did not test it with a remote peer. The neural network that gets trained is also pretty basic and is artificially slowed down with `sleep` statements.

To test it follow these steps (the order doesn't really matter).
- Start the Flask web app. The model takes a few minutes to train, so you should execute the subsequent steps within this time in order to have some meaninfgul communication.
```bash
cd local_webapp
export FLASK_APP=webapp.py
flask run --port 5001
```
- Open `peerjs-backend/local_peerjs_interface.html` in 2 tabs and open the console in both tabs
- On the left tab, type in `left`, `right` for "Username" and "Receiver", then click Register
- On the right tab, type in `right` for "Username" (don't fill in the "Receiver" field) and click Register.

Now in the console of the right tab you should see that the data from the peer on the left tab is periodically received.

### DIFF
--- a/local_webapp/task.py
+++ b/local_webapp/task.py
@@ -1,0 +1,55 @@
+import torch
+import time
+
+def generate_data():
+    from random import uniform
+    from math import pi
+    N = 100
+    x = torch.Tensor([uniform(0, 1) for _ in range(N)])
+    y = torch.Tensor([uniform(0, 1) for _ in range(N)])
+
+    classes = ((x - 0.5) ** 2 + (y - 0.5) ** 2 < 0.5 / pi).long()
+    one_hot_classes = torch.zeros(classes.shape[0], 2).scatter_(1, classes.view(-1,1), 1.0)
+    return torch.stack([x, y]).T, one_hot_classes
+
+
+class Producer:
+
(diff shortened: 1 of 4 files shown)


mmilenkoski:
@martinjaggi should we remove this because it is quite slow/unstable? Alternatively, we could move it to `experiments`. What do you think?

martinjaggi:
yes let's move it to experiments

mmilenkoski:
Closing this because it was move to `experiments` with 65004cb

PR #19, opened 2020-11-20
Model/training sharing prototype

giorgiosav:
This adds some more functionality to the `tensorflow-peerjs` branch with the `PeerJS` and `ModelBuilder` classes. Specifically:
* `PeerJS` is a wrapper around the PeerJS library and allows to send/receive arbitrary data and handle it with any input function. The same module also defines "command codes" to determine what to do with the data (is it a description of the model? model weights? etc...)
* `ModelBuilder` is a basic tracing object that keeps track of how a model was built and is able to serialize/deserialize it. This way a model can be sent from one peer to another. It's not very advanced yet and there might be better approaches to implement this.

Finally there is a small demo, with `sender.html` and `receiver.html` that shows how a peer can send a model to another one. I haven't implemented weight injection in the demo yet but the infrastructure to do so is in place. 

To run this demo, you will need to set up a local PeerJS server running on `localhost:9000` (https://github.com/peers/peerjs-server):
```bash
npm install peer -g
peerjs --port 9000 --key peerjs --path /myapp
```
Once this is done, just open `sender.html` and `receiver.html`, register a username on both tabs and click "Send" on the sender tab. You will see all the action in the console.

### DIFF
--- a/peerjs-backend/index.html
+++ b/peerjs-backend/index.html
@@ -7,44 +7,91 @@
 
 <div>
 <input type="text" placeholder="Receiver" id="receiver">
-<input type="text" placeholder="Message" id="message">
 <button onclick="send()">Send</button>
 </div>
 
 <script src="https://unpkg.com/peerjs@1.3.1/dist/peerjs.min.js"></script>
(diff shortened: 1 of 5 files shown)


giorgiosav:
Closing in favor of PR #20

PR #24, opened 2020-12-18
simulator first version

ThomasDCepfl:
Dear all,
Here is a first version of the simulator I have been working on throughout my semester project at MLO.
I tried to make it as generic and easy-to-use as possible.
I split the different aspects into different packages, which all have their dedicated notebook, Python files and README. 
I am open to any suggestions or remarks.
Best,
Thomas


koloskova (changes requested):
The code looks good and implements many useful features, however, the code structure could be improved. There are many repetitions in the code which are good to be avoided.

simulator/A-NC/A_NC.py
```diff
@@ -0,0 +1,39 @@
+from utils import *
```
IamTao, line 1: it is better to explicitly list the functions/classes names.

simulator/A-NC/A_NC.py
```diff
@@ -0,0 +1,39 @@
...
+    """
+    Runs a decentralized optimization algorithme for the given learning rate for
+    a number of rounds, over some network. Outputs the accuracies and returns them.
+
+    Params:
+        train_loader (array): the list of all train datasets, one per client
+        test_loader (array): the list of test datasets, one per client
+        comm_matrix (numpy.array): the communication matric modeling the network
+        num_rounds (int): the number of data exchanges between nodes
+        epochs (int): the number of optimization steps between each communication (minimum 1)
```
IamTao, line 14: the `epoch` here is confusing. is it correspond to the number of local update steps?

simulator/A-NC/A_NC.py
```diff
@@ -0,0 +1,39 @@
...
+    accs = []
+    global_model, client_models = model_init(num_clients, net)
+    opt = optimizer_init(client_models, lr, optimizer)
+
+    loss, test_loss, acc = 0.0, 0.0, 0.0
+    for r in range(num_rounds):
+        loss = 0
+        for i in range(num_clients):
+            loss += client_update(client_models[i], opt[i], train_loader[i], epoch=epochs)
+        diffuse_params(client_models, comm_matrix)
```
IamTao, line 32: maybe `diffuse_params` could be a class that can provide different choices of diffusion, e.g. with compressed communication?

simulator/A-NC/README.md
```diff
@@ -0,0 +1,78 @@
+# A-NC
```
IamTao, line 1: it is a bit hard to infer the exact meaning of `A-NC`

simulator/A-NC/utils.py
```diff
@@ -0,0 +1,344 @@
...
+    client_model.train()
+    for e in range(epoch):
+        for batch_idx, (data, target) in enumerate(train_loader):
+            data, target = data.cuda(), target.cuda()
+            optimizer.zero_grad()
+            output = client_model(data)
+            loss = F.nll_loss(output, target)
+            loss.backward()
+            optimizer.step()
+    return loss.item()
```
IamTao, line 109: in addition to the loss, we may need to track other metrics, e.g. top-1 acc?
also, right now, it seems like the function will only return the loss value of the last iterate. I think it is not sufficient?>

simulator/failures/utils.py
```diff
@@ -0,0 +1,344 @@
...
+    and average the consensus values of each parameter vector.
+
+    Params:
+        global_model (nn.Module): the global neural network averaging all the clients
+        client_models (array of Net): the list of all the client neural networks
+
+    Returns:
+        cons(array): the list of consensus values, one per client
+    """
+    cos = nn.CosineSimilarity(dim=0, eps=1e-6)
```
IamTao, line 188: why we initialize the `cos` here?

simulator/failures/utils.py
```diff
@@ -0,0 +1,344 @@
...
+        traindata = datasets.MNIST('./data', train=True, download=True,
+                       transform=transforms.Compose([transforms.ToTensor(),transforms.Normalize((0.1307,), (0.3081,))])
+                       )
+        traindata_split = torch.utils.data.random_split(traindata, [int(traindata.data.shape[0] / num_clients) for _ in range(num_clients)])
+        train_loader = [torch.utils.data.DataLoader(x, batch_size=batch_size, shuffle=True) for x in traindata_split]
+
+        test_loader = torch.utils.data.DataLoader(
+                datasets.MNIST('./data', train=False, transform=transforms.Compose([transforms.ToTensor(),transforms.Normalize((0.1307,), (0.3081,))])
+                ), batch_size=batch_size, shuffle=True)
+    else: # distribution == 'non-iid'
```
IamTao, line 225: we may include more description details and other degrees of non-iid-ness for the `non-iid` data distribution.

simulator/A-NC/utils.py
```diff
@@ -0,0 +1,344 @@
+import numpy as np
+import networkx as nx
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+import torch.optim as optim
+from torchvision import datasets, transforms
+from tqdm import tqdm
+
+class Net(nn.Module):
```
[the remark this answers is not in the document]
IamTao: agreed

IamTao:
the simulator code looks good to me. 

my main concern is that, why not build a unified simulation code (i.e., one package) that can cover different cases, e.g., topologies, failures, latency?

PR #33, opened 2021-03-14
DeAI mobile UI

MarcelTorne:


martinjaggi:
for the dependency on tensorflow.js, maybe we can use git submodule, or via npm (not sure what's best practice?) to avoid the many files in our git?

MarcelTorne:
> for the dependency on tensorflow.js, maybe we can use git submodule, or via npm (not sure what's best practice?) to avoid the many files in our git?

Sorry, I didn't realise. I did it using npm, but I will verify to see how I could do it to not have this many files.

martinjaggi:
very nice! maybe we can merge this one already, and then make separate PRs for the image or .csv functionality, and to connect to the learning algorithm and task server?
a short readme on which `npm` commands are needed to start playing with the demo would be nice too

martinjaggi:
thanks guys! i'll merge this for now, then it's easier to make smaller changes & issues on top and involve more people

PR #36, opened 2021-03-23
Change architecture to vue app

MarcelTorne:


mobile-browser-based-version/mobile-app/src/components/TitanicComponents/titanic_training.vue
```diff
@@ -0,0 +1,1018 @@
...
+      return model;
+    },
+
+    /**
+     * Function called once data has been loaded and processed. Updates the model 
+     * stored in local memory
+     
+     * @returns Returns a tf.model for training
+     */
+    async train(Xtrain, ytrain) {
```
martinjaggi, line 681: towards separating UI & training algorithm: could you externalize this function train() function into its own .js file? it would need to get the model and UI pointer as well somehow, so that the callback can update the accuracy display? later we could then possibly give a flag to the trainer what kind of training we want (distributed or local training), but we can make a new PR for that later.

mobile-browser-based-version/mobile-app/src/components/TitanicComponents/titanic_training.vue
```diff
@@ -0,0 +1,1018 @@
...
+          }
+        };
+
+        reader.readAsText(file);
+      }
+    },
+
+    /**
+     * Creates the tf.model
+     * TO DO: fetch model from local storage
```
martinjaggi, line 655: clarify that this is architecture + model weights (initialized to zero for now. for future better warm-start, we can let the train loop take care of it).  load + save to local storage can be implemented later, no prob

mobile-browser-based-version/mobile-app/README.md
```diff
@@ -0,0 +1,66 @@
...
+    - **Description of the task** under the name `[taskName]_description.vue. It gives an overview of the task. 
+    - **Training of the task** under the name `[taskName]_training.vue. Allows the users to train the model. As a side note, components are created only when they are called by the user. Meaning that until the user reaches the training page of the task, the `[taskName]_training.vue`is not created. When the user reaches for the first time the page, the component is created, and only then the NN model is created and stored in the browser's indexdb database.
```
martinjaggi, line 52: clarify that the algorithmic parts are here in this file (for now). the training is mainly generic for all tasks, but the data pre-processing is very task-specific (also accuracy display)

martinjaggi (changes requested):

martinjaggi (approved):

PR #44, opened 2021-04-21
Refactor to simplify future image tasks creation

MarcelTorne:
In this PR I am refactoring the MNIST component into a similar manner as the Titanic component in order to simplify future image tasks creation.


mobile-browser-based-version/src/components/MNISTComponents/MNIST_script.js
```diff
@@ -0,0 +1,146 @@
...
+    // {String} description of the datapoint given as example
+    dataExampleText: "Below you can find an example of an expected image representing the digit 9.",
+};
+
+/**
+ * Object used to contain the model's training specifications
+ */
+export const training_information = {
+    // {String} model's identification name
+    model_id: "mnist-model",
```
PaulMansat, line 29: We can adopt the following rule: before handling a routing path, we lower case the task's name. This would mean that all paths in the router should be lowered case. 
This will help us modularize without errors the routing of tasks in each component.

mobile-browser-based-version/src/components/MNISTComponents/MNIST_training.vue
```diff
@@ -380,95 +322,149 @@
...
-      DataExampleText:
-        "Verum ad istam omnem orationem brevis est defensio. Nam quoad aetas M. Caeli dare potuit isti suspicioni locum, fuit primum ipsius pudore, deinde etiam patris diligentia disciplinaque munita. Qui ut huic virilem togam deditšnihil dicam hoc loco de me; tantum sit, quantum vos existimatis; hoc dicam, hunc a patre continuo ad me esse deductum; nemo hunc M. Caelium in illo aetatis flore vidit nisi aut cum patre aut mecum aut in M. Crassi castissima domo, cum artibus honestissimis erudiretur.",
+      // Variables for general informations
+      model_name: "",
+      DataFormatInfoText: "",
+      DataExampleText: "",
+      DataExample: null,
+      // Headers related to training task of containing item of the form {id: "", userHeader: ""}
+      headers: [],
```
PaulMansat, line 372: I don't think the variable "headers" is required for image tasks

mobile-browser-based-version/src/components/MNISTComponents/MNIST_script.js
```diff
@@ -0,0 +1,146 @@
...
+ 
+        Xtrain = tf.concat(image_tensors, 0)
+        // object to return 
+    } else {
+        console.log("Cannot start training.")
+    }
+    return {accepted: startTraining, Xtrain: Xtrain, ytrain: ytrain}
+}
+
+async function check_data(training_data){
```
PaulMansat, line 175: Maybe we can put this function in a helper script for all images script (as it's not task-specific). In the future, we could even add more methods that are image-specific in this script.

mobile-browser-based-version/src/components/MNISTComponents/MNIST_desc.vue
```diff
@@ -128,18 +128,24 @@
...
-      LimitationsText: "Hae duae provinciae bello quondam piratico catervis mixtae praedonum a Servilio pro consule missae sub iugum factae sunt vectigales. et hae quidem regiones velut in prominenti terrarum lingua positae ob orbe eoo monte Amano disparantur.",
-      TradeOffsText: "Per hoc minui studium suum existimans Paulus, ut erat in conplicandis negotiis artifex dirus, unde ei Catenae inditum est cognomentum, vicarium ipsum eos quibus praeerat adhuc defensantem ad sortem periculorum communium traxit. et instabat ut eum quoque cum tribunis et aliis pluribus ad comitatum imperatoris vinctum perduceret: quo percitus ille exitio urgente abrupto ferro eundem adoritur Paulum. et quia languente dextera, letaliter ferire non potuit, iam districtum mucronem in proprium latus inpegit. hocque deformi genere mortis excessit e vita iustissimus rector ausus miserabiles casus levare multorum.",
+      OverviewText: "",
```
PaulMansat, line 135: Yes, it's a better variable name!

PaulMansat:
Thanks for the changes Marcel!

mobile-browser-based-version/src/components/MNISTComponents/MNIST_script.js
MarcelTorne: That seems like a good idea to me. But it is what we are already doing right?

mobile-browser-based-version/src/components/MNISTComponents/MNIST_script.js
```diff
@@ -48,6 +47,86 @@ export const training_information = {
     },
 }
 
+ /**
+  * fetch model from local storage or indexdb
+  * 
+  * @returns Returns a tf.model or null if there is no model
+  */
+export async function get_model_from_storage(){
```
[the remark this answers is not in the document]
MarcelTorne: Yeah definitely, I added both since I am unsure on wether we already changed to indexedDB for all of the tasks

PR #48, opened 2021-05-16
Modularization

PaulMansat:
- introducing working agents to improve modularisation: training informant, training manager, model memory manager, and communication manager. 
- introducing modular routing: improve the routing procedure by removing the need to create a new route for each new task. Take advantage of the modularisation of Vue.js routers 
- Correction of minor graphical issues: (1) the sidebar was automatically open for small screens, it is now automatically closed, (2) fix some spacing issue between components, (3) change accuracy chart to not automatically scale 
- outsource the main building blocks of each component (such as training information, description …) 
- correction of chart id error: all charts were saved under the same name, now each component associated with a task has a unique ID that is associated with all of the tasks’ component
- re-organization of the folder hierarchy 
- Further improvement for the future: (1) modularise the testing component of images (can be done with tasting MNIST component to modularise); (2) Have a uniform file access frame, now the file access from CSV files and image file is different, have to be merged into one single file.


martinjaggi (approved):

mobile-browser-based-version/README.md
```diff
@@ -1,16 +1,21 @@
...
+(Official [instructions](https://github.com/peers/peerjs-server#run-server)).
 
 ## How to run the app
-All commands are ran from the present folder here. 
+All commands are ran from the folder ./mobile-browser-based-version.  
 ### NPM installation
 NPM is a package manager for the JavaScript runtime environment Node.js.  
-To start the application (running locally) run the following command. 
+To start the application (running locally) run the following command.    
+Note: the application is currently developed using [NPM 7.6.3](https://www.npmjs.com/package/npm/v/7.6.3).
```
mmilenkoski, line 14: could you add the Node version as well?

mobile-browser-based-version/src/Task Definition/mnist.js
```diff
@@ -0,0 +1,229 @@
+import * as tf from "@tensorflow/tfjs";
```
mmilenkoski, line 1: I believe it would be best if we try to avoid spaces in file and folder names. For example, it would be better if we change 'Task Definition' to 'task_definition'. It can cause some problems between operating systems, especially on Windows

s314cy:
We should stick to a single naming convention, i.e. either camel or snake case (this includes folder and file naming).

PaulMansat:
> We should stick to a single naming convention, i.e. either camel or snake case (this includes folder and file naming).

So which should we use? Snake case?

mmilenkoski:
> We should stick to a single naming convention, i.e. either camel or snake case (this includes folder and file naming).

I agree. But we do not have to use the same naming convention for file names and variables. For file names, I think the best would be to use a snake case. For variables, as you prefer, but should be consistent in the code.

PaulMansat:
> > We should stick to a single naming convention, i.e. either camel or snake case (this includes folder and file naming).
> 
> I agree. But we do not have to use the same naming convention for file names and variables. For file names, I think the best would be to use a snake case. For variables, as you prefer, but should be consistent in the code.

Right, let's choose snake for all variables name except the component name?
I'll go through each file and change that.

PaulMansat:
BTW, Hugo, could you put the receiver interface code in the communication_manager script, and I think the rendering code should be but in the MainTrainingFrame. Thanks!

MarcelTorne (approved):
Looks really good to me! Thanks for this awesome work Paul! This week I will use as example your work and bring it to the LUS-covid component and to add the testing images component!

PR #53, opened 2021-05-18
Modularise and improve testing component for images

MarcelTorne:
- Make a general modularised component for testing the model


mmilenkoski:
I think it would be best if you split this into 2 or 3 PRs, one for each point in the description.

mobile-browser-based-version/src/components/Building Frames/ImageTestingFrame.vue
```diff
@@ -0,0 +1,380 @@
...
+          </div>
+
+          <div class="relative p-4 overflow-x-scroll">
+            <span class="text-sm text-gray-500 dark:text-light">{{
+              DataExampleText
+            }}</span>
+          </div>
+
+          <!-- Data Point Example -->
+          <img src="../../../example_training_data/9-mnist-example.png" alt=""> <!-- TODO fix this-->
```
PaulMansat, line 78: Maybe we should modularize this

mobile-browser-based-version/src/components/Building Frames/ImageTestingFrame.vue
```diff
@@ -0,0 +1,380 @@
...
+  }, 
+  data() {
+    return {
+      title: "mnist-testing",
+      DataFormatInfoText:"",
+      DataExampleText: "",
+      // Different Task Labels
+      task_labels: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9],
+      IMAGE_HEIGHT: 28,
+      IMAGE_WIDTH: 28,
```
PaulMansat, line 271: Initialization from this.Task.training_information.IMAGE_HEIGHT, or something like that. 
I'm quite sure that once data() is initialized, the props have already been defined and thus can be accessed.

mobile-browser-based-version/src/router/index.js
```diff
@@ -69,6 +70,12 @@ const routes = [
         component: MainTrainingFrame, 
         props: dynamicTaskAllocationFn,
       },
+      {
+        path: 'testing',
+        name: 'testing',
+        component: MainTestingFrame,
```
PaulMansat, line 76: Great.

mobile-browser-based-version/src/components/Main Frames/MainTestingFrame.vue
```diff
@@ -0,0 +1,24 @@
+<template>
+    <!-- CSV tasks -->
+    <!--TODO-->
+
+    <!-- image tasks -->
+    <div v-if='Task.training_information.data_type == "image"'>
```
PaulMansat, line 6: Nice separation of CSV and images, like for training.

PaulMansat (approved):

mobile-browser-based-version/src/Task Definition/mnist.js
```diff
@@ -12,30 +12,16 @@ export class MnistTask {
      * 
      * @returns Returns a tf.model or null if there is no model
      */
-    async get_model_from_storage() {
-        let model = await tf.loadLayersModel('indexeddb:://working_'.concat(training_information.model_id))
+     async get_model_from_storage() {
+        let model = await tf.loadLayersModel(training_information.save_path_db)
```
mmilenkoski, line 16: Shouldn't this be `this.training_information` or am I missing something?

mobile-browser-based-version/src/Task Definition/mnist.js
```diff
@@ -84,6 +70,8 @@ export class MnistTask {
         // values sum to 1.
         model.add(tf.layers.dense({ units: 10, activation: 'softmax' }));
 
+        model.save(training_information.save_path_db);
```
mmilenkoski, line 73: same comment as above

mobile-browser-based-version/src/Task Definition/mnist.js
MarcelTorne: Both ways are right, since the `training_information`object is outside of the class. But then when we initialise the class we also create a `this.training_information`object out of the external one. Do you prefer that I use `this`?

mmilenkoski: I think it's best if you are consistent and use `this` inside of the class, so if we refactor it in the future, the class will be independent of the outside context. Thanks!

mmilenkoski (approved):

mobile-browser-based-version/src/Task Definition/mnist.js
MarcelTorne: Sounds good! It's changed to always using `this`inside the class

PR #61, opened 2021-05-26
Helper server as task provider

s314cy:
The helper server is now an ExpressJS application, serving peers with the usual information (i.e. the list of connected peers). This PR brings centralization of the tasks' description and training information, as well as the peer ID generation. Models are naively available for download. The next step is integrating all that to the client's code (src/task_definition/ folder).


mmilenkoski:
Hey @s314cy, thank you for your work!

Your PR is introducing many changes to the front-end code that I can see you used to test your code locally. These changes should not be included in the PR because they will cause the official tasks to stop working. Please do the following:

- remove all changes to the client-side code and leave only the server-side files in this PR. We should do the client-side integration in a separate PR and refactor the existing tasks to work with the new server. 
- move the folder `/mobile-browser-based-version/src/server` to `mobile-browser-based-version/server`
- include a README file in the newly created `mobile-browser-based-version/server` folder explaining the file structure and how to install, run and test the server. 

Thank you!

PaulMansat:
Hey @s314cy, I saw that part of your code includes a task router, that's great! 
It's just to warn you that we added a few entries to the task's definition. So if you want to update the task's definition, you can find them in the branch "new routing with new memory". If you cannot find the branch, it means that it has been merged, and you can find the new task's definition in the main branch. 
Thanks!

mobile-browser-based-version/server/run_server.js
```diff
@@ -1,28 +1,42 @@
...
+const tasks = JSON.parse(fs.readFileSync(myArgs[1]));
+
+const tasksRouter = express.Router();
+tasksRouter.get('/', (req, res) => res.send(tasks));
+tasks.forEach(task => {
+    tasksRouter.get('/' + task.taskId, (req, res) => res.send(models.get(task.taskId)))
+});
+
+app.get('/', (req, res) => res.send('DeAI Server'));
+app.use('/', peerServer);
```
mmilenkoski, line 40: I think this should be `app.use('/deai', peerServer);` to avoid conflicts

mmilenkoski (changes requested):
Could you please reverse all changes to the client-side? Currently, there are multiple changes to blank spaces and lines. I think we should not merge those. 

Additionally, why do we have normal and reverse proxied versions? Do we need both? If yes, we should specify when to use one or the other in the README.

mobile-browser-based-version/server/run_server.js
s314cy: This would just add an empty route, thus lengthening URLs. Currently, these two lines of code produce the following routes
- /: home page of the server
- /peerjs: home page of the PeerServer
- /peerjs/api/*: all services from the PeerServer
So there is no loss of clarity, nor any conflict with the server's home page as PeerServer doesn't serve anything on its mounting route (/).

mmilenkoski (approved):
Looks good, thanks for the work Hugo!

mobile-browser-based-version/server/tasks.json
```diff
@@ -0,0 +1,48 @@
...
+            "batchSize": 30,
+            "modelCompileData": {
+              "optimizer": "rmsprop",
+              "loss": "binaryCrossentropy",
+              "metrics": ["accuracy"]
+            },
+            "modelTrainData": {
+                "epochs": 10
+            },
+            "threshold": 1,
```
martinjaggi, line 41: what does threshold mean? (maybe also not very clear in the UI code which we could fix later with another PR)

mobile-browser-based-version/server/models.js
```diff
@@ -0,0 +1,36 @@
+const tf = require('@tensorflow/tfjs')
+
+
+function createTitanicModel() {
```
martinjaggi, line 4: how do we make sure the model architecture is kept compatible with the client-side task code? can the model itself only be defined here and be removed from the client-side code?

mmilenkoski: yes, that is the idea. We will remove the model creation from the front end once this is integrated.

mobile-browser-based-version/server/tasks.json
mmilenkoski: this is the number of neighbour weights we wait for before continuing training. It will be removed from the frontend, but we can give it a more descriptive name here. `received_messages_threshold` ?

martinjaggi: it wasn't changed yet i think. can someone rename it?

s314cy: I made the change (client-side included) in the #89 PR, to avoid reopening closed branches/PRs

PR #65, opened 2021-05-29
Fix some errors

PaulMansat:
- fix error in training alone (wasn't using the proper compile data)
- fix the asynchronous behavior of the  creation of models

### DIFF
--- a/mobile-browser-based-version/src/task_definition/lus_covid.js
+++ b/mobile-browser-based-version/src/task_definition/lus_covid.js
@@ -268,6 +268,7 @@ export const trainingInformation = {
         loss: "binaryCrossentropy",
         metrics: ["accuracy"],
     },
+    learningRate: 0.05,
     // {Object} Training information 
     modelTrainData: {
         epochs: 10,
--- a/mobile-browser-based-version/src/task_definition/lus_covid.js
+++ b/mobile-browser-based-version/src/task_definition/lus_covid.js
@@ -36,7 +36,7 @@ export class LusCovidTask {
 
         newModel.summary()
 
-        newModel.save(this.trainingInformation.savePathDb);
+        await newModel.save(this.trainingInformation.savePathDb);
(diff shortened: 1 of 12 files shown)


PaulMansat:
As soon as these changes are validated, I will merge them. 
This will fix the error we had with models with static performance after the 1st epoch. 
@MarcelTorne these enhancements might require you to change the optimizer defined in the LUS task (I saw that in the alone training loop the Adam optimizer was hardcoded, which is not the case anymore).

PaulMansat:
This will also fix the error we had when: 
(1) User fist log into the app 
(2) Joins the training of task 
(3) Error "Could not find the model working_[taskName]" was returned 

This is now fixed

mobile-browser-based-version/src/task_definition/mnist.js
```diff
@@ -240,7 +240,7 @@ export const trainingInformation = {
     // {String} model's identification name
     modelId: "mnist-model",
     // {String} indexedDB path where the model is stored
-    savePathDb: "indexeddb://working_lus_covid_model",
+    savePathDb: "indexeddb://working_mnist-model",
```
martinjaggi, line 243: underscore?

mobile-browser-based-version/src/task_definition/lus_covid.js
```diff
@@ -253,7 +253,7 @@ export const trainingInformation = {
     // {String} model's identification name
     modelId: "lus-covid-model",
     // {String} indexedDB path where the model is stored
-    savePathDb: "indexeddb://working_lus_covid_model",
+    savePathDb: "indexeddb://working_lus-covid-model",
```
martinjaggi, line 256: ok can say it's working_+id

PaulMansat: Right 👍

PR #68, opened 2021-06-01
Save working model after training to be able to test it

MarcelTorne:
After finished training save the working model in working + model_id to then be able to retrieve it in the testing component.

### DIFF
--- a/mobile-browser-based-version/src/task_definition/lus_covid.js
+++ b/mobile-browser-based-version/src/task_definition/lus_covid.js
@@ -265,7 +265,7 @@ export const trainingInformation = {
     batchSize: 2,
     // {Object} Compiling information 
     modelCompileData: {
-        optimizer: "adam",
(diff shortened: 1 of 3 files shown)


mobile-browser-based-version/src/task_definition/lus_covid.js
```diff
@@ -265,7 +265,7 @@ export const trainingInformation = {
     batchSize: 2,
     // {Object} Compiling information 
     modelCompileData: {
-        optimizer: "adam",
+        optimizer: "rmsprop",
```
mmilenkoski, line 268: is this and the deletion of the `threshold` property below intentional? If not, please remove

MarcelTorne: It was not intentional, I just changed it. Thanks for pointing this out!

mmilenkoski (approved):
Thank you Marcel!

PR #69, opened 2021-06-01
Improve testing csv results

MarcelTorne:
In the testing csv results add the id of the predicted data point to be able to identify it later.

This PR is dependent on #68

### DIFF
--- a/mobile-browser-based-version/src/helpers/training_script/training_manager.js
+++ b/mobile-browser-based-version/src/helpers/training_script/training_manager.js
@@ -67,6 +67,7 @@ export class TrainingManager {
                     this.trainingInformation.learningRate
                 );
             }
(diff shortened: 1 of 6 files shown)


mobile-browser-based-version/src/helpers/training_script/training_manager.js
```diff
@@ -117,6 +118,13 @@ export class TrainingManager {
         );
     }
 
+    saveWorkingModel(){
```
PaulMansat, line 121: Perfect, I think that loading/storing model for each function will be the new way of handling models throughout the entire application.

PaulMansat (approved):

mmilenkoski (approved):
Thanks for the changes Marcel!

PR #70, opened 2021-06-01
Fix gpu memory leak and lost WebGL context issue

### LINKED ISSUE #62: WebGL context lost error
I encounter an issue with WebGL when trying to train an image model (lus or mnist) on a large dataset. 

Full error:
lus-deai-error

mmilenkoski:
Resolves #62 

The issue was that intermediate value tensors from image preprocessing were unnecessarily kept in memory. This is fixed now with the use of `tf.tidy`


PaulMansat (approved):
Great, thank you Martin!

MarcelTorne (approved):
Thanks for fixing this issue Martin !!

PR #72, opened 2021-06-02
Fix deadlock on multiple peers communicating 

MarcelTorne:
For the moment I have fixed the deadlock but we should verify wether it is always waiting for all of the weights to arrive or if it already detects if all weights receive before the 10 seconds.

### DIFF
--- a/mobile-browser-based-version/src/helpers/communication_script/helpers.js
+++ b/mobile-browser-based-version/src/helpers/communication_script/helpers.js
@@ -117,12 +117,10 @@ export function dataReceivedBreak(recvBuffer, key, maxTries) {
  * @param {Array} arr 
  * @param {int} len 
  */
-export function checkArrayLen(arr, len) {
+export function checkArrayLenSync(arr, len) {
     let maxTries = 100
     return new Promise((resolve) => {
         (function waitData(n) {
-            console.log(arr.length)
-            console.log(n)
             if (arr.length >= len || maxTries <= n) {
                 return resolve();
             }
@@ -130,6 +128,26 @@ export function checkArrayLen(arr, len) {
         })(0);
     });
 }
+
+/**
+ * Waits until an array reaches a given length. Used to make 
+ * sure that all weights from peers are received.
+ * @param {Array} arr 
+ * @param {int} len 
+ */
+ export function checkArrayLenCommon(recvBuffer, len) {
+    let maxTries = 100
+    return new Promise((resolve) => {
+        (function waitData(n) {
+            const arr = Object.values(recvBuffer.avgWeights).flat(1)
+            if (arr.length >= len || maxTries <= n) {
+                return resolve();
+            }
+            setTimeout(() => waitData(n+1), 100);
+        })(0);
+    });
+}
+
 
 // generates a random string
 export async function makeid(length) {
@@ -167,7 +185,7 @@ export async function onEpochEndSync(model, epoch, receivers, recvBuffer, peerjs
         await dataReceived(recvBuffer.avgWeights, epoch.toString())
     }
     console.log("Waiting to receive all weights for this epoch...")
-    await checkArrayLen(recvBuffer.avgWeights[epoch], receivers.length)
+    await checkArrayLenSync(recvBuffer.avgWeights[epoch], receivers.length)
         .then(() => {
             console.log("Averaging weights")
             for (i in recvBuffer.avgWeights[epoch]) {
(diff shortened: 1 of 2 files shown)


PaulMansat:
Thanks for noticing this issue Marcel! 
Maybe it's a bit strange to wait for data to arrive when checking the array's length. Maybe we can return the array's length all together without waiting for data to arrive (we are doing the waiting with the function dataReceivedBreak). 
I'll have a look into it tomorrow. 
Thanks again, I hope you are now able to train with 4 peers 👍.

MarcelTorne:
> Thanks for noticing this issue Marcel!
> Maybe it's a bit strange to wait for data to arrive when checking the array's length. Maybe we can return the array's length all together without waiting for data to arrive (we are doing the waiting with the function dataReceivedBreak).
> I'll have a look into it tomorrow.
> Thanks again, I hope you are now able to train with 4 peers 👍.

Yeah that's true, my thing is just a quick fix. But the problem with just returning the arrays length is that once you receive one weight it will directly continue to averaging. So there should still be a break until you receive `threshold` weights or 10 seconds elapsed. What do you think?

MarcelTorne:
I found the other problem we had, on `onEpochEndCommon` we use create a copy of the average weights, so once it was passed into the function it would never change again since it was a copy of the array not a pointer to the recvBuffer object. Please let me know what do you think of the fix and wether you have some better ideas.

mobile-browser-based-version/src/helpers/communication_script/helpers.js
```diff
@@ -116,18 +117,38 @@ export function dataReceivedBreak(recvBuffer, key, maxTries) {
...
+    });
+}
+
+/**
+ * Waits until an array reaches a given length. Used to make 
+ * sure that all weights from peers are received.
+ * @param {Array} arr 
+ * @param {int} len 
+ */
+ export function checkArrayLenCommon(recvBuffer, len) {
```
mmilenkoski, line 138: This seems like code duplication to me. Is it possible to have only `checkArrayLenSync` and call it with `Object.values(recvBuffer.avgWeights).flat(1)` to achieve the functionality of `checkArrayLenCommon` or have a flag that indicates if you need to call `Object.values(recvBuffer.avgWeights).flat(1)` inside?

mobile-browser-based-version/src/task_definition/lus_covid.js
```diff
@@ -280,7 +280,7 @@ export const trainingInformation = {
     modelTrainData: {
         epochs: 10,
     },
-    threshold: 1,
+    threshold: 2,
```
mmilenkoski, line 283: is this change intentional?

mobile-browser-based-version/src/helpers/communication_script/helpers.js
```diff
@@ -116,18 +117,38 @@ export function dataReceivedBreak(recvBuffer, key, maxTries) {
  * @param {Array} arr 
  * @param {int} len 
  */
-export function checkArrayLen(arr, len) {
+export function checkArrayLenSync(arr, len) {
+    let maxTries = 100
```
mmilenkoski, line 121: please make this a constant on top of the file

mmilenkoski (changes requested):

mobile-browser-based-version/src/task_definition/lus_covid.js
MarcelTorne: Well, it depends on how many peers we want to wait for, so I increased in for the benchmarks. However, I suppose that in the future we can either adapt it dynamically or let the user set it in the interface. I'll still set it back to 1.

mobile-browser-based-version/src/helpers/communication_script/helpers.js
MarcelTorne: Before we used to have only `checkArrayLenSync` and we called it with `Object.values(recvBuffer.avgWeights).flat(1)` but the problem is that this is not a pointer to the `recvBuffer.avgWeights` anymore but it's a copy so its value will never change and hence even if a new weight is received you will not detect and you will wait for 10 seconds.

MarcelTorne: I can add the flag but I don't really like it with the flag, and I don't have any better idea for a relatively quick fix. Do you?

mmilenkoski: okay, I understand. Why don't you like it with a flag though? In my opinion, it is better than having two functions with 95% duplicated code.

MarcelTorne: Yes definitely, I already changed it before, but I was asking wether you would have a better idea, sorry for the misunderstanding.

PaulMansat:
Hey @MarcelTorne, could you resolve the conflict on this branch? I think it's simply about making the choice of keeping or net the check array function in sync. Thanks! 
When this is done, we can merge the branch to close all this PR because there are 2 other big PR to merge (one changing how files are uploaded and the other changing the routing architecture). Thanks!

PaulMansat (approved):

Issue #71, opened 2021-06-02
Bug: Multiple peer communication (3 or more)


MarcelTorne:
When training in a distributed manner with 3 peers or more there is a deadlock if one of the peers is not sharing weights and you set a threshold of 2.

The threshold should only be present for waiting until you receive all weights before you average weights but we should still put a time threshold (for example you will wait max 10seconds). 

The file to be fixed is `helpers.js` in `helpers/communication_script/helpers`. One first problem I have seen is that `checkArrayLen` function in line 120 is useless and creating an infinite loop. The `arr` argument will never change and we are doing a loop until its length increases. There is a similar problem with `dataReceivedBreak`. I think the way to fix this is to have a general object accessible to all of these functions and then add a limit in number of tries to `checkArrayLen`.

PR #74, opened 2021-06-03
Improve weights aggregation

MarcelTorne:
In this PR I am improving the weights aggregation during distributed learning. Now, we are doing an average at each epoch over all weights received in that epoch. 

This PR is made on top of #71

### DIFF
--- a/mobile-browser-based-version/src/helpers/communication_script/helpers.js
+++ b/mobile-browser-based-version/src/helpers/communication_script/helpers.js
@@ -44,13 +44,19 @@ export function assignWeightsToModel(serializedWeights, model) {
     });
 }
 
-export function averageWeightsIntoModel(serializedWeights, model) {
+export function averageWeightsIntoModel(peersSerializedWeights, model) {
     model.weights.forEach((weight, idx) => {
-        const serializedWeight = serializedWeights[idx]["$variable"];
-
-        const tensor = deserializeTensor(serializedWeight.val);
-        weight.val.assign(tensor.add(weight.val).div(2)); //average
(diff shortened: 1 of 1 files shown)


PaulMansat (approved):
Great, thank you Marcel!

mobile-browser-based-version/src/helpers/communication_script/helpers.js
```diff
@@ -44,13 +44,19 @@ export function assignWeightsToModel(serializedWeights, model) {
...
+        let tensorSum = weight.val
+        peersSerializedWeights.forEach(
+            (serializedWeights, peer) =>{
+                const serializedWeight = serializedWeights[idx]["$variable"];
+                const tensor = deserializeTensor(serializedWeight.val);
+                tensorSum = tensor.add(tensorSum)
+                tensor.dispose()
+            }
+        )
+        weight.val.assign(tensorSum.div(peersSerializedWeights.length)); //average
```
mmilenkoski, line 58: I think it should be changed to 
```
weight.val.assign(tensorSum.div(peersSerializedWeights.length+1)); //average
```
because we also include the local weight in the sum

mmilenkoski (changes requested):
just a small change needed, otherwise all good.

mobile-browser-based-version/src/helpers/communication_script/helpers.js
MarcelTorne: Oh yes definitely! Thanks for pointing this out!

mmilenkoski (approved):
thank you Marcel!

PR #77, opened 2021-06-05
New routing with new vue memory management

PaulMansat:
- solve the issue that resulted in a task's component being mounted multiple times (in fact, to see this issue, one had to first log into one task and then move to another task, there the user would be connected twice to the server because the component would be mounted twice)
- to solve the issue a new routing procedure is introduced with a new architecture: all components share a global main component where the routes for the child components (i.e desc. component, training component, and testing component are called). I will add in the documentation a scheme of the new routing procedure.
- the global task component being shared amongst all tasks, it has to store its state in a Vuex store (used to store custom component's state)
- Vuex is added as a new dependency (to work the npm dependencies of the application need to update by running the command npm install)
- this issue was also related to the fact that some variables used in a component were defined outside the component's scope. This meant that these variables were shared by all of the component's instances. Originally, these variables were used to store a TFJS model, because TFJS models cannot be defined in the component's data, due to memory conflicts between TensforflowJS and VueJS. The solution is to store the TFJS model in indexedDB, and to load the model only when we need it (i.e only when training and not before in a component).
- the new routing procedure required some modifications in many files, that is the reason why this pull request is quite large.
- finally, a minor correction was added to the communication manager. The port number in the communication number was fixed to 9000 and was not connected to the port number information defined in the task's training information.


mobile-browser-based-version/src/components/building_frames/ImageTestingFrame.vue
```diff
@@ -326,17 +324,12 @@ export default {
     },
     downloadPredictionsCsv(){
       console.log(this.predictions)
-      let pred = ""
-      let header_length = 0
-      for (const [id, prediction] of Object.entries(this.predictions)){
-        header_length = prediction.length
-        pred += id + ","+ prediction.map(dict => dict["className"]+","+dict["probability"]).join(',')+"\n"
```
PaulMansat, line 333: I might have deleted a part you've updated @MarcelTorne. Could you confirm that the new version is correct? Thanks!

mobile-browser-based-version/src/helpers/communication_script/helpers.js
```diff
@@ -44,19 +44,13 @@ export function assignWeightsToModel(serializedWeights, model) {
...
-            (serializedWeights, peer) =>{
-                const serializedWeight = serializedWeights[idx]["$variable"];
-                const tensor = deserializeTensor(serializedWeight.val);
-                tensorSum = tensor.add(tensorSum)
-                tensor.dispose()
-            }
-        )
-        weight.val.assign(tensorSum.div(peersSerializedWeights.length + 1)); //average
-        tensorSum.dispose();
+        const serializedWeight = serializedWeights[idx]["$variable"];
```
PaulMansat, line 49: I might have overwritten some files again...

PaulMansat: Resolved: brought back the old file

mobile-browser-based-version/src/components/building_frames/ImageTestingFrame.vue
PaulMansat: Resolved: brought back the original file in from the maste branch.

mobile-browser-based-version/src/task_definition/mnist.js
```diff
@@ -201,14 +200,12 @@ export class MnistTask {
...
-                classes_dict[i] = classes
-                
-                i++
+                classes_array.push(classes)
             }
 
             console.log("Prediction Sucessful!")
 
-            return classes_dict;
+            return classes_array;
```
PaulMansat: Resolved: brought bach the original file from the master branch.

mobile-browser-based-version/src/task_definition/lus_covid.js
```diff
@@ -197,72 +191,6 @@ export class LusCovidTask {
...
-        }
-
-        const xs = tf.concat(xsArray, 0)
-    
-        console.log(xs)
-        console.log(patients)
-
-        return {xTest: xs, ids: patients}
-    }
-
```
PaulMansat: Resolved: brought back the original branch from the master branch.

martinjaggi (dismissed):

mobile-browser-based-version/src/store/store.js
```diff
@@ -0,0 +1,29 @@
+import {createStore} from 'vuex'
```
mmilenkoski, line 1: this file needs to be updated as we discussed

mmilenkoski (changes requested):

mobile-browser-based-version/src/store/store.js
PaulMansat: The issue is resolved, the new code is in the file MainTaskFrame and not GlobalTaskFrame 👍

mmilenkoski (approved):
Thanks a lot for working on this Paul!

mmilenkoski (approved):

PR #79, opened 2021-06-06
Fix creation model error + change "train alone" button's name

PaulMansat:
Fix issue when the model is created: when the "join training" button was clicked, every time a new model was created (even when a model had previously been created).

### DIFF
--- a/mobile-browser-based-version/src/components/building_frames/DescriptionFrame.vue
+++ b/mobile-browser-based-version/src/components/building_frames/DescriptionFrame.vue
@@ -224,6 +224,7 @@ export default {
   },
   data() {
     return {
+      isModelCreated: false,
       savedModelExists: false,
       readyToTrain: false,
       choicePreModel: false,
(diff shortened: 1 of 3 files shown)


mobile-browser-based-version/src/components/building_frames/DescriptionFrame.vue
```diff
@@ -235,15 +236,19 @@ export default {
   methods: {
     async goToTraining() {
       if (!this.choicePreModel) {
-        await this.createNewModel();
-        this.readyToTrain = true;
+        if (!this.isModelCreated) {
```
mmilenkoski, line 239: nitpicking, but could you please merge these two ifs in one?

mmilenkoski (approved):
small change and ready to merge. Thanks Paul!

martinjaggi:
maybe let's keep the sad 'Train alone' naming for now. It's ok if that has a negative connotation (and button could maybe have a different grey color, compared to the green main collaborative training button)

mmilenkoski (approved):

mmilenkoski (approved):

PR #81, opened 2021-06-08
Add relaySGD backend to server

mmilenkoski:
This PR adds the server backend needed for RelaySGD. The main files of interest are:

- `run_server.js` - this will run PeerJS as part of a DeAI app. The point is that we will be able to use the same port for all interaction related to one task. For now, we add only one additional endpoint `/neighbours/:id` that returns the neighbours of the node with id `:id` according to the used topology.
- `topology.js` - we will use this file to define the topologies. Currently, we use only a binary tree topology. However, we have modular implementation and we can easily exchange it with another topology in the future. 

Few words regarding the binary tree topology:

- When a user joins, it is added as a child to the shallowest existing node that doesn't already have two children. This is done in order to keep the tree as balanced as possible. If no user leaves, the tree grows left to right until the last layer is filled before starting a new layer. In the case below, a new user would be added as a left child to the node `42`
add_tree

- When a user leaves, we find the deepest branch starting from the node that leaves and we propagate the nodes upwards by one step, and then delete the deepest node. This is also done to keep the tree balanced with the minimum change in the tree topology. In the tree below, if user `37` leaves, we would move the nodes as indicated by the red arrows, and we would delete the deepest node that had an id `23` before. As discussed with @tvogels, we might further optimize this in the future. 
bst_remove_2

- As neighbours of a node, we return the parent, left and right child if they exist. 

Next step is to modify the front end to integrate RelaySGD


mobile-browser-based-version/server/index.html
```diff
@@ -0,0 +1,23 @@
+<script src="https://unpkg.com/peerjs@1.3.1/dist/peerjs.min.js"></script>
+<script>
+    peer_ids = 'abcdefghi'.split('')
```
tvogels: wow, lazy 😝

tvogels:
Looks good to me! Only nitpick style-wise is a mix between camelCase and underscores, but that's none of my business ...

How are you planning to switch between different training algorithms for different tasks/projects?
The server code maybe doesn't need to be algorithm dependent? I guess the server only cares about which topology is used.

tvogels (dismissed):

martinjaggi:
looking very good. also we can maybe clarify a bit more what the signaling patterns are, ideally mostly local: 
- if someone detects that their peer is not reachable, do they send a message to the server? how do we determine if it's them or the peer who dropped? and once the topology changes, only the affected peers get notified i suppose?
- and for relaySGD: do all nodes need to know the global n, or not needed (as it can change)

tvogels:
> for relaySGD: do all nodes need to know the global n, or not needed (as it can change)

officially yes, but maybe we can drop it and make it part of the learning rate. That would give automatic learning rate scaling as more workers join, and it simplified the algorithm a bit

mmilenkoski:
> * if someone detects that their peer is not reachable, do they send a message to the server? how do we determine if it's them or the peer who dropped? and once the topology changes, only the affected peers get notified i suppose?

The server receives an event when a user drops and can notify the peers. My idea was for the affected clients to receive an event when their neighbor list change. What do you think? I would add that in a separate PR.

martinjaggi:
> > * if someone detects that their peer is not reachable, do they send a message to the server? how do we determine if it's them or the peer who dropped? and once the topology changes, only the affected peers get notified i suppose?
> 
> The server receives an event when a user drops and can notify the peers. My idea was for the affected clients to receive an event when their neighbor list change. What do you think? I would add that in a separate PR.

sounds good! (in the longer term we'd like to have as many of these mechanisms as possible to happen decentralized. but as long as we have the peerJS helper server it's all fine as now)

martinjaggi (approved):

martinjaggi:
@tvogels i hear rumors about this interesting RelaySGD, is there some arxiv paper about it?

tvogels:
> @tvogels i hear rumors about this interesting RelaySGD, is there some arxiv paper about it?

Haha, it’s accessible through our private arxiv. Shouldn’t we wait for review to finish?

tvogels (approved):

PR #83, opened 2021-06-09
Add server-sent events on node removal

mmilenkoski:
This PR adds the server-side capability for the peers to subscribe for changes in their neighbour list.  The client changes will be in a separate PR. The peers will become subscribed the first time they ask for the list of neighbours, so they will only need to ask once (as is the case right now).  

I also modify the `removePeer` function in the Binary Tree to return the list of affected peers (the peers who will be notified of the change). In the case of the binary tree, the affected trees are the nodes on the deepest branch starting from the node that is going to be removed and their neighbours. 

In the image above, if we are deleting the node `37`, the nodes `23`, `22` and `28` are moved up by one node, and the list of notified nodes will be `[23, 22, 29, 28, 38, 40]`
bst_remove_2

For this, we use the [server-sent events](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events) functionality.


mobile-browser-based-version/server/run_server.js
```diff
@@ -1,14 +1,17 @@
...
 var topologies = require('./topologies.js')
 
 var myArgs = process.argv.slice(2);
 
 const app = express();
+app.use(cors());
+app.use(express.json());
+app.use(express.urlencoded({extended: false}));
 
 const topology = new topologies.BinaryTree()
```
tvogels, line 13: Do you plan this to be configurable per experiment in the long run?

mobile-browser-based-version/server/run_server.js
```diff
@@ -19,10 +22,46 @@ const peerServer = ExpressPeerServer(server, {
...
+  
+    peers.push(newPeer);
+  
+    request.on('close', () => {
+      console.log(`${peerId} Connection closed`);
+      peers = peers.filter(peer => peer.id !== peerId);
+    });
+  }
+  
+  app.get('/neighbours/:id', eventsHandler);
```
tvogels, line 53: Something's off with the indentation here. Got a bit confused thinking that this was inside `eventsHandler`.

mobile-browser-based-version/server/topologies.js
```diff
@@ -92,6 +96,8 @@ class BinaryTree {
             deepestParent.rightChild = null
         }
         delete this.index[nodeId]
+
+        return affectedPeers
```
tvogels, line 100: API-wise it seems a little uninutitive to me that the `removePeer` frunction returns `affectedPeers`. If it you want this behavior, I think `addPeer` should do the same.

I think the code would be easiest to understand if the trees are immutable, and `removePeer` returns a new tree. You can then have a separate function `topologyDiff` or something to find out which peers are affected. But I can also live with this, if you prefer it.

tvogels (commented):
Looks good.
I'm wondering how reliable the `disconnect` event from `ExpressPeerServer` is. Do you know if this uses websockets or long-running HTTP connections?  I wonder if these can transparently reconnect if the connection was lost very briefly.

I kind of expect it to trigger a disconnect&connect when a mobile device briefly looses connection (which is quite common). Maybe some heartbeat would be a more reliable way to detect peers that disappeared? Also, could a node report to the server that it cannot reach a peer?

I guess those challenges are not necessarily related to this commit.

lippoldt:
I would love to discuss this in a meeting - and get an overview over the whole update system (diagram).
I really found the binary tree to be an elegant solution - because somewhat we preserve the distributed idea by only knowing parent and children per node.

mobile-browser-based-version/server/run_server.js
mmilenkoski: yes, once we have more than one topology

mobile-browser-based-version/server/topologies.js
mmilenkoski: I agree that `addPeer` should also return `affectedPeers` since we need to notify them (in this case only the parent) as well, but I think copying the entire tree on every addition and removal is not really scalable in a dynamic system with a lot of nodes.

mmilenkoski:
> I'm wondering how reliable the `disconnect` event from `ExpressPeerServer` is. Do you know if this uses WebSockets or long-running HTTP connections? I wonder if these can transparently reconnect if the connection was lost very briefly.
> 
> I kind of expect it to trigger a disconnect&connect when a mobile device briefly loses connection (which is quite common). Maybe some heartbeat would be a more reliable way to detect peers that disappeared? Also, could a node report to the server that it cannot reach a peer?
> 

yes, this is definitely something we need to check and handle. Thanks for the suggestion!

mobile-browser-based-version/server/topologies.js
tvogels: In terms of scalability, there are very nice libraries for immutable data structures (https://immutable-js.github.io/immutable-js/) that don't actually need to duplicate the whole tree if you edit it (although it reliably behaves like you duplicated the whole tree). But no need to look into that now.

PR #89, opened 2021-06-17
Server integration

s314cy:
This PR adds client's support for the helper server. Peer ID generation, tasks and the model architectures are now centralized. A global server manager object (client-side) handles the server's requests (subject to change, not the cleanest).

Tasks and their descriptions, training informations and routes are loaded on mounting the TaskList component. A task's model is downloaded on calling the createModel function in the task class (first call is on mounting the DescriptionFrame).

Only the Titanic task is supported as of now, I'll add the rest of the tasks after review to avoid unnecessary modifications.


mmilenkoski:
closing this PR because it was integrated with #113 and #107

PR #91, opened 2021-06-21
Fix HTTPS error in Github Pages deployment

mmilenkoski:
Currently, the deployed app on Github Pages can't connect to the server because Github Pages uses SSL, and we do not have a valid SSL certificate on the current Peer.js server. 

However, Google App Engine (GAE) automatically creates an HTTPS certificate if we deploy the app there. 

This PR introduces the necessary files and README instructions to deploy the app on GAE. 

I already deployed the app on GAE and pointed our frontend to the new host. It is available on this [link](https://deai-313515.ew.r.appspot.com).

After merging, this should enable the frontend to connect to the PeerJS server when deployed on Github Pages.


mobile-browser-based-version/server/README.md
```diff
@@ -18,8 +18,19 @@ Tasks are stored in ```tasks.json```. The models are declared in ```models.js```
 
 ### Running the server
```
martinjaggi, line 19: better call it helper server (only keeping list of available peers per task, not data)

martinjaggi (dismissed):

Issue #82, opened 2021-06-08
Allow task owner to set a password for joining a task


martinjaggi:
For allowing joint training within a smaller set of trusted participants only (in addition to public tasks)

- [ ] Add an optional password hash to the task description. 
- [ ] If present, new users will be asked the password before being able to join a task (like "joining this task is password restricted").
- [ ] Add this to the documentation of how users can create a new task.

This functionality basically recovers federated learning as a special case. This is not giving the same security standard as a full PKI, and is not a replacement for a selection mechanism for helpful clients (as opposed to Byzantine ones), but is a start (see the same discussions in federated learning literature)

martinjaggi:
we can close this i think? except maybe should add the feature to the documentation somewhere?

martinjaggi:
and remove from mnist again

martinjaggi:
ping

mmilenkoski:
Closing the issue since the functionality is now implemented.

We can allow the user to enter a password when they create a task through the form in the UI. 

I also removed the password protection for Titanic already.

PR #96, opened 2021-06-24
Add password protection for tasks

mmilenkoski:
This PR implements the client-side changes for #82. 

For this demo, I keep the password in the task information of the MNIST task. After #89 is merged, we can move to server-side authentication. 

Demo:

https://user-images.githubusercontent.com/10998465/123310626-9201d800-d526-11eb-87e9-fd90a295115b.mp4


martinjaggi (changes requested):
we can make this a bit more secure, and without needing trust in any server.

whenever peers exchange a model, they also send the hash of `from-peer-id+" "+password`. the receiving peer then checks if this received hash is the same as what they would compute with what they think is the password. if not it will decline the update.
to initially let the user know if they have the right password or not, we check agains the hash of the password alone. this hash is publicly stored in the task definition.

the scheme above doesn't yet protect from peers imitating others or sybil attacks, but that can be solved in a later stage by adding authentification if necessary

this way others can't simply eavesdrop the password. does not prevent si

martinjaggi:
we can for example use SHA256 from the library here: https://github.com/h2non/jshashes

mmilenkoski:
hey @martinjaggi, I added your suggestions. Let me know what you think. I moved the password to titanic to test more easily. The hash is from the password `deai_titanic`

mmilenkoski:
> should this function be called receive sth, or handle?

I think it is okay like this since it handles the data received by peerjs based on the type. We can change it if you think it would be better.

martinjaggi (approved):

Issue #84, opened 2021-06-14
generic data processing capabilities I


lippoldt:
Generalize and improve current data processing pipeline, with the goal to allow easy utilization of data as part of the projects, facilitate technical debugging, facilitate understanding of data science background and challenges and to separate those challenges from another.

In a minimalistic setting, this includes:
1) - [x]    Catching data related errors as soon as possible in the MLOps pipeline
2) - [ ]   Providing functions that connect the tasks with the interface to include/exclude data types from loading
3) - [x]   Check on data loading success status 

In addition, the following features would be nice to have: (now or later)
4) Ability to get data science style diagnosis of data on the interface (label distribution, before/after training evaluation)
5) capabilities to handle advanced bugs relating to data loading - handle corrupt/invalid data in an optimal way 
6) minimize RAM/VRAM data and possibly provide overview on the UI

As a follow up, this would require current tasks/projects to be tested and checked again and verified, with possibly changes in the testing functions currently set up separately. Some tasks might be better connected to the UI and UI changes would improve the overall usability. (possibly in a separate issue)

If data loaders work similar to the python version, a lot of these tasks can be handled naturally, but some sub tasks would require follow ups.

lippoldt:
Minimalistic version has been done with #100  - follow up on 2) with #101

PR #100, opened 2021-07-04
image validation functionality

lippoldt:
This update proposes the following changes relating to issue #84 

1) Validate images before preprocessing and outside of task file and only proceed if validation succeeds
2) update MNIST example (remove double data validation call)
3) update validation function for images to include minimum height and width as determined in the task specifications
4) catch images that cannot be loaded


mmilenkoski:
This looks good, but I think we are now loading the images twice, once for validation and once for preprocessing in the function `imagePreprocessing` in `mnist.js`. Maybe we can move the validation in the preprocessing function to avoid this. What do you think?

martinjaggi (approved):

lippoldt:
> This looks good, but I think we are now loading the images twice, once for validation and once for preprocessing in the function `imagePreprocessing` in `mnist.js`. Maybe we can move the validation in the preprocessing function to avoid this. What do you think?

When I started on the data pipeline I also thought that doing loading once only is the best option. However, I have found a bunch of arguments that would go against this (load of work required for changes and also future compatibility thinking about not only short term but medium term outcomes).

Arguments for image validation separated from tensor loading and network training:
1) currently the tensor loading happens directly when loading the image. this means the image is not actively checked or passed in the task (mnist for example) file itself. It would require rewriting and redesigning the task files to be able to check on all images as well. 
2) If we do transition from load all images to load only necessary images (formal dataset loading),  we would forward the image loading error into the training process instead of catching it before. 
3) if we validate images before training, we have further future capabilities for checking on bad images or detect data that is not "good for training" (we have not determined that exact definition yet, but it would be open for further progress)

That is why, assuming the garbage collection works well, and time in seconds is not an issue, I would strongly suggest validating images before loading tensors.

Happy to also discuss this in a meeting.

mmilenkoski (approved):

Issue #112, opened 2021-08-11
Define one general Task class


mmilenkoski:
We should define one Task class that can be completely specified by a config file, so that a task can be created and served on the server without any task-specific code on the client-side. 

The main challenge is to generalize the data preprocessing steps by offering few predefined options that can be chosen in the config.

PR #113, opened 2021-08-11
Load tasks and models from server

mmilenkoski:
This PR incorporates, modifies, and augments the client-side changes introduced in #89.

I will close that PR when we merge this one. 

The next step is to create one Task class that can be completely specified by a config file. The main challenge there is to generalize the data preprocessing and offer few predefined options. I opened #112 to resolve this.

### DIFF
--- a/mobile-browser-based-version/src/components/TaskList.vue
+++ b/mobile-browser-based-version/src/components/TaskList.vue
@@ -8,7 +8,7 @@
       >
         <section class="flex-col items-center justify-center p-4 space-y-4">
           <div
-            v-for="task in ALL_TASKS"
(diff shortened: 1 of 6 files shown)


martinjaggi (approved):

lippoldt (commented):
Nice! So this would still be required to be extended to the other tasks apart from Titanic?

mmilenkoski:
> Nice! So this would still be required to be extended to the other tasks apart from Titanic?

No, all three are now moved to the server and functional. What we need is to generalize it to avoid the switch statement in `TaskList` and have only one Task class that can be reused for different tasks based on the task config on the server

PR #116, opened 2021-08-19
Export models from model library

### LINKED ISSUE #109: export models from model library
new functionality to export as a file

mmilenkoski:
resolves #109 

New look:
download-button


martinjaggi (approved):

mmilenkoski:
> nice. maybe call the whole thing library instead of memory, to be more clear?
> 
> later we'll also add another button to start use any of these models, i.e. link to the inference page, right?

Maybe it is better to have the name of the model be clickable and point to the inference page to avoid cluttering the panel with more buttons?

PR #135, opened 2021-10-14
Simulation scripts

laurislopata:
Added simulation script for Titanic, MNIST and CIFAR10. Added Different Data Split options for MNIST and Covid Lungs


martinjaggi (approved):
looking good. cosmetic detail: for easier use for newcommers, can you add some more instructions to each task as to where to download the datasets and in which format? also let's add an overall readme in the folder to explain what this does? to the readme you could maybe also add some preliminary results (when changing the number of workers say, for example reporting time-to-accuracy or similar metrics)

laurislopata:
Good suggestions, I will work on that

laurislopata:
Added CPU and RAM usage metrics, some intermediate results and modularised and cleaned up the code. I believe this is ready for merging

laurislopata:
Yes, that was a mistake, fixed it now

Nacho114:
I have not followed this at all, is it ok to merge? (I assume the code did not modify the app itself since it's still possible to merge without any conflicts).

laurislopata:
since the results here are outdated, I will update the readme with the new results when I test out the platform after my changes will be merged (BatchWise preprocessing and the use of MobileNet), and then we can merge that PR so that the results would be more advertiser-friendly. I'll do this next week after my exams

martinjaggi (changes requested):
the folder structure and potential data files duplication should be revisited.

for example `selenium-automation-scripts/labels.csv` and `selenium-automation-scripts/preproc... ` do seem specific to cifar/image tasks. better to keep the simulation framework task agnostic and make a data folder for each task

laurislopata:
I agree, I'll address it

PR #138, opened 2021-10-31
Task creation form

giordano-lucas:
Implementation of the form to add new tasks to the platform.

There are still a few things to do :
1. Ensure `model.json` and `weights.bin`  are valid in `TaskForm.vue` before upload
2. Remove `other` option field for `dataType` in `TaskForm.vue` 
3. Change tailwind CSS for some input tag and make them responsive (not fully working with mobile version)
4. Update texts on the home page if needed

Remarks / Questions :
1. I had to increase the maximum file size on the server to be able to send the `weights.bin` file. I set it to 20mb. Is this value ok ?
2.  Should I initialize `weights.bin` with random values if the user does not provide this file?


giordano-lucas:
@martinjaggi :
1. Do you mean the `dataType` field which can type the `csv` and `image`values` ? If that's the case then yes I guess I could try to implement that.
2. I will rename the file.
3. Which title are you referring to here ?

martinjaggi:
ah sorry that my comments appear a bit out of context here. if you click on the commit above though then they are linked to the respective code lines. for 1) i mean the whole filename of the vue file

martinjaggi:
model initialization: this should be specified as a field, either by selecting predefined options (random with a fixed seed, or better some standard supported initialization method e.g. for CNNs.), and then as the last alternative, the user can specify a custom initial weights file here. (same format as we export model weights)

mobile-browser-based-version/src/components/TaskList.vue
```diff
@@ -101,95 +106,110 @@ export default {
         query: {Id:id}
       });*/
     },
+    setUpTask(task) {
```
martinjaggi, line 109: can you call it `newTaskCreation`?

mobile-browser-based-version/src/components/TaskForm.vue
```diff
@@ -0,0 +1,1051 @@
...
+            as: "input",
+            type: "checkbox",
+            default: false,
+          },
+        ],
+        other: [],
+      },
+      // *** Section ***
+      {
+        title: "Model Compile Data",
```
martinjaggi, line 851: i don't understand the meaning of this one

mobile-browser-based-version/src/components/TaskForm.vue
```diff
@@ -0,0 +1,1051 @@
+<template>
```
martinjaggi, line 1: can you call the file `NewTaskCreationForm`?

martinjaggi: also we should document somewhere what code people have to change here, in case they add a new feature preprocessing code. how is it linked here, what can be shared and not? 

another question: is there any code-duplication with the UI code of each task? (names, descriptions?) if so, could we only define those in one js place and import here without duplicating them?

martinjaggi (changes requested):

mobile-browser-based-version/src/components/TaskList.vue
giordano-lucas: done

mobile-browser-based-version/src/components/TaskForm.vue
giordano-lucas: I just copy-pasted the fields that already are in the `server/tasks.json` file

giordano-lucas: I guess they should be used by a component of the platform at some point but I have to say I haven't checked

mobile-browser-based-version/src/components/TaskForm.vue
giordano-lucas: 1. I renamed the required files 
2. Yes I think we should do that. Right now, I believe we have to add another file in `src/task_definition` to add a new preprocessing pipeline.  
3. Actually yes and it's a bit of a problem because it's not limited to the UI code of each task, there is code duplication all over the `vue` files. I think it would be good to create some pre-stylised containers and components that people can reuse

PR #141, opened 2021-11-06
UI code refactoring

giordano-lucas:
# Better modularisation of UI code

Already done :

1. Create a `containers` folder with reusable containers such as a `BaseLayout` (containing the footer and the mains divs) 
2. Create a `simple` component folder with pre-stylised components such as the custom footer and buttons
3. Started to rewrite the `Home`, `TaskList`, `Information` pages.

Next steps :
1. Identity new modularisation opportunities in `building_frames`,  `main_frames` and `Trophee`
2. After the merge with `task-creation-form`, update its code as well

Should I rename the custom components with a shorted name such as `CFooter` instead of `CustomFootter` ?


giordano-lucas:
What  remains to do
1.  Bugs in rendering strings as HTML in `CSVTrainngFrame.vue` and others.
2. Solve display bugs in `TrainingFrame` (and forward the changes to `TestingFrame`)
2. I tried to modularize further the `SingleUploadFile` but I couldn't manage to get a working implementation
3. The `App` file is not yet done
4. Use [vue-i18n](https://vue-i18n.intlify.dev/)  internationalisation package to define two versions of the texts one for DeAI and one for  FeAI.

Further ideas :

1.  Put all data into data attribute and use `v-for`  structure to reduce code size (may not be possible with `svg` tags since they cannot be stored in data properties and I couldn't make it work with strings and an inline svg tag)
2. Try again the modularisation of `SingleUploadFile`

Questions :

1. Naming conventions of Vue components : eg. `TaskForm` or  `task-form` ?
2. Should the names of `svg` icons files contain the word Icon ?

s314cy (commented):
Great! I like how we use i18n to "translate" between platforms available on DeAI. However, I'm not entirely sure we want to use this for the server URLs. Indeed, this requires the backend to receive an environment (i.e. a component) as argument, which makes it UI dependent. It is dependent on a Vue package too.

Also, it feels like quite a workaround for something I think could be a simple object reading from a JSON. Indeed, from a dev POV, we are already using i18n (a language translation library) to switch between our platforms, which is fine, and then use it even further to create our API endpoints per platform. All in all, I think we should limit i18n to literal strings, rather than depend entirely on it to switch between platforms and make it encapsulate every single platform-dependent string.

Finally, there are two types of server requests. Those made within the functions in communication manager, a class containing platform specific code. Hence a different class will be used per platform anyway, no need to make its request URLs dynamic. Then, there are those made throughout the code, usually to fetch initial task metadata or to connect/disconnect. This is code executed no matter the platform, that communicates on the same endpoint server-side, e.g. `/tasks/*`, again, no matter the platform. Hence the need for dynamic URLs is not there either. (This paragraph is just describing the following fact: the code corresponding to the possibly differing URLs is either so different per platform that different code files will be required anyway, or it is exactly the same code, and then, exactly the same URL.)

What do you think?

Also, great changes on the server modularization, it will make the FeAI->DeAI merge even easier.

giordano-lucas:
@s314cy I agree with you, if we use different communication protocols then it does not make sense to use i18n for this. I will update the code anyway after the merge.

s314cy (approved):

mobile-browser-based-version/src/components/building_frames/csv/CsvTrainingFrame.vue
```diff
@@ -0,0 +1,138 @@
...
+                v-for="header in headers"
+                :key="header.id"
+              >
+                <div
+                  class="select-none p-2 transition duration-500 ease-in-out transform hover:-translate-y-2 rounded-2xl border-2 p-6 hover:shadow-2xl border-primary-dark"
+                >
+                  <div class="grid grid-cols-3 items-center p-2">
+                    <div class="pl-1">
+                      <div class="font-medium">
+                        <div class="flex flex-row justify-start">
```
Nacho114, line 63: Is it maybe better to combine class attributes into a single div element? (Each new element adds x2 lines of code).

e.g.

<div class="grid grid-cols-3 items-center p-2 pl-1 font-medium flex flex-row justify-start">
</div>

Nacho114 (commented):
Really nice work, the code is much more readable with all the components 🥇 

As @s314cy said, the i18n is really nice, and we'll figure out  next week a robust way to deal with using different code based on FeAI vs DeAI.

s314cy:
> Really nice work, the code is much more readable with all the components 1st_place_medal
> 
> As @s314cy said, the i18n is really nice, and we'll figure out next week a robust way to deal with using different code based on FeAI vs DeAI.

Exactly. For now I'll put all FeAI-specific backend files in a subfolder, so that DeAI still runs perfectly. For instance, the resulting code base would contain both `communication_manager/feai/{communication_manager, helpers}.js` and the usual `communication_manager/{communication_manager, helpers}.js`.

PR #164, opened 2021-12-02
MobilenetV2 implementation

laurislopata:
Implemented mobileNetv2, I would like to merge this so that I could test this with a  normal server, not having to use a local one, then I would have more resources to train with bigger image sets.


Nacho114 (approved):
Looks good; currently the hosted version is not up to date for testing the server -- will update once it is ready.

PR #165, opened 2021-12-03
Task creation form after merge

giordano-lucas:
Features :
1. `NewTaskCreationForm` component (tasks are uploaded to the server and displayed in real-time in `taskList`)
2. Add `Vue dev tools` config (see [chrome extension](https://chrome.google.com/webstore/detail/vuejs-devtools/ljjemllljcmogpfapbkkighbhhppjdbg?hl=en))
3. Clean `vuex` store. tasks frames are now saved exclusively in the store
4. Solved display bugs in `dark` theme in `TaskList.vue`
5. Further modularisation of `card` component. Now `TitleCard` is also available.


Nacho114 (dismissed):
- No error message when trying to add a task without a model. (It would be good to have an error message, in the same way as there is if we don't add a training rate for example)

This can be added in a future PR, not urgent.

giordano-lucas:
> * No error message when trying to add a task without a model. (It would be good to have an error message, in the same way as there is if we don't add a training rate for example)
> 
> This can be added in a future PR, not urgent.

I tried to solve this evening and event before when I was developing it. However, for some reason the `yup` validation is not possible for `file` input. This due to the fact that we cannot require the user to actually input the requested files. This means that we have to use something different than `yup` for this specific field validation

giordano-lucas:
@Nacho114 I merged the branch with `develop` it should be good now

Nacho114 (approved):
Server reviewed by @s314cy who gave thumbs up , UI changes look good too,

PR #172, opened 2021-12-14
Synchronous training scheme for federated training

ghost:
The client now goes through the following training process (during model.fit):
1. Connect to the server for a given task. Fetch the corresponding model.
2. Wait to get selected by the server then start the next training round. Selection only occurs after enough clients asked to get selected. This step ensures we don't start training alone and ditches AFK clients. See this as an activity ping from client to server.
3. Train locally until the end of the round. A round is made of several epochs. As of now, all these parameters are provided in the task's json.
4. Synchronize on round ends. Synchronizing means sending our local weights to the server, then waiting for the server to gather enough local weights to perform aggregation on its side. Once this has been done, we can fetch the aggregated weights and update our local model. This step ensures the model we get from the server was aggregated upon enough contributions.
5. Go to 2.

And server-side:
- Provide an API endpoint for selection status.
- Provide an API endpoint for aggregation status.
- Once enough weights were received from clients, override the task's served model's weights served. This is the aggregation step.
- Ensure we avoid any weird states for clients, or process API requests unwanted at reception time.

Many more small changes were made to further improve server code readability and modularity.
Note: This is a basic scheme serving as a baseline for improvement, as the one currently in production makes no sense. For instance, @giordano-lucas and I will be looking into asynchronous training soon enough.


s314cy:
I made this PR with the wrong Github account... See #173

PR #173, opened 2021-12-14
Synchronous training scheme for federated learning

s314cy:
The client now goes through the following training process (during model.fit):

1. Connect to the server for a given task. Fetch the corresponding model.
2. Wait to get selected by the server then start the next training round. Selection only occurs after enough clients asked to get selected. This step ensures we don't start training alone and ditches AFK clients. See this as an activity ping from client to server.
3. Train locally until the end of the round. A round is made of several local steps, for example one or several local epochs. As of now, all these parameters are provided in the task's json.
4. Synchronize on round ends. Synchronizing means sending our local weights to the server, then waiting for the server to gather enough local weights to perform aggregation on its side. Once this has been done, we can start the next round by fetching the aggregated weights and update our local model. This step ensures the model we get from the server was aggregated among enough contributions.
5. Go to 2.

And server-side:

- Provide an API endpoint for selection status.
- Provide an API endpoint for aggregation status.
- Once enough weights were received from clients, override the task's served model's weights served. This is the aggregation step. Increase the model (or round) counter, so we're always aware on which version of models the gradients have been computed, even in case they arrive late on the server
- Ensure we avoid any weird states for clients, or process API requests unwanted at reception time.

Many more small changes were made to further improve server code readability and modularity.
Note: This is a basic scheme serving as a baseline for improvement, as the one currently in production makes no sense. For instance, @giordano-lucas and I will be looking into asynchronous training soon enough.


mobile-browser-based-version/server/src/logic/federated/handlers.js
```diff
@@ -0,0 +1,547 @@
...
+ * to proceed to the pooling step.
+ */
+const AGGREGATION_THRESHOLD = 0.8;
+/**
+ * Number of selected clients required to start federated training.
+ */
+const TRAINING_THRESHOLD = 2;
+/**
+ * Once a certain threshold has been hit, leave a small window for late clients
+ * to join the next federated training session.
```
martinjaggi: time unit?

martinjaggi: and before we go to fully async, we can keep the rounds-based scheme here, but should also allow for late joiners (after several rounds), and some users to drop out (still ok as the threshold can still be reached with the remaining ones)

mobile-browser-based-version/server/src/logic/federated/handlers.js
```diff
@@ -0,0 +1,547 @@
...
+  AGGREGATION_STATUS: 'aggregation-status',
+  POST_WEIGHTS: 'post-weights',
+  GET_WEIGHTS: 'get-weights',
+  POST_SAMPLES: 'post-samples',
+  GET_SAMPLES: 'get-samples',
+  GET_TASKS: 'get-tasks',
+});
+/**
+ * Fraction of client gradients required on final round of federated training
+ * to proceed to the pooling step.
```
martinjaggi: this is in every round, right? and pooling means aggregation here i think?

mobile-browser-based-version/server/src/logic/federated/handlers.js
```diff
@@ -0,0 +1,547 @@
...
+ */
+const AGGREGATION_THRESHOLD = 0.8;
```
martinjaggi, line 28: can you give more info also on how the total number of clients is determined and when? (it might change later if someone joins or drops late). just to make clear what the fraction is of

mobile-browser-based-version/server/src/logic/federated/handlers.js
```diff
@@ -0,0 +1,547 @@
...
+/**
+ * Contains the model weights received from clients for a given task and round.
+ * Stored by task ID, round number and client ID.
+ */
+const weightsMap = new Map();
+/**
+ * Contains the number of data samples used for training by clients for a given
+ * task and round. Stored by task ID, round number and client ID.
+ */
+const dataSamplesMap = new Map();
```
martinjaggi: the name 'samples' is a bit confusing here. couldn't we make a map containing all `metadata` which clients send to server in the round (including model version which they received from server and computed gradients on (so that's the previous round number), number of samples and local steps done since then, and possibly other stuff so it's extensible?)

mobile-browser-based-version/server/src/logic/federated/handlers.js
```diff
@@ -0,0 +1,547 @@
...
+  response.status(200).send({ aggregated: aggregated });
+}
+
+async function _aggregateWeights(task, round, id) {
+  // TODO: check whether this actually works
+  const serializedAggregatedWeights = await averageWeights(
+    Array.from(weightsMap.get(task).get(round).values())
+  );
+  /**
+   * Save the newly aggregated model to local storage. This is now
```
martinjaggi: clarify this happens only on the server

mobile-browser-based-version/server/src/logic/federated/handlers.js
```diff
@@ -0,0 +1,547 @@
...
+  const modelFilesPath = config.SAVING_SCHEME.concat(
+    path.join(config.MODELS_DIR, task, 'model.json')
+  );
+  const model = await tf.loadLayersModel(modelFilesPath);
+  assignWeightsToModel(model, serializedAggregatedWeights);
+  model.save(path.dirname(modelFilesPath));
+  /**
+   * The round has completed training.
+   */
+  status.get(task).training = false;
```
martinjaggi: i wouldn't call this training false, but better round completed/pending. (as training is the whole process of FL, going on forever)

mobile-browser-based-version/server/src/logic/federated/handlers.js
```diff
@@ -0,0 +1,547 @@
...
+  /**
+   * Communicate the correct round to selected clients.
+   */
+  status.get(task).round += 1;
+  /**
+   * Start the next training round.
```
martinjaggi: just start next round

martinjaggi (changes requested):

mobile-browser-based-version/server/src/logic/federated/handlers.js
s314cy: definitely, I was struggling with the name and ended up choosing a bad one

mobile-browser-based-version/server/src/logic/federated/handlers.js
s314cy: the doc might be outdated here and there, this is the case here, I'll update it along with the other changes you requested!

mobile-browser-based-version/server/src/logic/federated/handlers.js
s314cy: exactly, pooling means aggregation and clients should this on the final epoch of a round only :)

mobile-browser-based-version/server/src/logic/federated/handlers.js
s314cy: @giordano-lucas

Nacho114 (dismissed):
- Changes requested by @martinjaggi implemented.
- Some issues remain, but will be addressed later, namely what happens on Server side, what if threshold is not satisfied?

PR #174, opened 2021-12-15
Add iFedAvg framework 

dgengler6:
**Features**

- Created a custom layer class _InteroperabilityLayer_ that implements the personnalization operations from David Rochewitz's paper.
- Training script can now train a model in a distributed manner, using the interoperability framework. (It takes a given model and wraps it between two InteroperabilityLayers)
- Create a class for each type of personalized model, that wraps a TF.js model inside an object and allows us to have methods for creation as well has getting the whole model or just the personalized part of it. 
- Add communication with the server to push your personal interoperability parameters to the server and query the aggregation parameters from all other clients. 
- Add Heatmaps for Interoperability Interpretation. 

**UI Change**

- Added a pannel where one can select Distributed Training Options. 
- Add a heatmap in the Training information panel.


dgengler6:
This should be woking now, I am waiting for Hugo's Synchronized training PR to be up to do the communication.

dgengler6:
I am switching this to Ready for review. 

There are still some changes I need to do in order to make it work correctly in the synchronous training, but it is WIP. 

I think you guys can already give a look at the code as most of it is here.

mobile-browser-based-version/src/helpers/training/training_manager.js
```diff
@@ -48,39 +49,44 @@ export class TrainingManager {
...
+        );
+        break;
     }
+    await this.model.init();
 
     // Continue local training from previous epoch checkpoint
-    if (this.model.getUserDefinedMetadata() === undefined) {
-      this.model.setUserDefinedMetadata({ epoch: 0 });
+    if (this.model.getModel().getUserDefinedMetadata().epoch === undefined) {
+      this.model.getModel().getUserDefinedMetadata().epoch = 0;
```
Nacho114: Is it possible to set this.model.getModel().getUserDefinedMetadata().epoch = 0 earlier, such that on creation it is guaranteed to be defined (then we can remove these 3 lines).

mobile-browser-based-version/src/helpers/training/training_manager.js
```diff
@@ -48,39 +49,44 @@ export class TrainingManager {
...
+      case modelDefinition.personalizationType.NONE:
+        this.model = new modelDefinition.Model(this.task, this.useIndexedDB);
+        break;
+
+      case modelDefinition.personalizationType.INTEROPERABILITY:
+        this.model = new modelDefinition.InteroperabilityModel(
+          this.task,
+          this.useIndexedDB
+        );
+        break;
```
Nacho114: It would be better to have a function in modelDefinition, an additional constructor, that takes in also personalizationType. and then returns the appropriate object.

mobile-browser-based-version/src/helpers/model_definition/model.js
```diff
@@ -0,0 +1,180 @@
...
+  }
+
+  /**
+   *  This is a getter for the part of the model we want to share with the network.
+   *  For personalizationType.NONE this is simply the model.
+   *  The behaviour depends on the personalization type of the model.
+   *  @returns the part of the model we want to share.
+   */
+  getSharedModel() {
+    return this.model;
```
Nacho114, line 125: It's confusing that both getModel and getSharedModel return the same object, when we want to use it. for someone new it might not be clear which to use. Maybe we should only keep getSharedModel? What is the reason to have 2, just for naming?

mobile-browser-based-version/src/helpers/training/training_manager.js
dgengler6: Yup though about this as well, gonna add this in model creation.

mobile-browser-based-version/src/helpers/training/training_manager.js
dgengler6: Yep that sounds good having a function out of the classes that return the right model ! good idea

mobile-browser-based-version/src/helpers/model_definition/model.js
dgengler6: Yeah that's the thing, this only the case for personalizationType.NONE. because here both the local and shared model are the model itself. You actually change this when redefining a new subclass.

Nacho114:
The work of iFedAvg needs to merged into the `experimental` branch, and so a separate PR is necessary.

PR #176, opened 2021-12-29
Stop training button & remove idle clients

### LINKED ISSUE #171: Properly handle disconnected clients
Make sure a client properly disconnects. This is currently not the case and can cause training error due to ghost clients preventing the threshold to be reached. 

  

- [x] Server-side, "if we haven't heard from a client, delete him" sort of policy.
- [x] Client-side, properly disconnect from server on tab reload/close.

### LINKED ISSUE #175: Stop training button
(1) Instead of running for a set number of epochs, run until the user presses a "stop training" button.

Solving this issue would also solve the following:
(2) The user should not be able to spam the "train {federated, decentralised}" button. Solving (1) would mean pressing the button again would stop (pause) training.

Further enhancement, requires (1) & (2) beforehand. I'll open another issue if enough people vote for this:
(3) Allow the user to enter their own desired number of training epochs, or a training time limit, through the UI.

s314cy:
Partially fixes #175 and fixes #171.

### Client-side
Additions:
* new button allowing the user to stop the training process, hence they are no longer able to spam the "start training button" no matter the kind of training performed (local/distributed)
* if training is performed locally, pressing the button effectively stops the training process and the user is able to resume to it any time with great reactivity
* if training is performed in a distributed fashion, effectively stopping the process remains to be implemented

The "beforeunload" event is used to handle client disconnection before unloading the page's resources, which corresponds to closing/refreshing a tab or moving to another website within the same tab. However, this event isn't made for performing asynchronous tasks such as web requests (e.g. disconnecting from a remote server), indeed resources might be unloaded before the asynchronous request is actually performed. This makes the event unreliable for our task. For instance, it seems to work just fine on Firefox when closing the client's tab, but does nothing when leaving the tab or moving to another website (includes refreshing the tab). Furthermore, listening to this event might cause browsers to turn off their [bfcache](https://web.dev/bfcache/) which is undesirable as explained [here](https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event). Finally, the event is [unreliable](https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event) on its own.

TL;DR: The client should not be trying to maintain the server's state of connected clients.

### Server-side
Additions:
* after each successful API request made by a client, leave it a certain delay to perform another successful API request before disconnecting it
* currently, more delay is given after a successful "connect" request, to give time to clients to configure their training options and inputs
* other API requests have a much shorter delay before considering the client as idle and thus disconnecting it


mobile-browser-based-version/src/components/building_frames/containers/TrainingFrame.vue
```diff
@@ -125,27 +139,63 @@ export default {
...
     goToTesting() {
       this.$router.push({
         path: 'testing',
       });
     },
-    async saveModelButton() {
+    async stopTraining() {
+      /**
+       * Stopping distributed training in a clean manner requires many
+       * additions and changes to make everything robust.
```
Nacho114: These changes are not w.r.t telling the server that we are leaving right? (Since the server is taking care of checking for non active users as you mention).

What is currently missing for it to be more robust?

s314cy: Right, even if _ideally_ it should still include a disconnection process.

What's missing is to interrupt all ongoing processes made by a call to model.fit. This is to prevent multiple model.fit calls from running simultaneously (typically when the user resumes training), which causes an error on the side of TF.js. It can easily happen as model.fit makes API calls to the server which sometimes are performed many times (selection + aggregation status). Hence, to make it robust, one should add code to fully stop (on pressing the "stop" button) every single ongoing process after a call to model.fit.

Another option would be for the stopping process to wait on all API calls to complete (either succeeding or failing), instead of actively trying to interrupt those calls. This is counter-intuitive from a user's POV, since pressing the "stop" button would not immediately stop the training process. Furthermore, in the case of waiting for a "selection status" call to finally complete, this would make the remote server select the client, and then this very client would directly disconnect as it is trying to stop the training process. Hence, it would grief the federated training overall.

This was much easier in the local setting, as all model.fit does here is gradient updates at the end of every epoch, so the "stop" button has very little lag.

s314cy: Oh and yes, all these struggles should not occur with the **asynchronous** version of federated training.

Nacho114: Yeah, waiting for the current round to finish instead of trying to stop on-going calls seems much cleaner from a dev POV.

Nacho114:
Minor comment
- Stop button should be read

Behavior comment
After pressing stop training, the only options is continue. 

The current stop button seems more like a pause button, since you can continue it after, and it does not bring you back to the train indiv or train distrib buttons. 

Ideally there should be 2 buttons, pause (which is the current functionality) and a stop button, that stops it and allows you the to train indiv or distrib.

martinjaggi:
i would put just one ideally, stop=`quit task`, as opposed to `join task` (reflecting the 1-bit status of a client relative to each task)

s314cy:
I'll go with a single button that stops the training and puts the user into the initial "train alone/distributed" state :)
I'll try and see if I can smoothly stop the distributed training too.

Nacho114:
Yeah that seems good, will merge once you change that.

Nacho114 (dismissed):
Changed STOP button to red to make it more intuitive, otherwise code + behavior is correct.

Nacho114 (approved):
Code looks good and behavior is correct.

PR #180, opened 2022-01-04
Reset on platform change

Nacho114:
### Make platform change more robust

Currently a platform change does not reset any components and this causes strange behaviors if training is still on-going while the platform is changed for example. To solve this this we do the following:

On platform change request, the user is asked to confirm platform change, and that if he continues progress might be lost if training is on-going. If the user confirms, the training components are refreshed and the user is brought to home.

Additional small features include saving the last used platform in cache as well as changing colors when the platform is changed.


Nacho114:
Side note:

By binding the MainTrainingFrame key to the store platform value, we force a refresh of said component whenever the platform changes. 

If the component was never mounted then nothing happens when we change platform.

If the component was mounted (e.g. the user trained), then the component will be refreshed, this means that when the user is brought to the home page after the refresh, it will be connected also to the appropriate server as during the refresh of the MainTrainingFrame, the client will be created and will ask to be connected to the appropriate server.

giordano-lucas (approved):
all good on my side

s314cy (approved):

PR #181, opened 2022-01-05
add DEV_TOOLS option in env.dev

giordano-lucas:
By changing the line 
```
DEV_TOOLS=disabled
```
in the `.env.development` file, you can enable the Vue dev tools.

This PR aims at removing the `404` error in the console due to the dev tools that were loaded by default previously.


Nacho114 (approved):
Great thanks! no more 404s :)

PR #182, opened 2022-01-06
Implemented batch wise preprocessing

laurislopata:
I tried to use Core API using the example given in their docs here: https://www.tensorflow.org/js/guide/train_models to be able to perform batch-wise image preprocessing and for the code to look cleaner but the problem with this is that optimizer.minimize is not an async function so when I would start the training loop the website would crash, which is obviously not ideal. That is why I used fitDataset and generator function to do implement the preprocessing


martinjaggi:
could you comment on the speed impact of this change on training performance, for the main tasks we have? does it have a memory impact too?

laurislopata:
In terms of memory, for CIFAR10 using the old method where the images would be resized all before training with my computer I could at most train with 1200-1300 images, now I can train at least up to 35000. In terms of speed

- For 500 images the current version was averaging 35.5s/epoch while the updated one 57.6s/epoch
- For 750 images the current version was averaging 67.6s/epoch while the updated one 96.6s/epoch

so the new version is about 35% slower but the memory improvement is very big. What I though we could do is implement a checker for the size of the dataset a user uploads, if it's big then the model would train using the new but slower version if it's small we could continue using the old but faster version. Do you think that would be a good and necessary improvement or should we just stick with the new version?

laurislopata:
Updated the comment about speed and memory implications of this change

martinjaggi:
ok maybe depends on batch size also? in the code it wasn't clear to me what batch size is going to be used.

laurislopata:
`    const trainData = tf.data.generator(
      trainDataGenerator(
        data,
        labels,
        trainingInformation))
        .batch(trainingInformation.batchSize);`

Batch size is set here, and for these experiments I used a batch size of 30

Nacho114:
> 

Looks good, btw you need to merge it with the latest develop.

I'd say to implement the size based approach that you mention, 35% is a lot and I think you can do it without too much overhead. Plus we know the old .fit was working well, so it'd be good to still have it. Can also discuss more tomorrow during the meeting.

martinjaggi:
sounds good. please also check that it works with other kinds of datasets and tasks (cifar looks good already)

laurislopata:
Okey, I'll implement your proposals until tomorrow's meeting

mobile-browser-based-version/src/helpers/training/training_manager.js
```diff
@@ -4,7 +4,43 @@ import {
...
+          tensor,
+          [trainingInformation.RESIZED_IMAGE_H,
+          trainingInformation.RESIZED_IMAGE_W]);
+      }
+      yield { xs: tensor, ys: tf.tensor(labels.arraySync()[i]) }
+    }
+  }
+}
+
+function validationDataGenerator(dataset, labels, trainingInformation) {
```
s314cy: You can make a single helper function for both the train & validation datasets, which would be called for two different sets of indexes. The function name could be something more specific, as it returns an image dataset. Also, functions should not be in a class file but rather in a separate helper file :)

Technically you could go even further and pass a lambda such as `(tensor) => tf.image.resizeBilinear(tensor, [...]])` to the helper function, to generalize its purpose :)

mobile-browser-based-version/src/helpers/training/training_manager.js
```diff
@@ -4,7 +4,43 @@ import {
...
+          [trainingInformation.RESIZED_IMAGE_H,
+          trainingInformation.RESIZED_IMAGE_W]);
+      }
+      yield { xs: tensor, ys: tf.tensor(labels.arraySync()[i]) }
+    }
+  }
+}
+
 const MANY_EPOCHS = 9999;
+const SIZE_THRESHOLD = 1e9; //1 GigaByte
```
s314cy: Constants should be declared on top of the file

mobile-browser-based-version/src/helpers/training/training_manager.js
```diff
@@ -144,12 +180,32 @@ export class TrainingManager {
...
 
-    await this.model.fit(this.data, this.labels, {
-      initialEpoch: this.model.getUserDefinedMetadata().epoch,
-      epochs: info.epochs ?? MANY_EPOCHS,
-      batchSize: info.batchSize,
-      validationSplit: info.validationSplit,
-      shuffle: true,
+    var dataSize = -1;
+
+    if (info.resize) {
```
s314cy: What happens if `info.resize` is not set? Will `dataSize` remain equal to -1 and "fast training mode" be the default?

mobile-browser-based-version/src/helpers/training/training_manager.js
```diff
@@ -144,12 +180,32 @@ export class TrainingManager {
...
+        info))
+      .batch(info.batchSize);
+    const valData = tf.data.generator(
+      validationDataGenerator(
+        this.data,
+        this.labels,
+        info))
+      .batch(info.batchSize);
+
+    await this.model.fitDataset(trainData, {
```
s314cy: Great!

mobile-browser-based-version/src/helpers/training/training_manager.js
```diff
@@ -179,32 +235,38 @@ export class TrainingManager {
...
-  async _trainDistributed() {
-    const info = this.task.trainingInformation;
+  else {
+    console.log("Fast training");
 
-    await this.model.fit(this.data, this.labels, {
-      initialEpoch: 0,
+    let resized_data = this.data;
+
+    if(info.resize) {
```
s314cy: don't forget to lint your files before pushing :) we still don't have any pre-commit hooks, so you'll have to manually use ESLint

s314cy (changes requested):

mobile-browser-based-version/src/helpers/training/training_manager.js
```diff
@@ -144,79 +150,263 @@ export class TrainingManager {
...
-            (logs.val_acc * 100).toFixed(2)
-          );
-          if (this.useIndexedDB) {
-            await updateWorkingModel(
-              this.task.taskID,
-              info.modelID,
-              this.model
+      });
+    } else {
+      console.log('Fast training');
```
martinjaggi: can you give more expressive names to the two modes? now it might be unclear later to people what 'fast' means here. batch-wise preprocessing/training? and in-memory training?  (now sometimes you call it memory-efficient, sometimes fast, a bit unclear which one is which)

mobile-browser-based-version/src/helpers/training/training_manager.js
```diff
@@ -144,79 +150,263 @@ export class TrainingManager {
...
+              this.stopTrainingRequested = false;
+            }
+          },
         },
-      },
-    });
-  }
+      });
+    } else {
+      console.log('Fast training');
```
martinjaggi: same comment as below on expressive naming

mobile-browser-based-version/src/helpers/training/training_manager.js
```diff
@@ -144,79 +150,263 @@ export class TrainingManager {
...
+      dataSize =
+        8 *
+        3 *
+        info.RESIZED_IMAGE_H *
+        info.RESIZED_IMAGE_W *
+        this.data.shape[0];
+    }
+
+    if (dataSize > SIZE_THRESHOLD) {
+      console.log('Memory Efficient training');
```
martinjaggi: batch-wise  preproc and train?

mobile-browser-based-version/src/helpers/training/training_manager.js
```diff
@@ -2,9 +2,15 @@ import {
...
   getWorkingModelMetadata,
+  imageDatasetGenerator,
 } from '../memory/helpers';
 
+import * as tf from '@tensorflow/tfjs';
+
+// Extra functions that are used to create Dataset object to be able to perform batch wise preprocessing of images.
+
 const MANY_EPOCHS = 9999;
+const SIZE_THRESHOLD = 1024 * 1024 * 1024; //1 GigaByte
```
martinjaggi: size of what? explain in a comment. maybe memory is a better name

mobile-browser-based-version/src/helpers/memory/helpers.js
```diff
@@ -138,3 +138,28 @@ export async function downloadSavedModel(taskID, modelName) {
...
 }
+/**
+ * Creates a dataset generator function for memory efficient training
+ * @param {Array} dataset the dataset of the task
+ * @param {Array} labels the labels of the task
+ * @param {Integer} startIndex staring index of the split
+ * @param {Integer} endIndex ending index of the split
+ * @param {Function} transformation transformation function to be applied to the data
+ */
+export function imageDatasetGenerator(
```
martinjaggi: this one should not be image specific, but can work for all datasets, tabular data etc. 
also state what assumptions you made on the input (which dimensions are datapoint, or others such as channels, h,v, etc)

mobile-browser-based-version/src/helpers/training/training_manager.js
```diff
@@ -144,79 +150,263 @@ export class TrainingManager {
...
-          this.trainingInformant.updateGraph(
-            epoch + 1,
-            (logs.val_acc * 100).toFixed(2),
-            (logs.acc * 100).toFixed(2)
-          );
-          console.log(
-            `EPOCH (${epoch + 1}):
+    var dataSize = -1;
+
+    // If info resize is not defined it will default to Fast Training
```
martinjaggi: what is the semantics of 'resize'? this seems to be in a code part which is general, not image-specific, so a bit surprising to see the 8*3*... hardcoded below

mobile-browser-based-version/src/helpers/training/training_manager.js
```diff
@@ -144,79 +150,263 @@ export class TrainingManager {
...
+          },
         },
-      },
-    });
-  }
+      });
+    } else {
+      console.log('Fast training');
+
+      let resized_data = this.data;
```
martinjaggi: resizing is a bit too specific, resizing can be just part of some general preprocessing anyway (no matter if batch-wise or done globally before training)?

martinjaggi (changes requested):

mobile-browser-based-version/src/helpers/training/training_manager.js
laurislopata: The only problem I see is that in general preporcessing for tabular datasets will not make the size of the dataset much bigger so defaulting those the training mode that is faster is a satisfactory solution?

mobile-browser-based-version/src/helpers/memory/helpers.js
laurislopata: I agree, since at the moment we don't have any preprocessing pipelines for the other tasks besides the CIFAR10 maybe a good solutions would be to have create another helper function that would take in a lambda and apply preprocessing on the entire dataset and when in the future other preprocessing functions are implemented for other tasks they could be passed as an argument for that lambda. Do you think that would be a good solution?

mobile-browser-based-version/src/helpers/training/training_manager.js
laurislopata: So in that case calculating the data size only for images should be fine?

martinjaggi: not sure what 'resize' means here. do you mean batch size? also the overall dataset size (memory used) is rougly the same anyway, no matter if batching is turned on or off? (or do you mean to store the results of the preprocessing, or compute on the fly?). please clarify in the code comments once you have time. not urgent of course

mobile-browser-based-version/src/helpers/memory/helpers.js
martinjaggi: yes that would be ideal. i think it's already a bit like this if you look at the preprocessing selection option in the task creation form?

Nacho114: this is general right? -> if yes we should just call this datasetGenerator?

martinjaggi:
please try to name things differently than 'fast training' (meaning 'in memory'?) and 'memory training'. can say batch-wise and explain the differences as to when the preprocessing is applied or not. batching code, preprocessing code and training code should be well separated. 

is there as much new code needed in `training_manager` or could some be avoided?

`imageDatasetGenerator` does not seem to need to know the start/end index and split (and can be independent of data modality), it seems to just apply a preprocessing (lambda) function, which could be applied per batch anyway, no matter for what data modality

laurislopata:
Yes, I am currently working on addressing these issues, will push soon

laurislopata:
Implemented the discussed version. Now each task has an extra field in its .json file  ` "preprocessFunctions": [],`, for 3 of them it's empty and for cifar10 it has the "resize" string in it, based on it I construct a preprocessing function, also for CIFAR there's and extra field ` "batchwise": true`, this sets how to apply the preprocessing(by default it's on the entire dataset before training). Also clarified the comments and cleaned up the code.

Nacho114:
> Implemented the discussed version. Now each task has an extra field in its .json file ` "preprocessFunctions": [],`, for 3 of them it's empty and for cifar10 it has the "resize" string in it, based on it I construct a preprocessing function, also for CIFAR there's and extra field ` "batchwise": true`, this sets how to apply the preprocessing(by default it's on the entire dataset before training). Also clarified the comments and cleaned up the code.

It's looking much better. 

Question: Is it necessary to include "resize" as a tag? 

It seems that every task has an IMAGE_H and RESIZED_IMAGE_H, if these don't equal doesn't it mean that we want to resize? (So the resize would be implicit in these values being equal or not).

mobile-browser-based-version/server/src/logic/federated/handlers.js
```diff
@@ -50,7 +50,7 @@ const IDLE_DELAY = 1000 * 10;
  * Same as IDLE_DELAY, except longer for clients that recently connected and thus
  * are not critical nodes (i.e. training nodes).
  */
-const LONGER_IDLE_DELAY = 1000 * 60;
+const LONGER_IDLE_DELAY = 1000 * 600;
```
Nacho114: @laurislopata @s314cy 

Is this only used for "clients that recently connected and thus are not critical nodes"?, if yes let's rename it to NEW_CLIENT_IDLE_DELAY.

mobile-browser-based-version/src/helpers/memory/helpers.js
```diff
@@ -138,3 +138,45 @@ export async function downloadSavedModel(taskID, modelName) {
...
 }
+/**
+ * Preprocesses the data based on the training information
+ * @param {Array} data the dataset of the task
+ * @param {Object} trainingInformation the training information of the task
+ */
+export function preprocessData(data, trainingInformation) {
+  var tensor = data;
+  //More preprocessing functions can be added using this template
+  if (trainingInformation.preprocessFunctions.includes('resize')) {
```
Nacho114: How to improve this:

Create a function in helpers, just above this maybe and call it withImageResize(arg) (or something simple that scribes what it does), where arg is trainingInformation.preprocessFunctions, and it's output is true when we want to resize and image. This makes it more legible, but more importantly, as I note in my other [comment](https://github.com/epfml/DeAI/pull/182#issuecomment-1016254376), the resize might be redundant, and so an equality check might be better, in this case the readability gain is even bigger!

mobile-browser-based-version/src/helpers/training/training_manager.js
```diff
@@ -144,79 +148,215 @@ export class TrainingManager {
...
-      callbacks: {
-        onEpochEnd: async (epoch, logs) => {
-          this.trainingInformant.updateGraph(
-            epoch + 1,
-            (logs.val_acc * 100).toFixed(2),
-            (logs.acc * 100).toFixed(2)
-          );
-          console.log(
-            `EPOCH (${epoch + 1}):
+    if (info.batchwise) {
```
Nacho114: info.batchwise, what does this represent? It's not clear to me at a first glance (might be worth making the name longer and explicit for legibility)

mobile-browser-based-version/src/helpers/training/training_manager.js
```diff
@@ -144,79 +148,215 @@ export class TrainingManager {
...
+        },
+      });
+    } else {
+      console.log(
+        'Fast training mode is used, data preprocessing is executed on the entire dataset at once'
+      );
+
+      const tensor = preprocessData(this.data, info);
+
+      await this.model.fit(tensor, this.labels, {
```
Nacho114: There is an if else here, where each does model fitting in a different way. They are each quite long, could you make a function for each:

if (info.batchwise) {
  modelFitDataBatchWise(...)
} else {
  modelFitData(....)
}

Something along these lines (maybe you can find better names).

mobile-browser-based-version/src/task_definition/image_task.js
```diff
@@ -17,7 +17,8 @@ export class ImageTask extends Task {
       img.width = this.trainingInformation.IMAGE_W;
       img.height = this.trainingInformation.IMAGE_H;
       img.onload = () => {
-        var output = tf.image.resizeBilinear(tf.browser.fromPixels(img), [this.trainingInformation.RESIZED_IMAGE_H, this.trainingInformation.RESIZED_IMAGE_W]);
+        var output = tf.browser.fromPixels(img)
+        // var output = tf.image.resizeBilinear(tf.browser.fromPixels(img), [this.trainingInformation.RESIZED_IMAGE_H, this.trainingInformation.RESIZED_IMAGE_W]);
```
Nacho114: if the comment is for testing you can delete it once your done

laurislopata:
@Nacho114 Implemented your proposals for the code to look cleaner, you can have a look

mobile-browser-based-version/src/helpers/training/training_manager.js
```diff
@@ -145,105 +145,115 @@ export class TrainingManager {
...
-          onEpochEnd: async (epoch, logs) => {
-            this.trainingInformant.updateGraph(
-              epoch + 1,
-              (logs.val_acc * 100).toFixed(2),
-              (logs.acc * 100).toFixed(2)
-            );
-            console.log(
-              `EPOCH (${epoch + 1}):
+      await this._modelFitDataBatchWise(this.model, info, {
+        onEpochEnd: async (epoch, logs) => {
```
Nacho114: The onEpochEnd is the same for ._modelFitDataBatchWis and .modelFit, you could turn the onEpochEnd and make a function and pass it to both to reduce code duplication.

mobile-browser-based-version/src/helpers/training/training_manager.js
```diff
@@ -252,109 +262,72 @@ export class TrainingManager {
...
+        onTrainBegin: async (logs) => {
+          await this._onTrainBegin();
+        },
+        onTrainEnd: async (logs) => {
+          await this._onTrainEnd();
+        },
+        onEpochBegin: async (epoch, logs) => {
+          await this._onEpochBegin(epoch);
+        },
+        onEpochEnd: async (epoch, logs) => {
```
Nacho114: same with the decentralised one (with the onEpochEnd being the same).

mobile-browser-based-version/src/helpers/memory/helpers.js
```diff
@@ -146,7 +150,7 @@ export async function downloadSavedModel(taskID, modelName) {
 export function preprocessData(data, trainingInformation) {
   var tensor = data;
   //More preprocessing functions can be added using this template
-  if (trainingInformation.preprocessFunctions.includes('resize')) {
+  if (_withResize(trainingInformation)) {
```
Nacho114: nice, if you can call it _preprocessDataWithResize that would be better.

Nacho114:
If you can also add comments to all the functions that would be great :) you can use the same template as the other commented ones. I think after this we can merge.

Nacho114:
Requested changes have been implemented, all seems to work.

- Will further review the question of having both .fit and .fitdata

PR #183, opened 2022-01-12
Fixes #179 & avoid blocking state

### LINKED ISSUE #179: Model update on late join
In case a user joins federated training while round R<sub>T</sub> is still processing, it should ensure it's got the latest available model (provided by the server) on starting R<sub>T+1</sub>. It is currently not checked as the late user does not enter any "wait for weights aggregation" step.

Note that solving this issue would prevent users from joining federated training with their own local model on their first round. If this is something desired, then this issue can be directly closed.

s314cy:
Ensures clients don't remain blocked in the aggregation phase, in case other clients disconnect before posting their necessary weights for the current round to complete. To do this, asking the server whether the round's model was aggregated can now potentially trigger this very aggregation.

Fixes #179: here is a brief overview of what's going on, where client 3 is "late".  are the selection, training and aggregation phases, respectively, for round . During selection phase, a client periodically asks the server to get selected for the next upcoming round. Then, each client trains locally for a few epochs. Ultimately, selected clients enter aggregation phase during which they periodically ask the server whether the round's model was aggregated yet. Additionally, with this PR, client 3 was "late" and thus not selected, so it directly goes from  to  to ensure it gets the latest model.

image


martinjaggi:
looks, ok, but i wonder if the selection phase could be simplified more, taking a slightly more server-centric view? the server could just send a "you are selected" signal to the ones it thinks are alive and selected? (and then sends the model belonging to the round?). anyway i guess it's a push vs pull type of question. 

and yes, received model updates which were computed based on an older round can simply be dropped on the server. later we'll slightly relax this when you add the nice buffer based async variant

mobile-browser-based-version/server/src/logic/federated/handlers.js
```diff
@@ -439,47 +450,47 @@ export function postWeights(request, response) {
...
+    /**
+     * If aggregation occured, make the client wait to get selected for next round so
+     * it can proceed to other jobs. Does nothing if this is a late client.
+     */
     selectedClients.get(task).delete(id);
+  } else if (
+    /**
+     * To avoid any blocking state due to the disconnection of selected clients, allow
+     * this request to perform aggregation. This is merely a safeguard. Ideally, this
+     * should obviously performed by the `postWeights` request handler directly, to
```
Nacho114: Typo: should obviously be performed

mobile-browser-based-version/server/src/logic/federated/handlers.js
```diff
@@ -439,47 +450,47 @@ export function postWeights(request, response) {
...
+  if (!clients.has(id)) {
+    return _failRequest(response, type, 401);
+  }
+  if (!tasksStatus.has(task)) {
+    return _failRequest(response, type, 404);
   }
-  /**
-   * No weight was posted for this task's round.
-   */
   if (!(weightsMap.has(task) && weightsMap.get(task).has(round))) {
```
s314cy: should return aggregated=false instead of 404, otherwise late clients might get disconnected (in addition to receiving weird 404 errors although everything is normal)

mobile-browser-based-version/server/src/logic/federated/handlers.js
```diff
@@ -269,7 +269,6 @@ export function selectionStatus(request, response) {
   if (!clients.has(id)) {
     return _failRequest(response, type, 401);
   }
-
   if (!tasksStatus.has(task)) {
     return _failRequest(response, type, 404);
```
Nacho114, line 273: For the errors, should we maybe add an error.js file, and create variables there:
- Make errors easier to understand in code + also in log output

Nacho114 (changes requested):
Once the "should return aggregated=false instead of 404," is resolved we can merge.

Nacho114:
> looks, ok, but i wonder if the selection phase could be simplified more, taking a slightly more server-centric view? the server could just send a "you are selected" signal to the ones it thinks are alive and selected? (and then sends the model belonging to the round?). anyway i guess it's a push vs pull type of question.

Yeah I agree, I feel like this might simplify the code, it would be great if @s314cy could discuss the current implementation in more detail in one of the next meetings, we could then brainstorm the most simple / elegant solution.

PR #193, opened 2022-01-25
Typescript migration

Nacho114:
#### Why?
- Makes code easier to read.
- Makes code easier to debug.
- Makes code easier to refactor.
- ...

#### Is it ready?
- Everything is working for titanic (DeAI/FeAI/ train alone)
- New features should branch from this one in order to test more TS, plus we will once everything is working

#### TODO:
- [x] Cifar-10 is currently not working, the training always gives 0 acc, there is an issue with the batchwise training. 
- [x] More testing
- [x] Onboarding documentation / Readme


martinjaggi:
i'm a big fan of typed style, but maybe there is a slight downside of full typescript: it breaks compatibility for people who don't know how to type an object (which can happen in deep nets), and also it introduces a mandatory compilation step.

instead i hear that this similar but slightly softer approach would give the best of both worlds:
https://fettblog.eu/typescript-jsdoc-superpowers/

meaning we should annotate all functions with `jsdoc`, so that then VS code actually shows the typing of each function during developping, which is very convenient

and we can still use the `typescript-cli` to lint all our stuff anyway (even for `.js` files), and to actually do the type checks (which is the main thing you're trying to achieve i think)

any thoughts?

s314cy:
I've worked on TypeScript projects in the past and it definitely helps with code readability for newcomers. Technically,
any vanilla JavaScript file is valid TypeScript as the latter strictly supersets the former. Furthermore, static typing is optional.
So, unless I'm missing something, one wouldn't be too much bothered by having to type TF.js objects.

I have little JSDoc experience with the few already annotated methods and functions throughout DeAI and it's great, as
VSCode indicates annotated types on top of variables. However, we all know how docs are easily omitted, and how annotating
a variable declared within the method/function's body would not be done 95% of the time.

As for the compilation step, one would have to check how long it takes. Otherwise, it is actually quite desirable to detect
errors on compilation rather than running time. It also ensures the provided typing typechecks, contrary to JSDoc.

Again, I am certainly not an expert on this subject so feel free to share your knowledge @giordano-lucas @Nacho114 :stuck_out_tongue: 

Note: Ideally, a better compromise would be something similar to what [Python proposes](https://docs.python.org/3/library/typing.html)

Nacho114:
Sorry I missed this thread! I need to activate the email notifications.

"it breaks compatibility for people who don't know how to type an object (which can happen in deep nets), and also it introduces a mandatory compilation step."
-> As I understand it, other people would only add their models through our task creation form, and I think the only constraint is that the input be a valid TF model, so not different from now. 

I think I have everything working now on TS, and in order to make it work I found a a bunch of bugs on the way due to JS being too relaxed (e.g. having a function called with 4 params even though it only accepts 3, using === between string and number for creating hot encodings, which would break silently if the labels included non-numbers, ... ). 

As a side note, js-doc is no longer active[https://github.com/jsdoc/jsdoc/blob/main/CHANGES.md] since 2020, I think the fact that ts is so popular makes it a big plus since we are likely to find resources if we find bugs.

After skimming around I get a feeling ts is overall better except for the initial cost of doing it, it's still flexible since you can choose how strict the compiler is and such things. ts was done to be a type safe js, and js-doc was mostly meant for documentation with added type checks.

I am also new to this so I cannot give a compelling argument, I would need to spend some more time reading/understanding.

Some links I read:

https://stackoverflow.com/questions/68421854/jsdoc-ide-vs-typescript

Nacho114:
> I've worked on TypeScript projects in the past and it definitely helps with code readability for newcomers. Technically, any vanilla JavaScript file is valid TypeScript as the latter strictly supersets the former. Furthermore, static typing is optional. So, unless I'm missing something, one wouldn't be too much bothered by having to type TF.js objects.
> 
> I have little JSDoc experience with the few already annotated methods and functions throughout DeAI and it's great, as VSCode indicates annotated types on top of variables. However, we all know how docs are easily omitted, and how annotating a variable declared within the method/function's body would not be done 95% of the time.
> 
> As for the compilation step, one would have to check how long it takes. Otherwise, it is actually quite desirable to detect errors on compilation rather than running time. It also ensures the provided typing typechecks, contrary to JSDoc.
> 
> Again, I am certainly not an expert on this subject so feel free to share your knowledge @giordano-lucas @Nacho114 stuck_out_tongue
> 
> Note: Ideally, a better compromise would be something similar to what [Python proposes](https://docs.python.org/3/library/typing.html)

I don't see any compilation time difference compared to pure js for deai on my side.

giordano-lucas (dismissed):

giordano-lucas:
@martinjaggi 

> i'm a big fan of typed style, but maybe there is a slight downside of full typescript: it breaks compatibility for people who don't know how to type an object (which can happen in deep nets), and also it introduces a mandatory compilation step.
> 

1.  I don't think that is going to be an issue here. Typescript offers the possibility to use types **NOT** the obligation. We can always convert any javascript file in typescript simply by changing  `.js` into `.ts` ( + possibly minor changes like removing semicolons). 
2. I also don't agree with you statement regarding compilation. Indeed, we are already using a mandatory compilation step through `Babel` in classical javascript. Going from the `Babel` compiler to the `Typescript` compiler is not going to add any overhead. 

Personally, like @Nacho114 and @s314cy, I am a big fan of typescript. There is a much better support for typescript than js-doc and it is more readable in my opinion. Given the fact that @Nacho114 already did all the work in this PR, it's going to be super easy to migrate.






Furthermore, the DeAI developper team is the main contributor to the platform anyway. We are the one who are going to maintain the platform and make use of the typescript features. The only part where external people are going to introduce code to the platform is through preprocessing steps.

martinjaggi (dismissed):

PR #199, opened 2022-02-04
Update and addition of code documentation

s314cy:
This PR contains docs I'm updating/adding to the code I've written throughout the semester.


s314cy:
Considering the client and server changes brought by asynchronous federated training, only the docs for `logic/memory` should be kept. Do you agree @Nacho114? If so, I'll quickly merge this.

Nacho114:
> Considering the client and server changes brought by asynchronous federated training, only the docs for `logic/memory` should be kept. Do you agree @Nacho114? If so, I'll quickly merge this.

Yeah sounds good (e.g. `_getModelMetadata`?)

PR #200, opened 2022-02-08
Frontend backend separation

### LINKED ISSUE #155: Cleaning logic and architecture
Cleaning the logic and architecture will give us:

- Scalability (having organisations use it reliably)
- Maintainability (easier for students to add features)
- Will make some enhancement easier, such as [Headless task](https://github.com/epfml/DeAI/issues/154)

In the UI side there is also cleaning to be done:

- Add more info to the home page
- Simplify the training page (remove a lot of redundancies like the log, keep it but with a drop down setting).

giordano-lucas:
⚠️ @Nacho114 @dgengler6  this will require your attention in your current PR as it involves changes in pretty much all files again. We should coordinate our efforts in order to minimise the amount of extra work. I already had to basically recode everything I did before the exams because my branch had diverged too much which was super time-consuming. I am sure you do not have the same situation.

# Description of the contributions

## Modularisation and refactoring of `NotFound.vue` and the `vue` files inside the `components/upload` directory. Indeed they were lacking modularity and the code was hard to read and debug (especially for the upload frames).

This refactoring also allowed me to solve different upload files and testing bugs mentioned in earlier issues.

# Standardise  notations in Vue files (following Vue standards)

1. Uppercase props:
     a. `Task` => `task`.
     b. `Id` => `id`.
2. 1. Reactive hooks: 
     a.  `v-on:` becomes `@`. E.g. `v-on:click="() => trainModel(titanic)"` becomes  `@click="() => trainModel(titanic)"`.
     b.  v-bind: becomes `:`. E.g. `v-bind:task="task"` becomes  `:task="task"`.

# Restyle diagrams of @Nacho114 

The container divs were not properly spaced and did not meet the platform requirements, design-wise.

See the screenshot below for the updated version

Screenshot 2022-02-08 at 00 33 28

# Separate front end backend code :

## Intro
This PR aims to prepare the way for the introduction of the CLI feature. The only step that will be left to do is to define a separated memory system that works with `node.js` (i.e. indexedBD is not available).

## Implementation details
In order to achieve a complete separation between interface and implementation, we need to define a series of JavaScript classes abstracting the behaviour currently inside the `Vue` files through a programming interface in the form of methods. These methods will then be called by the `Vue` component at the appropriate time to respect the specifications of the platform. 

To be consistent with the architecture proposed in the earlier modularisation steps, a `Trainer` and a `Tester` classes are defined. `TrainingFrame` component will instantiate `Trainer` object to handle the training process and similarly for testing. Both `Trainer` and  `Tester` are subclasses of an abstract class named  `Actor`. 

To obtain a correct implementation, the `Trainer` and `Tester` objects must have access to the processing functions of the tasks they are trained/tested on. Previously, these functions were defined in the corresponding `Vue` component (e.g. `CSVTrainingFrame`) and then passed as props to their child `TrainingFrame`/`TestingFrame`. This behaviour is not possible anymore since we aim to be separated from the UI. The solution is to define a specific helper per task. In the code, this is referred to as the `TaskHelper` abstract class. Each subclass must implement the following functions: 

1. `dataPreprocessing`: wrapper for the underlying `task.preprocessing` function.
2. `makePredictions`:  wrapper for the underlying `task.predict` function.
3. `predictionsToCsv: converts the predictions of the machine learning model into a valid CSV format.

The new architecture is summarised in the following figure:
Screenshot 2022-02-08 at 01 32 44

> **Note**: that this particular abstraction will greatly simplify the process of allowing the users of the platform to specify preprocessing `JavaScript` routines directly via the task creation form and therefore propose new data types.

# Last aspects to work on

This is quite a large PR so I thought it would be interesting to give some time to process it and hopefully give me some feedback and/or find bugs.

1. I already noticed that the training of `MNIST` has some issues. I don't really know what the problem is yet but I suspect it is again related to the weird relationship between two-way binding for reactivity in Vue and TFLS. I already had a similar problem with the Titanic task and it was not that simple to solve.
2. I will probably add more documentation into the `helpers/task_definition/README.md` file to explain what code should be added to create a new Datatype Task.
3. I think that there could be some confusion related to the use of the task word in the platform. Indeed, it is currently used both for a task such as `MNIST` (e.g. in the `tasks.json` file in the server) and a data type like `image` or `csv` (e.g. we have `task.js` and `TaskFrame.vue`).  I think it would be good to use a clear different terminology because we can have different tasks that have the same underlying task datatype.
4. The `ImageTestingFrame.vue` is completely outdated and needs to be refactored (similarly to the `containers/upload` files).


Nacho114:
Thanks for the very well documented PR!

I like the abstraction of the actor, but the name is too general so it's not clear what it does. I read an article not too long ago about this, how names should be succinct but also informative. For instance it's tempting to call something utils or helpers, but it's not clear at all what such a package would do. 

Instead of Actor, what about ModelActor? (or if you have another proposal).

Nacho114:
Most things seems to be working, currently debugging cifar-10, data loader is not working properly.

Nacho114:
Tested feai / deai, everything seems work good now. The separation has the right abstraction.

Nacho114 (approved):
Very good PR! A bit big, but the added changes are great for separations of concerns among modules. CLI + other types of tests will be easier to do now.

PR #206, opened 2022-02-15
Simulation scripts

laurislopata:
Wrote some doccumentation  inside the python files and updated the readme to show how well our platform performs and also some of limits.


Nacho114:
If you could add some documentation in the README of the structure of the scripts, a reference to selenium and how to use your scripts (run / change / etc)

laurislopata:
I added a short description in the README and have a more extensive documentation of what each function and constant does in the script files. Should I add all of that to the README as well?

Nacho114:
> I added a short description in the README and have a more extensive documentation of what each function and constant does in the script files. Should I add all of that to the README as well?

No need to copy comments in the scripts onto the README. It would however be nice if you describe what exactly it does:

- Opens n tabs
- In each tab it:
- Loads data
- trains
- logs performance

something along these lines. Basically to have an overview.

PR #215, opened 2022-02-23
Async federated training

### LINKED ISSUE #161: Async / Sync training for FeAI
Async / Sync training for FeAI

Nacho114:
We introduce async training scheme to the federated setting:

The main class is the `AsyncWeightHolder`, this contains a buffer of currently loaded weights.

It works as follows:
The current model has an associate version.
When a user posts weights, these are added to the buffer only if they were trained from the current version (we can easily relax this to allow weights from versions that are not older than n iterations of it). Once the buffer reaches capacity, we average the weights in the buffer, update the latest model and increment the version by one.

We also added several tests for the class and corresponding server requests.

TODO: 
- [x] Renamed version to round


Nacho114:
Did some refactoring of the helper classes/folders to make it more intelligible what they do, and refactored accordingly s.t. each file unit has a theme of functions that are similar in what they do

Nacho114:
Async is now default for federated since the sync is buggy, we can later add the option sync/async to tasks for example as suggested by @martinjaggi or in the UI.

When testing on e.g. titanic, the output in the server console is as follows:

```
************************************************************
-> Adding new weights of LXyFxBoyQl to buffer.
-> Adding new weights of leKwHW1H5G to buffer.

Buffer is full; Aggregating weights and starting round: 1

-> Adding new weights of LXyFxBoyQl to buffer.
	Updating weights of LXyFxBoyQl to buffer.
	Updating weights of LXyFxBoyQl to buffer.
	Updating weights of LXyFxBoyQl to buffer.
-> Adding new weights of leKwHW1H5G to buffer.

************************************************************
Buffer is full; Aggregating weights and starting round: 2


```

-> `Adding new weights of LXyFxBoyQl to buffer. `
This indicates that `LXyFxBoyQl` just contributed to the current round for the first time, if this is followed by:
`Updating weights of LXyFxBoyQl to buffer.`
This means that before the round ended said user pushed a new model, we proceed to update his model in buffer (the buffer is a map of user to weight).

Once the buffer is full we get:
`Buffer is full; Aggregating weights and starting round: 2`

Note that after each round the users get the latest round number, if this exceeds their own round number, than they pull the latest model from the server as this means a new aggregation phase took place.

mobile-browser-based-version/server/src/logic/federated/async_weights_holder.ts
```diff
@@ -0,0 +1,85 @@
...
+ */
+export class AsyncWeightsHolder {
+    taskID: string;
+    bufferCapacity: number
+    buffer: Map<string, any>;
+    round: number;
+    _aggregateAndStoreWeights: (weights: any) => Promise<void>;
+
+    constructor (taskId: string, bufferCapacity: number, aggregateAndStoreWeights: (weights: any) => Promise<void>) {
+      this.taskID = taskId
```
s314cy: taskID

mobile-browser-based-version/server/src/logic/federated/async_weights_holder.ts
```diff
@@ -0,0 +1,85 @@
...
+      this.bufferCapacity = bufferCapacity
+      this.buffer = new Map<string, any>()
+      this._aggregateAndStoreWeights = aggregateAndStoreWeights
+      this.round = 0
+    }
+
+    _resetBuffer () {
+      this.buffer = new Map<string, any>()
```
s314cy: `this.buffer.clear()`

mobile-browser-based-version/server/src/logic/federated/async_weights_holder.ts
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    /**
+     * Add weights originating from weights of a given round.
+     * Only add to buffer if the given round is not old.
+     * @param weights
+     * @param round
+     * @returns true if weights were added, and false otherwise
+     */
+    async add (id: string, weights: number, round: number): Promise<boolean> {
```
s314cy, line 73: LGTM

mobile-browser-based-version/server/src/logic/federated/async_weights_holder.ts
```diff
@@ -0,0 +1,85 @@
...
+
+    _updateRound () {
+      this.round += 1
+    }
+
+    _getWeightsFromBuffer () {
+      return Array.from(this.buffer.values())
+    }
+
+    async _updateWeightsIfBufferIsFull () {
```
s314cy, line 52: LGTM

mobile-browser-based-version/src/logic/communication/federated/client_async.ts
```diff
@@ -0,0 +1,140 @@
...
+    if (localRoundIsOld) {
+      // update local round
+      // TODO need to check that update method did not fail!
+      this.round = serverRound
+      // update local model from server
+      this._updateLocalModelWithMostRecentServerModel()
+    }
+  }
+
+  async onEpochEndCommunication (model, epoch, trainingInformant) {
```
s314cy, line 129: scheme seems correct to me

mobile-browser-based-version/src/logic/communication/federated/client_async.ts
```diff
@@ -0,0 +1,140 @@
...
+  }
+
+  _updateLocalModelWithMostRecentServerModel () {
+  // TODO: naming is not good, seems like you are just creating + only do so if it is more recent.
+  //! !!!!
+    this.task.createModel()
+    console.log('Updated local model')
+  }
+
+  async onEpochBeginCommunication (model, epoch, trainingInformant) {
```
s314cy, line 108: why duplicate the `_update` call from `onEpochEndCommunication`? the two callbacks are triggered very close in time, so I don't think it would be beneficial to make more frequent calls to `_update` in this way

s314cy (changes requested):
the async scheme looks good to me (except a few tweaks) & nice refactor :)

mobile-browser-based-version/src/logic/communication/federated/client_async.ts
Nacho114: I basically tried to copy what was in the sync client. Does it make more sense there to update twice? 
For the async, is it better than to just keep it at epoch begin? (since we need to load the weights, or maybe this is already done before?)

s314cy (approved):

martinjaggi:
excellent work. just some cosmetics for later:
- rename `holder` to `buffer`
- mark the old weights as sth like _weights update based on round x_
- add a time-cutoff which can be different from one. so if then if _x < now - cutoff_ discard them? (you already added a comment)
- clarify the logic documentation a bit at the main parts (in the buffer there can be max one weight/parameter per client.
- check if the round number is already attached as meta-date for each model received anyway? (we had this for federated statistics on the server).
- remove the non-async scheme, then code becomes easier to read, and no switches needed

Nacho114:
thanks, good suggestions, some questions:

- Currently we do not store old weights, in a current round if a new user updates new weights we override them. We could store the old ones, but I think we risk getting in memory troubles. 
- If we want to have the cryptographic stack that Vincent worked I think we need to have a sync version of federated. Although maybe it can be modified for async, but I'm not sure.

martinjaggi:
- no need to store old weights. only the freshest per worker makes sense

- for secure MPC, the buffer will be compatible (there is a recent facebook paper). so it's ok to fully drop sync

martinjaggi:
relevant paper, just for future reference:
Federated Learning with Buffered Asynchronous Aggregation
https://arxiv.org/abs/2106.06639

Issue #191, opened 2022-01-20
Re-working the adding data / data check code.


Nacho114:
Working on issue [188](https://github.com/epfml/DeAI/issues/188#issue-1107313713) was tricky and no clean solution was possible without making more significant code changes. The adding data / data check is functional, but is not very flexible code wise, and it's the last major component of DeAI that needs to be re-worked (as was done for the server / training manager etc). 

Doing this would make it much easier in the future to add features and fix bugs in the adding data side, the current state makes it hard to work with in my opinion.

PR #216, opened 2022-02-24
208 fractional epochs

### LINKED ISSUE #208: support fractional epochs, and callbacks on batch level
for large datasets, communication needs to be more frequently than once per epoch.

TF.js supports callbacks on the batch level
`callbacks.onBatchEnd`

we can use these to provide a callback for us on begin / end of round. a round always is at least one batch, and its size can be easily computed from epochs/rounds and bachsize in the task definition. this way we can allow a round to be any fraction of less or more than an epoch, suitable to get reasonable communication times both for large and small local data

### LINKED ISSUE #210: cleanup: remove 'Fast training mode', and make batch-wise default
simplify `training_manager.ts` to only offer batch-wise training.

so ` _modelFitData` can be removed, and `_modelFitDataBatchWise` can be the default. this is because any full preprocessing could be done offline (even outside deai), and then later using trivial preprocessing in batch-wise training mode.

this also makes it easier for the software architecture that we can base everything on batch-wise callback functions (no epochs etc needed)

also remove the switching logic such as from the task definition and documentation

maybe while doing it we can also update the preprocessing documentation / code a bit so it's easier for users to understand

Nacho114:
Issue: #208
In this PR we implement rounds which can be a fraction of an epoch (especially important for large datasets)

This is still work in progress, it seems to work for roundDuration <= 1 (original case + fractional case)

(Note I also address  suggestions on async by @martinjaggi such as handler->buffer)

@martinjaggi @s314cy 

I am worried about the naming: in the buffer we also have `rounds` now. However, in that context, a round is incremented when the buffer is filled up, while related to the local training round, it is not exactly the same, and I think that the naming being identical but slightly different might cause confusion. Perhaps we should rename the rounds in the buffer as cycles or something else (e.g. currentBufferCycle, bufferCycle, cycle, ...)? 

TODO:

- [x] Debug the roundDuration > 1 case (even though it might not be used, it is good to understand the error).
- [x] Address naming issue.
- [x] Implement it for decentralized.
- [x] Address #210 as this overlaps a bit.


martinjaggi:
thanks, looking good!

it should be fine to still call it `round`, as it is the same unit basically. here it's how many batches a client will need to train locally, until it can sent it to server or peers (to then become part of the corresponding round of the buffer, if selected and not delayed)

Nacho114:
yeah true, they are the same unit. But I think in the client, we should keep in mind that there are two "rounds", one, the current local training round, and then there is also the remote round which corresponds to the latest fetched model (i.e. in the remote server at what round was the buffer averaged). So we could distinguish them as round vs remoteRound. 

In the buffer and client we keep calling rounds, rounds, but in the client we have a separate name for the remote rounds.

martinjaggi:
in federated it's easiest to say the round number is dictated by the server, whenever a new averaging happens (buffer flush).

then when clients pull a model they also get the round metadata with the model, then do several local steps, and finally send back. when sending back it's crucial they attach the metadata of what time this update was originally based on (old model which was pulled). which the server (buffer) then uses to decide to drop it or not when received. 

let's use some terminology like `model update is based on round number T`, for these models going client->server

Nacho114:
fractional rounds are now working , still need to implement it for decentralized before merging the PR.

Nacho114:
@s314cy 
How are `postMetadata` and `getMetadataMap` used? I see them defined in fed client, but I can't see find where they are called?

s314cy:
> @s314cy 
> How are `postMetadata` and `getMetadataMap` used? I see them defined in fed client, but I can't see find where they are called? 

They are not being used yet. They were made for the future federated training informant to provide with stats shared by other nodes via the centralized server.

Nacho114:
> weights buffer

you mean rename asyncWeightsBuffer rneamed to WeightsBuffer?

martinjaggi:
> > weights buffer
> 
> you mean rename asyncWeightsBuffer renamed to WeightsBuffer?

no it was only affecting the comment, not the code

Nacho114:
> remove all epoch stuff, only use rounds?

We still use epochs for local training since there rounds dont matter. I could completely remove epochs and replace with round calls, but I would need to change a bit since I made this with only communication in mind.

martinjaggi:
ok will adapt local training later in a separate issue, not a high priority (it can also use rounds instead of epochs, config for users remains more clear then in both cases)

mobile-browser-based-version/src/logic/communication/federated/federated_client.ts
```diff
@@ -21,9 +21,16 @@ export class FederatedAsyncClient extends Client {
    */
   constructor (serverURL, task) {
     super(serverURL, task)
-    console.log('building Federated Async client')
     this.clientID = ''
-    this.round = -1 // The server starts at round 0, in the beginning we are behind
+    this.remoteModelRoundNumber = -1 // The server starts at round 0, in the beginning we are behind
```
martinjaggi: rename `remoteModelRoundNumber` -> `modelUpdateIsBasedOnRoundNumber`

mobile-browser-based-version/tests/logic/communication/client.test.ts
```diff
@@ -0,0 +1,83 @@
...
+    roundDurationLessThanOne.forEach((example) => {
+      expect(client._localRoundHasEnded(example.values[0], example.values[1], example.values[2], example.values[3], example.values[4])).equal(example.output)
+    })
+  })
+  it('Case: roundDuration > 1', () => {
+    roundDurationGreaterThanOne.forEach((example) => {
+      expect(client._localRoundHasEnded(example.values[0], example.values[1], example.values[2], example.values[3], example.values[4])).equal(example.output)
+    })
+  })
+})
```
martinjaggi: very nice work on the tests!

martinjaggi (approved):

Nacho114:
Note that to keep track of rounds we need to keep track of epochs. 

After reading your comments I realize I can improve the training manager a bit further, I will do these changes tomorrow morning and finalize this PR.

Nacho114:
After reading Martins comments and a discussion with @s314cy I've refactored the training classes.

1. I switched the names trainer and training_manager, since the primer was actually being the manager.
2. I made the trainer (previously training_manager) into an abstract class, and I made a separate LocalTrainer and DistributedTrainer classes that extend the trainer class that each perform local/distributed training. It's nice to separate these two into different classes since the trainer (previously training_manager) was quite packed. 
3. I deleted all unnecessary onEpoch callbacks, however for updating the UI graph it is necessary to call the onEpochEnd callback since onBatchEnd gives us no validation accuracy.
4. Keeping track of fractional rounds is not completely trivial, so I made a separate class for it, I explain in comments why it is necessary to keep track of the latest epoch in order to keep track of the current round. 
5. The result of the refactoring is that we now have small independent modules that do specific tasks and will hence be easily testable once @s314cy finishes re-working the loadData pipeline.

mobile-browser-based-version/src/logic/training/round_tracker.ts
```diff
@@ -0,0 +1,68 @@
...
+      this.numberOfBatchesInAnEpoch = RoundTracker.numberOfBatchesInAnEpoch(trainSize, batchSize)
+
+      console.log(`RoundTracker: roundDuration: ${roundDuration}, nb batches in an epoch ${this.numberOfBatchesInAnEpoch}`)
+    }
+
+    /**
+   * Returns true if a local round has ended, false otherwise.
+   *
+   * @remark
+   * Batch is the current batch, this goes from 1, ... , batchSize*trainSize.
```
martinjaggi: typo here, batch size doesn't matter, but you meant trainsize/batchsize. anyway would be nicer to just let batch count to infinity, not reset to zero every epoch

mobile-browser-based-version/server/tests/logic/federated/async_weights_buffer.test.ts
```diff
@@ -0,0 +1,69 @@
...
+let mockUpdatedWeights = []
+
+const mockAggregateAndStoreWeights = async (_weights: any) => {
+  mockUpdatedWeights = _weights
+}
+
+describe('AsyncWeightHandler tests', () => {
+  it('add weight with old time stamp returns false', async () => {
+    const t0 = -1 //
+    const asyncWeightBuffer = new AsyncWeightsBuffer(taskId, bufferCapacity, mockAggregateAndStoreWeights)
```
s314cy, line 19: `asyncWeightsBuffer`

mobile-browser-based-version/server/tests/logic/federated/async_weights_buffer.test.ts
```diff
@@ -0,0 +1,69 @@
...
+const bufferCapacity = 3
+const weights = [0, 1, 2]
+const newWeights = [3, 4, 5]
+let mockUpdatedWeights = []
+
+const mockAggregateAndStoreWeights = async (_weights: any) => {
+  mockUpdatedWeights = _weights
+}
+
+describe('AsyncWeightHandler tests', () => {
```
s314cy: `'AsyncWeightsHandler tests'`

mobile-browser-based-version/server/tests/router/federated.test.ts
```diff
@@ -30,15 +30,15 @@ function disconnectHeader (platformID: string, taskID: string, clientID: string)
   return `/${platformID}/disconnect/${taskID}/${clientID}`
 }
 
-function postAsyncWeightHeader (platformID: string, taskID: string, clientID: string) {
-  return `/${platformID}/async/weights/${taskID}/${clientID}`
+function postWeightHeader (platformID: string, taskID: string, clientID: string) {
```
s314cy, line 33: `postWeightsHeader`

mobile-browser-based-version/server/src/router/router.ts
```diff
@@ -40,18 +40,9 @@ federatedRouter.get('/', (req, res) => res.send('FeAI server'))
 federatedRouter.get('/connect/:task/:id', handlers.connect)
 federatedRouter.get('/disconnect/:task/:id', handlers.disconnect)
 
-federatedRouter.get('/selection/:task/:id', handlers.selectionStatus)
+federatedRouter.post('/weights/:task/:id', handlers.postWeights)
```
s314cy, line 43: :+1:

mobile-browser-based-version/src/components/sidebar/Settings.vue
```diff
@@ -191,7 +191,7 @@ export default defineComponent({
       'setPlatform'
     ]),
     toggleIndexedDB () {
-      this.setIndexedDB(!this.$store.state.useIndexedDB && window.indexedDB)
+      this.setIndexedDB(!!(!this.$store.state.useIndexedDB && window.indexedDB))
```
s314cy: the commit function in `store.js` already converts its argument to boolean, to avoid writing this trick in every call

mobile-browser-based-version/src/logic/communication/client.ts
```diff
@@ -1,57 +1,53 @@
 import * as tf from '@tensorflow/tfjs'
+import { Task } from '../task_definition/base/task'
+import { TrainingInformant } from '../training/training_informant'
 
-export class Client {
-  serverURL: any;
-  task: any;
-  constructor (serverURL, task) {
+export abstract class Client {
```
s314cy, line 4: :+1:

mobile-browser-based-version/src/logic/communication/client.ts
```diff
@@ -1,57 +1,53 @@
...
+  async updateModelAndRoundFromServer () {}
+
   /**
    * Handles the connection process from the client to any sort of
    * centralized server.
    */
-  async connect (epochs?): Promise<any> {
-    throw new Error('Abstract method')
-  }
+  abstract connect (epochs?:number): Promise<any>
```
s314cy, line 16: :+1:

mobile-browser-based-version/src/logic/communication/client.ts
```diff
@@ -1,57 +1,53 @@
...
 
   /**
    * The training manager matches this function with the training loop's
    * onTrainEnd callback when training a TFJS model object. See the
    * training manager for more details.
    */
-  async onTrainEndCommunication (model, trainingInformant) {
-    trainingInformant.addMessage('Training finished.')
-  }
+  abstract onTrainEndCommunication (model, trainingInformant: TrainingInformant): Promise<void>
```
s314cy, line 29: I think the callback methods should not be abstract but rather be some sort of `default` methods, since the `Client` abstract class may include some routine common to every subclasses, such as displaying the RAM usage

mobile-browser-based-version/src/logic/communication/client.ts
```diff
@@ -1,57 +1,53 @@
...
+  constructor (serverURL: string, task: Task) {
     this.serverURL = serverURL
     this.task = task
   }
 
+  /**
+   * Init holder for federated, we need to fetch latest server round
+   * and model before training.
+   */
+  async updateModelAndRoundFromServer () {}
```
s314cy: I'm not sure why this is in `Client`, as this is typically not used by `DecentralizedClient`

mobile-browser-based-version/src/logic/communication/client_builder.ts
```diff
@@ -1,25 +1,18 @@
...
+import { FederatedClient } from './federated/federated_client'
+import { Client } from './client'
+import { Task } from '../task_definition/base/task'
+import { Platform } from '../../platforms/platform'
 
-export function getClient (platform, task, password: string, federatedWithSyncScheme: boolean = false) {
+export function getClient (platform: Platform, task: Task, password: string): Client {
   switch (platform) {
-    case 'deai':
+    case Platform.decentralized:
```
s314cy, line 9: :1st_place_medal:

mobile-browser-based-version/src/logic/communication/decentralised/decentralised_client.ts
```diff
@@ -158,21 +160,17 @@ export class DecentralisedClient extends Client {
   /**
    * Disconnection process when user quits the task.
    */
-  disconnect () {
+  async disconnect () {
```
s314cy, line 163: we can discuss the `async`/`await` policy later on @Nacho114

mobile-browser-based-version/src/logic/task_definition/base/task.ts
```diff
@@ -1,6 +1,56 @@
 import * as tf from '@tensorflow/tfjs'
 import * as memory from '../../memory/model_io'
 
+class DataExample {
```
s314cy, line 4: :+1:

mobile-browser-based-version/src/logic/training/distributed_trainer.ts
```diff
@@ -0,0 +1,96 @@
+import * as memory from '../memory/model_io'
+import { TrainingInformation } from '../task_definition/base/task'
+import { Trainer } from './trainer'
+
+// TODO: user trainingInformant and task as class
+
+/**
+ * Class whose role is to train a model in a distributed way with a given dataset.
+ */
+export class DistributedTrainer extends Trainer {
```
s314cy: great! this simplifies my current work for #191

mobile-browser-based-version/src/logic/training/round_tracker.ts
```diff
@@ -0,0 +1,65 @@
...
+ *
+ * @remark
+ * In distributed training, the client trains locally for a certain amount of epochs before sharing his weights to the server/neighbor, this
+ * is what we call a round. A round is measured in epochs (1 round == 1 epoch), a round can be any number > 0, if roundDuration = 0.5, then
+ * every half epoch we share our weights with the server.
+ *
+ * The role of the RoundTracker is to keep track of when a roundHasEnded using the current batch number. The batch in the RoundTracker is cumulative whereas
+ * in the onBatchEnd it is not (it resets to 0 after each epoch).
+ */
+export class RoundTracker {
```
s314cy: :+1:

mobile-browser-based-version/src/logic/training/trainer.ts
```diff
@@ -1,147 +1,224 @@
...
+ * locally or in a distributed way.
+ */
+export abstract class Trainer {
+  task: Task;
+  client: Client;
+  trainingInformant: TrainingInformant;
+  useIndexedDB: boolean;
+  stopTrainingRequested: boolean = false;
+  model: tf.LayersModel;
+  data: any; // TODO: use class type @s314cy once you have created the dataset class
```
s314cy: :+1:

s314cy (changes requested):

mobile-browser-based-version/src/logic/communication/client.ts
Nacho114: The RAM usage should be displayed inside the onRoundEnd call backs in the trainer? If there are no common subroutines, is it not better to keep it abstract?

mobile-browser-based-version/src/logic/communication/decentralised/decentralised_client.ts
Nacho114: yes, this is important, it's not clear to me what it is right now

mobile-browser-based-version/src/logic/communication/client.ts
Nacho114: you are right, it's also no longer used due to refactoring so I will delete it.

PR #229, opened 2022-03-09
Update issue templates

Nacho114:
Added bug and feature issue templates, I basically used the templates and removed some redundant info.


tharvik (changes requested):
Great idea to add issue templates, that really helps collaboration :smile:

A few remarks
* why is there two bug reports template? `---bug-report-.md` looks like an old version
* can you rename the `---feature-request.md` to avoid this weird `---` at the begining? (and I let you choose between dash-case and underscore-case :stuck_out_tongue_closed_eyes: )

PR #230, opened 2022-03-09
Add lint, build and test jobs to CI

### LINKED ISSUE #225: Enhance Github actions for CI/CD automation
Enhance Github actions for CI/CD automation:

On merging from `develop` -> `production`

- Run server tests
- Build server docker
- Run server docker
- Run client tests
- Build and update to github pages
- Update google app engine (server)

On PR creation

Allow only a PR to be pushed for review if it passes auto tests (PR-draft would still work without tests)

tharvik:
Mine first PR to disco, tell me if I missed anything :grin: 

Takes care of #225

It triggers upon merging from `develop` -> `production` and on PR creation.
 
> * Run server tests
> * Build server docker
> * Run client tests

Done and
* add some caching while at it
* change the client's `npm run test` to also start the server

> * Run server docker

Hum, I'm using the one build from sources for the test, is it better to start the docker one?

> * Build and update to github pages

The deploy workflow already takes care of that. I changed the trigger to do it iff the `lint-build-test` passes.

> * Update google app engine (server)

I don't know what are the credentials for GAE.
One can use [Google's GitHub Action](https://github.com/google-github-actions/deploy-appengine)  for that, feel free to push to this PR if you have theses.

> Allow only a PR to be pushed for review if it passes auto tests (PR-draft would still work without tests)

Enforcing this need an admin change to the setting of the repo. It would be best to enforce it after merging this PR.

### DIFF
--- a/.github/workflows/deploy-server.yml
+++ b/.github/workflows/deploy-server.yml
@@ -6,16 +6,12 @@ on:
     types: [completed]
 
 env:
-  server_dir: mobile-browser-based-version/server
   server_url: https://deai-313515.ew.r.appspot.com/
 
 jobs:
   deploy:
     name: deploy-gae
     runs-on: ubuntu-latest
-    defaults:
-      run:
-        working-directory: ${{ env.server_dir }}
 
     steps:
       - name: checkout
@@ -24,6 +20,7 @@ jobs:
       - name: deploy
         uses: google-github-actions/deploy-appengine@v0.2.0
         with:
+          working_directory: mobile-browser-based-version/server
           deliverables: app.yaml
           version: prod
           project_id: ${{ secrets.GCP_PROJECT }}
--- a/.github/workflows/deploy-server.yml
+++ b/.github/workflows/deploy-server.yml
@@ -10,21 +10,25 @@ env:
 
 jobs:
   deploy:
-    name: deploy-gae
+    name: deploy
     runs-on: ubuntu-latest
 
     steps:
       - name: checkout
         uses: actions/checkout@v2
 
-      - name: deploy
-        uses: google-github-actions/deploy-appengine@v0.2.0
+      - name: authenticate-to-google-cloud
+        uses: "google-github-actions/auth@v0"
+        with:
+          credentials_json: "${{ secrets.GCP_CREDENTIALS }}"
+
+      - name: deploy-to-google-cloud
+        uses: google-github-actions/deploy-appengine@v0.8.0
         with:
           working_directory: mobile-browser-based-version/server
           deliverables: app.yaml
           version: prod
           project_id: ${{ secrets.GCP_PROJECT }}
-          credentials: ${{ secrets.GCP_CREDENTIALS }}
 
       - name: test
         run: curl "${{ env.server_url }}"
--- a/.github/workflows/deploy-server.yml
+++ b/.github/workflows/deploy-server.yml
@@ -1,0 +1,33 @@
+name: Deploy to GAE
+on:
+  workflow_run:
+    workflows: ["Lint, test and build"]
+    branches: [production]
+    types: [completed]
+
+env:
+  server_dir: mobile-browser-based-version/server
+  server_url: https://deai-313515.ew.r.appspot.com/
+
+jobs:
+  deploy:
+    name: Deploying to Google Cloud
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: ${{ env.server_dir }}
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Deploy to App Engine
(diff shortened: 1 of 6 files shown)


Nacho114:
We feel honored to receive your first PR! :)

`Hum, I'm using the one build from sources for the test, is it better to start the docker one?`

Good question, maybe it is best to run the docker image we build? We build it during the process as you point out, so we can verify it builds correctly, might as well go the extra step and check it works.

mobile-browser-based-version/package.json
```diff
@@ -7,7 +7,7 @@
     "prod": "vue-cli-service serve --mode production",
     "build": "vue-cli-service build",
     "lint": "vue-cli-service lint",
-    "test": "env TS_NODE_COMPILER_OPTIONS='{\"module\": \"commonjs\" }' mocha --exit -r ts-node/register --file tests/setup.ts 'tests/**/*.ts'",
+    "test": "( cd server && npm start ) & env TS_NODE_COMPILER_OPTIONS='{\"module\": \"commonjs\" }' mocha --exit -r ts-node/register --file tests/setup.ts 'tests/**/*.ts' && kill $! && wait",
```
Nacho114: the server takes a bit of time to build (2-3s), maybe in the future it might be longer, could that be a problem? e.g. the tests start before the server finishes building.

Nacho114:
`I don't know what are the credentials for GAE.`

I have them, I can take care of this.

Nacho114:
Ah I just saw, there is a GitHub section now with:

`All checks have passed`

Very cool stuff <3

mobile-browser-based-version/package.json
tharvik: Right, I'm now using a script to ensure that it is started before running the tests: `with_server`

tharvik:
>> I'm using the one build from sources for the test, is it better to start the docker one?
>
> Good question, maybe it is best to run the docker image we build? We build it during the process as you point out, so we can verify it builds correctly, might as well go the extra step and check it works.

I added a script, `with_server`, that ensure that a server is answering before running the given command, then takes care of stopping it. It is used by `npm run test`.

>> I don't know what are the credentials for GAE.
>
> I have them, I can take care of this.

Alright, fell free to push to this PR :)

Nacho114:
I added the gcloud deploy automation script, added the corresponding gcloud credentials to github secrets + renamed the names in the yml files to follow dash-case. Let me know if it looks good / naming is better like this.

.github/workflows/deploy-server.yml
```diff
@@ -0,0 +1,33 @@
...
+  server_dir: mobile-browser-based-version/server
+  server_url: https://deai-313515.ew.r.appspot.com/
+
+jobs:
+  deploy:
+    name: Deploying to Google Cloud
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: ${{ env.server_dir }}
```
tharvik: This is only set for `run` steps, if you want to use it for `google-github-actions/deploy-appengine`, you can use its [`working_directory` input](https://github.com/marketplace/actions/deploy-to-app-engine#inputs)

mobile-browser-based-version/package.json
```diff
@@ -7,7 +7,8 @@
     "prod": "vue-cli-service serve --mode production",
     "build": "vue-cli-service build",
     "lint": "vue-cli-service lint",
-    "test": "env TS_NODE_COMPILER_OPTIONS='{\"module\": \"commonjs\" }' mocha --exit -r ts-node/register --file tests/setup.ts 'tests/**/*.ts'",
+    "blam": "false; true",
```
Nacho114: what is blam for?

.github/workflows/deploy-server.yml
```diff
@@ -0,0 +1,33 @@
...
+
+    steps:
+      - name: checkout
+        uses: actions/checkout@v2
+
+      - name: deploy
+        uses: google-github-actions/deploy-appengine@v0.2.0
```
tharvik: Is there a reason behind using an old version?

```suggestion
        uses: google-github-actions/deploy-appengine@v0.8.0
```

btw, to ensure that it works, I usually force it to run, by branching of and changing the trigger to `on: push`

mobile-browser-based-version/package.json
tharvik: hoo, good catch, it was a debug line; removing it

tharvik:
> I added the gcloud deploy automation script, added the corresponding gcloud credentials to github secrets + renamed the names in the yml files to follow dash-case. Let me know if it looks good / naming is better like this.

Indeed, that more consistent.

.github/workflows/deploy-server.yml
Nacho114: thanks for upgrading it, the reason for old version: copy pasted code. How to test: thanks that's a good tip, will test it out.

PR #231, opened 2022-03-14
222 training fix damien

dgengler6:
### What this PR does. 
- Federated training was only fetching the global model, but never updating its local model with it. 
- On top of that some asynchronous callbacks were never preceded by `await` and thus the promises never resolved
- All of this is now done and federated training should be actually working.

### DIFF
--- a/mobile-browser-based-version/package.json
+++ b/mobile-browser-based-version/package.json
@@ -24,6 +24,7 @@
     "papaparse": "^5.3.1",
     "peerjs": "^1.3.2",
     "seedrandom": "^3.0.5",
+    "stable": "^0.1.8",
     "tippy.js": "^6.3.1",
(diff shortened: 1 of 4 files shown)


mobile-browser-based-version/package.json
```diff
@@ -24,6 +24,7 @@
     "papaparse": "^5.3.1",
     "peerjs": "^1.3.2",
     "seedrandom": "^3.0.5",
+    "stable": "^0.1.8",
```
Nacho114, line 27: What is this for?

s314cy (approved):
LGTM

mobile-browser-based-version/package.json
s314cy: Ah sorry I missed this package

PR #233, opened 2022-03-14
add local finetuning

### LINKED ISSUE #221: Add Local Fine-tuning personalization
- Once the collaborative training of the global model is done, we perform a final training locally in order to personalize the global model to the local dataset.

dgengler6:
### What this PR will do 
- Fixes #221 
- Add Local Fine-Tuning, a simple method of model personalization. 
- The idea is to train for some more rounds once the global model has converged. That way we personalize the global model to the local data. 

### Questions 

- I find very little about this online but most of what I read concerns FL, but I guess this could be applied both to Federated and Decentralized Learning, is that right ?


martinjaggi:
yes this makes sense in both federated and decentralized.

maybe it can be achieved by just setting the epochs per round (task config) to a very large value and then stop before aggregating again

dgengler6:
I was thinking about adding a parameter `finetuningEpochs` in task config and run a new training for this duration. I think it would maybe add a bit of complexity but it would be easier to understand when looking at the code. This approach allows us ( if we want ) to finetune for more epochs than `epochsPerRound`

Nacho114:
> I was thinking about adding a parameter `finetuningEpochs` in task config and run a new training for this duration. I think it would maybe add a bit of complexity but it would be easier to understand when looking at the code. This approach allows us ( if we want ) to finetune for more epochs than `epochsPerRound`

The task config is quite dense already, so I am hesitant to add complexity. Maybe at first, just set a constant inside the fine tuning code, so it's the same for all for the time being. We can brain storm once you have written and tested the code for fine tuning, maybe then it will be more clear what to do with `finetuningEpochs`.

PR #235, opened 2022-03-15
github/lint-test-build: fix push trigger

tharvik:
When pushing to a branch, the CI triggers, but that's not supposed to do so. It should only do it when pushing to the `production` branch or opening a PR.

It seems that the push trigger spec was wrong (but not flagged as so by GitHub!)


Nacho114 (approved):
Ahh these kind of typos that fail silently are the hardest to spot!

PR #236, opened 2022-03-15
Run tests without docker

tharvik:
* `npm run test` needs docker to run, which is not that user friendly. Avoid using it by default.
* add some explaination on how to run a server for the tests


Nacho114 (approved):
perfect, makes it easier to run for us shell noobs :p

Issue #240, opened 2022-03-18
add end2end testing


martinjaggi:
document for us internally the key metrics for some of the standard tasks (e.g. titanic), and test those by hand before/after changes which could affect it. can be done by human, does not need to be automated.
this is important in addition to the unit tests. 

1. test accuracy for training alone 
2. test accuracy for 2 clients, with the same dataset strictly partitioned between the two

no code needed for now but clear values needed and file splits and documentation of expected results in a readme etc (or on the task)

ideally we'd also have these 2 models tested on a completely separate holdout set, removed from the training set before given to disco.

will make a new issue to support validation on separate set

martinjaggi:
done in #380 and some subsequent PRs like #404

PR #244, opened 2022-03-23
Simplify the data pipeline & various changes

### LINKED ISSUE #191: Re-working the adding data / data check code.

s314cy:
### **Core changes**
Key idea: Rework the data loading process in order to be able to run simulations more easily to verify federated learning. Further, by making it modular we can also test disco end2end and make it easier to extend to new datasets. Closes #191.

Here is the list of the file naming changes:
`logic/` -> `core/`
`logic/data_validation/` -> `core/dataset/`
`logic/dataset/` -> `core/dataset/`
`logic/task_definition/` -> `core/task/`
`logic/task_definition/base/task.ts` -> `core/task/task.ts`
deleted `logic/task_definition/*` except `task.ts`
deleted `src/logic/data_validation/{check_data.ts, file_upload_manager.ts, helpers_image_tasks.ts}`
added `src/core/dataset/{dataset.ts, source_type.ts}`
added `src/core/dataset/data_loader/{data_loader.ts, text_loader.ts, image_loader.ts}`

There are many ongoing changes in this PR, thus I will try to be as exhaustive as possible. A few code tweaks will be omitted.

The `logic/` folder was renamed to `core/` for clarity. It also is the standard folder name used in open source projects for this kind of files.

The data loaders defined in `core/dataset/data_loader/` include a single `load()` function, taking either URLs or file system paths as input, and returning a `tf.data.Dataset` object. This is used by the dataset builder to read the content from its bank of files. 
The goal here is to be environment invariant, i.e. to provide code working on both the browser and Node.js. Note that the dataset builder class aims to replace the file upload manager.

Mentions of CSV/csv were replaced by Text/text, since CSV is merely a subset of textual data with a predefined information delimiter (namely, the comma). This very delimiter is now an argument in `core/dataset/data_loader/text_loader.ts`.

A lot of routines (preprocessing, model fetching, label prediction, data validation) were located in the `task_definition/` folder, under the form of the file hierarchy `task_definition/task/{base/*, csv/*, image/*}`. Each data type had corresponding `task` and `task_helper` objects, including the aforementioned routines. `{csv, image}_task_helper.ts` called routines defined in `{csv, image}_task.ts`. It is counter-intuitive to have these routines called in such a way in the `task_definition/` folder, which was meant to contain code related to the very definition of a task and utilities to create one (i.e. task creation form). Furthermore, the code itself is quite old and hard to read/understand at times. It is also lengthy since it does not reuse existing TF.js utility functions (e.g. `tf.data.csv()`, one-hot encoding).

Preprocessing (on loading the dataset, opposed to batch preprocessing that happens during training) defined in the `task` and `task_helper` files of `logic/task_definition` was removed, @Nacho114 and I are planning to create its own dedicated `src/core/preprocessing` folder for clarity for future devs. This is part of the endeavor to simplify the task creation process brought to the users, such as the selection of predefined data preprocessing. The preprocessing defined in `core/preprocessing/` would thus be called by both the dataset builder (`core/dataset`) and the task creation routine (`core/task/`).

The extensive data checks performed in `logic/task_definition` and defined in `logic/data_validation/check_data.ts` were removed. The goal here is to move the data checks to the newly created `core/dataset/` folder and keep it as simple as possible. It would then be called within the same folder, namely by `core/dataset/dataset_builder.ts`.

The `logic/data_validation/` folder contained mostly unrelated code and was poorly organized, thus it is removed and its core content will be found in `core/dataset`.

### **Component changes**

Here is the list of file naming changes:
`building_frames/upload/` -> `building_frames/dataset_input/`
`building_frames/upload/UploadingFrame.vue` -> `building_frames/dataset_input/DatasetInputFrame.vue`
`building_frames/upload/SingleUploadFrame.vue` -> `building_frames/dataset_input/FileSelectionFrame.vue`
`building_frames/csv/` -> `building_frames/text`
`building_frames/csv/CsvTrainingFrame.vue` -> `building_frames/text/TextTrainingFrame.vue`

Since there is no data uploading process (everything remains local), the “upload” term is no longer used. `FileSelectionFrame.vue` is a much simpler version of `SingleUploadFrame.vue`. Unnecessary code was removed, whereas necessary code was simplified, yet yielding the same functionality. Now, the file selection frame simply receives file input and communicates it to its parent, namely `DatasetInputFrame.vue`.

`DatasetInputFrame.vue` simply includes file selection frames and receives file input from them. It then populates the dataset builder. The dataset builder is first defined in `building_frames/{text, image}/{Text, Image}TrainingFrame.vue` and given the corresponding `{text, image}_loader.ts` data loader. It is then passed down to `building_frames/container/TrainingFrame.vue`, which contains the aforementioned `DatasetInputFrame.vue`.

All-in-all, this removed the many interactions happening between the training manager and the file upload manager. Indeed, the file upload manager was passed to many different objects (both in `core/` and `components/`), basically making it a global variable. Instead, the training manager is now simply fed a dataset whenever the user starts training, and has no interaction with the newly created dataset builder class, which is simply used by frames and frozen once the dataset is finally created. It is then much easier to write module testing and extend to new data types.


Nacho114:
Nice, with this it will be very easy to finally do #240, end to end testing.

martinjaggi:
26k lines deleted, that's quite amazing!

maybe as it's so big could you break it up into two PRs, one for the main files restructuring/renaming, and then later a separate PR for the functional changes on the preprocessing itself? this way it would be easier to see what has changed later

or keep as now as it's probably hard to disentangle.
well done!

Nacho114:
or keep as now as it's probably hard to disentangle. >

I think this, there is so much he changed anyway in this PR so breaking it down would also be a lot of work. But for sure, in the future we should aim to have smaller and more concise PR's as they are easier to read + go back on to understand how things evolved.

s314cy:
Actually, I think I would be able to properly and easily separate the two kind of changes (renames and dir moves vs. code changes), and I might want to do this before going further to keep it a simple task.

I'll open a separate non-breaking PR for file changes, which can then be merged today :)

mobile-browser-based-version/src/core/training/training_manager.ts
```diff
@@ -1,3 +1,35 @@
+<<<<<<< HEAD
```
martinjaggi: there is a merge conflict here currently it seems

mobile-browser-based-version/src/components/main_frames/MainTestingFrame.vue
```diff
@@ -7,16 +7,16 @@
...
+    v-else-if="task.trainingInformation.dataType == 'text'"
     :id="id"
     :task="task"
   />
 </template>
 
 <script>
 import ImageTestingFrame from '../building_frames/image/ImageTestingFrame.vue'
-import CsvTestingFrame from '../building_frames/csv/CsvTestingFrame.vue'
+import TextTestingFrame from '../building_frames/text/TextTestingFrame.vue'
```
martinjaggi: maybe better to keep this one named `csv` or tabular. ok it's for general numerical features, but now it's easily confused with text/NLP input which we might add soon, but this one is not

martinjaggi:
the `package-lock` thing let's do it in a separate pr? (did you want to add it to gitignore?)

s314cy:
> the `package-lock` thing let's do it in a separate pr? (did you want to add it to gitignore?)

you're right, I didn't mean to commit changes to `package*.json`, I'll revert the files to their current value on the develop branch

mobile-browser-based-version/src/core/dataset/data_loader/tabular_loader.ts
```diff
@@ -0,0 +1,40 @@
...
+    this.delimiter = delimiter
+  }
+
+  /**
+   * Expects delimiter-separated data made of N columns.
+   * Every source should contain N-1 feature columns and 1 single label column.
+   * @param sources List of URLs or file system paths.
+   * @param features List of feature column names.
+   * @param label Label column name.
+   * @returns A TF.js dataset built upon read textual data stored in the given sources.
```
martinjaggi: tabular or csv (also in function name to be consistent?)

mobile-browser-based-version/src/core/dataset/data_loader/image_loader.ts
```diff
@@ -0,0 +1,69 @@
...
+        }
+        return iterator
+      })
+    } else {
+      return tf.data.generator(() => {
+        let index = 0
+        const iterator = {
+          next: () => {
+            return {
+              value: { xs: '', ys: '' },
```
Nacho114: The if and else return object look very similar, maybe we can use the same but only change the `value` of the return?

in pseudo code:

f(images, index) = ImageLoader.readImageFrom(images[index++])

g(images, index) = { xs: '', ys: '' }

valueFun = if (config.labels === undefined) f else g

and then have valueFun(images, index) called to set the { value:  valueFun(images, index) } 

maybe if we define f and g using anonymous fun we get less code? (just an idea)

mobile-browser-based-version/src/core/dataset/data_loader/data_loader.ts
```diff
@@ -0,0 +1,10 @@
+import { Dataset } from '../dataset_builder'
+
+export type Source = string | File
```
Nacho114, line 3: Might be a good idea to create another type for string, string at the end will hold a url right?

so maybe we could do

export type StringURL = string
export type Source = StringURL | File

and then do the changes accordingly, e.g in `image_loader.ts`
`if (typeof image === 'string')` -> `if (typeof image === StringURL')`
I also think the latter is more clear when reading the code.

Nacho114:
It's looking really good!

mobile-browser-based-version/src/core/dataset/data_loader/data_loader.ts
s314cy: In the case of the tabular loader, the input may either be a File object or a URL (i.e. a regular web URL or one of the form `'file://file/system/path'`). However, for the image loader, the input may either a File object or a file system path, but not a URL (no `'file://'` prefix).

This is why it can become quite cumbersome to differentiate the string type. However, I planned to modify the tabular loader so that it only accepts file system paths for now, to be consistent with the image loader. I would simply prepend the required prefix within the function.

What do you think?

mobile-browser-based-version/src/core/dataset/data_loader/image_loader.ts
s314cy: Good idea, let's hope the scope resolves correctly 👍

mobile-browser-based-version/src/core/dataset/data_loader/data_loader.ts
Nacho114: @tharvik , at the end with Hugo we agreed that a `type Path = string` was a good idea, however it seems we cannot use it for type checking, i.e. `source instanceof Path` nor `source typeof === Path` work, where `source` if of type `Source`, which is defined as `type Source = string | File`.

It [seems](https://stackoverflow.com/questions/62769751/can-i-determine-if-a-value-matches-a-type-alias) that types are erased at runtime, so there is no way to check it. However this seems too bad since it would make the code easier to read. Do you know if there is another way to check types here?

martinjaggi (approved):

PR #251, opened 2022-03-28
Server : add async weight informant to display contribution metrics

### LINKED ISSUE #223: Store (server side) and display (client side) AsyncWeightBuffer data
In order to display contribution metrics, the first important object will give understanding what is happening in the server. 

The user would be interested in (among other things):

- How many people have contributed to the current model (overall, in the last round)
- How relevant are the other users data to me.
- ...

The first step for this is to build a class for the `AsyncWeightBuffer`, say `AsyncWeightInformant` that owns an object of the latter class and keeps track of what is going on. This information can then be queried by the client in order to display said information on the web page, e.g. number of current participants.

High level the following tasks need to be done for this:

- [x] Brainstorm `AsyncWeightInformant` class and define high level functions / attributes.
- [x] Build `AsyncWeightInformant` class.
- [x] Create test for `AsyncWeightInformant` class (in `server/tests/`)
- [ ] Create a route to query the information.
- [ ] Create test for the new route. (in `server/tests/`)
- [ ] Add a fetch `AsyncWeightInformant` to the federated `api`
- [ ] Create test `api` (in `tests/`)
- [ ] Add `api` call in the onRoundEndCommunication and update the peer information (the same one used for updating peers in decentralised).

dgengler6:
Fixes #223 
**What does this add ?** 
- Add the asyncWeightInformant class that watches asyncWeightBuffer
- Keeps statistics about the Buffer
- The clients can request the statistics and display it in the training frame
- Add tests for both server and client side 

**TODO steps**
- [x] asyncWeightInformant class that keeps statistics about asyncWeightBuffer 
- [x] test file serverside for this class
- [x] add route to query statistics 
- [x] add display in client 
- [x] test file clientside


mobile-browser-based-version/server/src/logic/federated/async_weights_buffer.ts
```diff
@@ -22,6 +23,7 @@ export class AsyncWeightsBuffer {
     buffer: Map<string, any>;
     round: number;
     roundCutoff: number;
+    observers: AsyncWeightsInformant[]
```
Nacho114: Why would there be a list of weight informants rather than just one?

mobile-browser-based-version/server/src/logic/federated/async_weights_informant.ts
```diff
@@ -0,0 +1,92 @@
+import { AsyncWeightsBuffer } from './async_weights_buffer'
+
+/**
+ * @remarks
+ * taskID: corresponds to the task that weights correspond to.
+ */
+export class AsyncWeightsInformant {
+    taskID: string;
```
Nacho114: Since a weight informant is observing an `asyncWeightsBuffer`, he can retrieve the taskID from there? (maybe redundant to also store it here?)

mobile-browser-based-version/server/src/logic/federated/async_weights_informant.ts
```diff
@@ -0,0 +1,92 @@
...
+    asyncWeightsBuffer: AsyncWeightsBuffer
+
+    // This could rapidly grow so we should have some way to do garbage collection ?
```
Nacho114, line 6: Yes! good point, to be discussed.

mobile-browser-based-version/server/src/logic/federated/async_weights_informant.ts
```diff
@@ -0,0 +1,92 @@
...
+
+      this._updateRound()
+      this._updateNumberOfParticipants()
+
+      // DEBUG
+      console.log('After update')
+      this.printAllInfos()
+    }
+
+    _updateRound () {
```
Nacho114: In TS you can make private functions by using the `private` tag, just like in java, so no more need to add a _ prefix.

mobile-browser-based-version/server/tests/logic/federated/async_weights_informant.test.ts
```diff
@@ -0,0 +1,53 @@
+
```
Nacho114, line 1: Tests <3 :)

mobile-browser-based-version/server/src/logic/federated/async_weights_buffer.ts
dgengler6: Good question, I'll remove that

mobile-browser-based-version/server/src/logic/federated/async_weights_informant.ts
dgengler6: Sure thing !

dgengler6: Oh Btw I noticed that when creating the `asyncWeightsBuffer` we were passing the whole task object and not the taskId, I'll double check that

mobile-browser-based-version/src/core/communication/federated/federated_client.ts
```diff
@@ -125,9 +125,15 @@ export class FederatedClient extends Client {
...
   }
 
+  private async fetchServerStatisticsAndUpdateInformant (trainingInformant: TrainingInformant) {
+    const serverStatistics = await api.getAsyncWeightInformantStatistics(this.task.taskID, this.clientID)
+    trainingInformant.updateWithServerStatistics(serverStatistics.data.statistics)
+  }
+
   async onRoundEndCommunication (model: tf.LayersModel, round: number, trainingInformant: TrainingInformant) {
     await this._postWeightsToServer(model.weights)
     await this._fetchChangesAndUpdateModel(model)
```
Nacho114: while you are at it, could you remove the _ suffix and make the functions private :)

mobile-browser-based-version/src/core/communication/federated/federated_api.ts
```diff
@@ -35,11 +35,22 @@ export async function postWeights (taskID: string, clientID: string, weights, ro
   })
 }
 
-export async function getRound (taskID: string, clientID: string) {
-  const url = serverUrl().concat(`round/${taskID}/${clientID}`)
+// I was thinking that since simple gets like this might be used quite commonly it could be nice to
+// have a function that crafts the query directly, what do you think ?
+async function fetchFromServer (route: string, ...requiredParameters: string[]) {
```
Nacho114, line 39: nice, maybe `getFromServer` is a better name? Since it's based on a `GET` request

mobile-browser-based-version/server/src/logic/federated/request_handlers.ts
```diff
@@ -340,6 +347,34 @@ export async function getRound (request, response) {
...
+/**
+ * Get the JSON containing statistics about the async weight buffer
+ *
+ * @param request
+ * @param response
+ * @returns
+ */
+export async function getAsyncWeightInformantStatistics (request, response) {
+  // Check for errors
+  const type = REQUEST_TYPES.GET_ASYNC_ROUND
```
Nacho114, line 359: You should create a new `REQUEST_TYPES` for the stats (there is one for each request).

mobile-browser-based-version/server/src/logic/federated/async_weights_informant.ts
```diff
@@ -0,0 +1,95 @@
...
+
+    // Future statistics
+    getDistanceFromGlobalModel (id) {}
+
+    getShapleyValueForUser (id) {}
+
+    // Debug
+
+    printAllInfos () {
+      console.log('task : ', this.asyncWeightsBuffer.taskID)
```
Nacho114, line 89: There is a shorter way to edit strings for the log using `
the above is equivalent to:
```js
console.log(`task : ${this.asyncWeightsBuffer.taskID}`)
```

Also for debugging, it's even easier to do the following:

console.log({ a, b, c})
In the output you will get both the variable name and inputs.

much shorter than:

console.log('a', a)
...

mobile-browser-based-version/server/src/logic/federated/async_weights_buffer.ts
```diff
@@ -31,6 +33,17 @@ export class AsyncWeightsBuffer {
...
       this.round = 0
+      this.observer = null
+    }
+
+    registerObserver (observer: AsyncWeightsInformant) {
+      this.observer = observer
+    }
+
+    _notifyObserver () {
+      if (this.observer != null) {
```
Nacho114, line 44: @tharvik we are no longer using `null` right? Should we replace `null` with `undefined`?

mobile-browser-based-version/server/src/logic/federated/async_weights_informant.ts
```diff
@@ -0,0 +1,95 @@
+import { AsyncWeightsBuffer } from './async_weights_buffer'
+
+export class AsyncWeightsInformant {
+    asyncWeightsBuffer: AsyncWeightsBuffer
+
+    // This could rapidly grow so we should have some way to do garbage collection ?
+    // This could be interesting for measures based on gradient history
+    // Maybe we could have some sort of observer on the buffer
+    // so everytime we aggregate, we store the history
+    bufferHistory: Map<number, Map<string, any>>
```
Nacho114, line 10: It's a good idea to use types, as suggested by @tharvik even for basic things, I think this is a good example.
When you look at `bufferHistory: Map<number, Map<string, any>>`, you have no idea of what's inside.

but if you define say
```js
type age = number
type  name = string
const hist = map<age, name>
```

becomes more readable, it's even more so the case when you embed maps in maps, might even be worth to define the map as a type :)

Nacho114:
Starting to take a nice shape!

Nacho114:
I think you can open the PR soon :) , tests are passing 🥇

Nacho114:
@tharvik since you've been working on the server side, you can review his server code. (added some initial comments already)

mobile-browser-based-version/server/src/logic/federated/async_weights_buffer.ts
Nacho114: also, if you can make this private / remove `_` suffix

Nacho114 (approved):
lgtm!

PR #252, opened 2022-03-29
Fix for issue 219: NAN in the training graph after stoping.

Nacho114:
Fix for issue 219: NAN in the training graph after stoping.

Basically mid training we are able to stop training, however at this point onEpochEnd might be reached, and because of this the train accuracy is NAN, to fix the graph displaying NAN values and doing funny things I only update the graph if the values are not NAN. (seems like an easy reasonable fix).

I removed an intermediary `updateGraph` function, since the parameters were inverted in the function call, in the UI, validation accuracy was actually train accuracy and vice-versa. Since the function was redundant (and being used wrongly) to begin with I removed it and made explicit calls to updateTrainGraph, and updateValid...

I removed redundant code from the old interoperability feature.

Fixes 219

### DIFF
--- a/mobile-browser-based-version/src/components/building_frames/TrainingInformationFrame.vue
+++ b/mobile-browser-based-version/src/components/building_frames/TrainingInformationFrame.vue
@@ -1,129 +1,128 @@
 <template>
-<div>
+  <div>
+    <div
+      x-transition:enter="transition duration-300 ease-in-out"
+      x-transition:enter-start="opacity-0"
+      x-transition:enter-end="opacity-100"
+      x-ref="trainingBoard"
+      x-show="isTraining"
+    >
+      <!-- Validation Accuracy users chart -->
+      <div class="grid grid-cols-2 p-4 space-x-4 lg:gap-2">
+        <div class="col-span-1 bg-white rounded-md dark:bg-darker">
+          <!-- Card header -->
+          <div class="p-4 border-b dark:border-primary">
+            <h4 class="text-lg font-semibold text-gray-500 dark:text-light">
+              Validation Accuracy of the Model
+            </h4>
+          </div>
+          <p class="p-4">
+            <span class="text-2xl font-medium text-gray-500 dark:text-light">{{
+              currentValidationAccuracy
+            }}</span>
+            <span class="text-sm font-medium text-gray-500 dark:text-primary"
+              >% of validation accuracy</span
+            >
+          </p>
+          <!-- Chart -->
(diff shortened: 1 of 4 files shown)


Nacho114:
As a side note, I modified `src/components/building_frames/TrainingInformationFrame.vue`, and this was linted in the process. It seems that perhaps this was not done before. @tharvik should it not have been linted with the github actions? (Or maybe my own machine is not rightly set up? Although looking at the diff, it looks like the lint my machine did is the right thing.

tharvik:
> As a side note, I modified `src/components/building_frames/TrainingInformationFrame.vue`, and this was linted in the process. It seems that perhaps this was not done before. @tharvik should it not have been linted with the github actions? (Or maybe my own machine is not rightly set up?

From [what I gathered](https://cli.vuejs.org/core-plugins/eslint.html), there is not linter in place for HTML/vue file. We can add a linter for that, such as [prettier](https://prettier.io/), to ensure that.

> Although looking at the diff, it looks like the lint my machine did is the right thing.

btw, you can mask the whitespace changes in GitHub when seeing the diff, by clicking on the "setting cog" on the same line as "review changes"

Nacho114:
> > As a side note, I modified `src/components/building_frames/TrainingInformationFrame.vue`, and this was linted in the process. It seems that perhaps this was not done before. @tharvik should it not have been linted with the github actions? (Or maybe my own machine is not rightly set up?
> 
> From [what I gathered](https://cli.vuejs.org/core-plugins/eslint.html), there is not linter in place for HTML/vue file. We can add a linter for that, such as [prettier](https://prettier.io/), to ensure that.
> 
> > Although looking at the diff, it looks like the lint my machine did is the right thing.
> 
> btw, you can mask the whitespace changes in GitHub when seeing the diff, by clicking on the "setting cog" on the same line as "review changes"

thanks for the tip! 

As for the linting, we are using plugin:vue/essential (see [here](https://github.com/epfml/disco/blob/develop/mobile-browser-based-version/.eslintrc.json), I thought this would also deal with html.

tharvik:
> we are using plugin:vue/essential, I thought this would also deal with html.

Héhé, just found smth: replace `plugin:vue/essential` by `plugin:vue/recommended`, that does the trick :)

Nacho114:
> `recommended`

Ah nice, I'll add this to this PR then. thanks :)

Nacho114:
Actually I'll do a separate PR for this since a lot of files will be linted due to this change.

tharvik (approved):

PR #254, opened 2022-03-29
added recommended option to eslint to lint html code as well

Nacho114:
added recommended option to eslint to lint html code as well (as per @tharvik suggestion :) )

TODO: Fixing warnings + legacy code

Note, by activating the recommended plugin more warning were enabled, I discovered that in some .vue files we were still using jsx for the scripts (I had done this during the merge). When switching TS on I noticed several instances of calling functions without the required params or params of the wrong type. Currently cleaning this.


s314cy (approved):
LGTM

s314cy:
Note: This PR opts for the camel case naming style for components (name field), however we can still reference them in component templates by their snake case names (Vue supports this).

PR #255, opened 2022-03-29
peerjs -> simplepeer

### LINKED ISSUE #224: Simple peer
For decentralised we currently use peerjs, as noted by @giordano-lucas, this has two issues:

- It is no longer maintained.
- It needs browser integration.

Both are problematic, due to the latter this means we can have 0 automatic tests. Since decentralised is an important part of the platform it is worth using a solid api for it.

tharvik:
* rewrite the decentralized client
* server: add a signaling server via WS for webRTC peers
* avoid diff-ing npm package lock
* simplify network serialization
* server: fix compilation of MetadataMap handler

and with that, a bunch of TODOs
* drop authentication support, put it back (but secure this time) when need arise
* check that average works, can one use `tf.layers.average` to do so?
* clean old peers, both in server and client


mobile-browser-based-version/src/core/communication/decentralized_client.ts
```diff
@@ -0,0 +1,308 @@
...
+type Weights = tf.LayerVariable[]
+type PeerMessage = { epoch: number, weights: SerializedVariable[] }
+
+// TODO take it from the server sources
+type PeerID = number
+type EncodedSignal = Uint8Array
+type ServerOpeningMessage = PeerID[]
+type ServerPeerMessage = [PeerID, EncodedSignal]
+
+function isPeerMessage (data: unknown): data is PeerMessage {
```
Nacho114, line 25: Oh! This is cool, I didn't know the return type could be a statement like that! (I guess implicitly it's a boolean? but we show explicitly what we are checking?)

tharvik: > Oh! This is cool, I didn't know the return type could be a statement like that!

Yep, that's a very nice feature of TS, look at https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates for more information :)

> I guess implicitly it's a boolean?

Indeed, it's common for types in JS to be checked this way (`Array.isArray`, `Buffer.isBuffer`, ...).

> but we show explicitly what we are checking?

This way, the compiler knows that when branching with this function, the object is of this type (± implementing the `is` function correctly)

```
if (!isPeerMessage(msg)) {
  throw new Error('msg is not valid')
}

// here `msg` is of type `PeerMessage`
```

Nacho114 (approved):
Looks good, the addition of `immutable` is very nice, and also the use of narrowing for type guard is awesome! As discussed, it will be worth to look into the current averaging method used as mentioned in your comment and our meeting.

Did not go too much in detail as there were a lot of changes and there is a lot to do, but for smaller PRs we should strive to check code more in detail.

Issue #243, opened 2022-03-22
Remove i18n from disco


Nacho114:
We used i18n to take care of the merge of DeAI and FeAI, however its use is now redundant and just adds extra (non trivial) amount of code / packages / files to disco, it should be removed. 

It was a "hack" to basically choose the strings based on the platform, but its originally meant for translations and there are easier ways to do this. More importantly, few strings change between the platform so using a whole package is unnecessary.

Nacho114:
New issue from @dgengler6 has a better approach to this

PR #265, opened 2022-04-07
Refactor : Simplify usage of platform by getting rid of platform in the store 

### LINKED ISSUE #262: Simplifying the Platform usage
Due to the merge of DeAI and FeAI, the specification of platform is a bit convoluted. We currently choose the platform via the side-bar and this is stored in the `store`. At the end we should just choose the training type at the very end, i.e. in the trainingFrame, there is no reason to store the platform in the store. As a side note, we when doing this, in the training manager there is a separate flag for training alone, but this could be integrated with the platform so that there are 3 types, the 2 original + local (train alone).

dgengler6:
Fixes #262 #243

**What this does ?**
- Remove the Federated/Decentralized split in platform
- Remove platform from the store and leaves it up to the training manager 
- Added both `Train Federated` and `Train Decentralized` buttons to the training frame
- The client is created when one of these options is chosen
- Remove platform switch button in settings
- Create Locales files, for now there is only english but now we are open to translation
- It would be great to integrate all text of the app in the I18n.json and keep this as a good practice
- Change platform I18n to an actual locale manager for later translations ( One could create a french.json file, copy pasting default.json (english locale) and translate the texts )

**TODO steps**

- [x] Training platform is not dependent on the store anymore
- [x] Adapt Training frame so that we can chose between fed/dec
- [x] Remove platform completely from the store
- [x] Adapt I18n content so they fit the hybrid platform
- [x] Add scheme field to the task
- [x] Create english locale and lay the base for translation


dgengler6:
There is still a reference to the platform in the store because it is used for TaskCreationForm, we can change this I am just not sure what is the use of that there.

martinjaggi:
thanks for doing this, this will make things simpler indeed.

while you're already doing it, you could simplify even more, we actually don't need any UI element.
you could just add a boolean field to the task config (each task will be predefined either federated or decentralized). 

the reason is that we don't want to allow that people join the same task in potentially different modes. if both modes are wanted people should create 2 separate tasks, one federated one decentralized

Nacho114:
> thanks for doing this, this will make things simpler indeed.
> 
> while you're already doing it, you could simplify even more, we actually don't need any UI element. you could just add a boolean field to the task config (each task will be predefined either federated or decentralized).
> 
> the reason is that we don't want to allow that people join the same task in potentially different modes. if both modes are wanted people should create 2 separate tasks, one federated one decentralized

I agree, but there are different ways to implement this, might be good to brainstorm together. The training scheme choice is an important part of Disco so it's good to spend some time discussing on how it is being offered to the user.

dgengler6:
So last push changed quite a lot. 

- Now we only have one I18n (i.e default.json) which actually corresponds to the english locale
- the platform directory is now called locales for obvious reasons
- I renamed trainingType into trainingSchemes and moved it in the training folder
- Fully removed any reference to store.getters.platform

Nacho114:
You can also add in the description that it fixes #243, that issue was in regards to cleaning i18n which is done here.

mobile-browser-based-version/src/components/building_frames/containers/TrainingFrame.vue
```diff
@@ -21,15 +21,21 @@
...
             :center="true"
-            @click="training_manager.joinTraining(false)"
+            @click="training_manager.joinTraining(localScheme)"
           >
             Train Locally
           </custom-button>
           <custom-button
             :center="true"
-            @click="training_manager.joinTraining(true)"
+            @click="training_manager.joinTraining(decentralizedScheme)"
```
Nacho114: @s314cy @dgengler6  maybe we could add:

joinFederatedTraining(), joinDecentralizedTraining(), joinLocalTraining(), into the training manager, that way we don't pass arguments and it's resolved in the training_manager?

martinjaggi:
thanks!
BTW the vue components where we join training maybe don't even need to know if it's federated or not.

it could be figured out later in `training_manager`. so instead of `training_manager.joinFederatedTraining()` one could just make a `training_manager.joinTraining()` and that one would then look at the task config so it would know what to join (federated or decentralized). training alone is different and has it's own button as usual

Nacho114:
> thanks! BTW the vue components where we join training maybe don't even need to know if it's federated or not.
> 
> it could be figured out later in `training_manager`. so instead of `training_manager.joinFederatedTraining()` one could just make a `training_manager.joinTraining()` and that one would then look at the task config so it would know what to join (federated or decentralized). training alone is different and has it's own button as usual

yeah i agree, each component should know as little as possible 👍🏼

dgengler6:
Done, Although I think we should be clear that we should be clear about what is happening under the hood somewhere (i.e the toast alerts) 
Tbh I don't know if having two times the same task, but one Federated and one Decentralized makes much sense, it might just be confusing for the user..

Nacho114:
> Done, Although I think we should be clear that we should be clear about what is happening under the hood somewhere (i.e the toast alerts) Tbh I don't know if having two times the same task, but one Federated and one Decentralized makes much sense, it might just be confusing for the user..

hmm yeah, UI design is not obvious, we can have a discussion on Tuesday -- but keeping it simple is a good rule of thumb, although what simple means is what's hard to figure out

dgengler6:
I think I managed to merge everything with the new architecture ! 

Some tests don't pass but I'll fix that asap, It might have smthg to do with the training errors we had.

s314cy:
looking really nice! on a side note, the `activePage` store has been removed on the current develop version, as the components now communicate between each other instead

would you like to meet with @Nacho114 and me to ensure the CI tests are successful on your next commit?

dgengler6:
> looking really nice! on a side note, the `activePage` store has been removed on the current develop version, as the components now communicate between each other instead
> 
> would you like to meet with @Nacho114 and me to ensure the CI tests are successful on your next commit?

Yeah we can meet a bit later this afternoon !

Nacho114 (changes requested):
Nice work :)

I tried it out but there is an issue with the stop button not appearing after starting training. Once this is fix we will rebase it with devdelop!

Nacho114 (approved):
Great simplification, good work :)

martinjaggi:
thanks a lot @dgengler6 !

Issue #261, opened 2022-04-04
Disco benchmark


Nacho114:
As suggested by @martinjaggi, it would be good to have a performance benchmark of certain tasks, that way when we start trying out different things we see how performance is affected. 

At a first time performance will be mainly in terms of prediction accuracy, but we should also include run time performance in order to see what the bottle necks are. 

The idea is to have a `benchmark/` folder living at the `mobile-...`, it will be similar to the `tests` folder except we want to know how well it works (rather than the simple question of "does it work?").

The first benchmark will be on [face detection](https://www.kaggle.com/frabbisw/facial-age) since this is also useful for our start up collaborators. We can then extend this to include CIFAR-10 and other standard data sets.

To make this extendable for future use we will create a simple infrastructure that allows to simulate several clients.

dgengler6:
That is really great, I was talking about it at the beginning of the semester because it would be great for performance evaluations of personalization schemes ! 

I think it's really important that it generates log files so we can plot that and have a use for it ! 

I can give you a hand if you want @Nacho114 !

Nacho114:
Great 😃, we can discuss this when we meet later!

s314cy:
Linked to #240.

PR #270, opened 2022-04-08
Benchmark: benchmark framework for disco

Nacho114:
This PR introduces a `benchmark` framework for Disco.

The goal of the `benchmark` is to easily test the performance of distributed training among multiple clients and different setups. The benchmark suit has to be fully automated in order to ensure that results can be *reproduced* and easily experimented with. It will have:

- Config for the experimental setup (number of users, etc) -> this is easily extendable to add other things in the future such as delays for certain simulated users.
- Benchmark manager that takes care of the simulation
- Specialized logger that takes the output of train and validation accuracy
- io to store the performance

What is still missing is:

- [x] Make plots using nodejs (or maybe python)
- [ ] Integrate it with Disco (Currently it is mocked)
- [ ] Need to get ts-node working, since this is not really a test, but due to issue with the .env config I have not yet done it.

Solves #240 
Solves #261


Nacho114:
Currently porting this PR into a new one since develop has changed too much

Nacho114:
Will come later, too outdate and other priorities

PR #274, opened 2022-04-11
Image training on Node.js

s314cy:
This PR enables training from an image dataset on Node.js. This is ensured by two new tests:
1. loading a small subset of CIFAR10
2. training from this small dataset

@Nacho114 wdyt of test 2 (`tests/core/image_loader.test.ts`)? You might want to reproduce the same code for the benchmarks, I'm afraid chai might not be reporting any error thrown during the training loop, despite the timer I set.


Nacho114:
yes perfect, this is what I had in mind -- was also using a timer; though we should get ts-node going to avoid using timers^^

Nacho114 (approved):
lgtm, nice new name for the training manager!

PR #285, opened 2022-04-21
browser: fix start training buttons

tharvik:
* `disco` object wasn't watch for changes, it is now
* rm old/unused training_manager.ts
* fix `memory` import in `ModelLibrary`
* throw error in `disco.startTraining`


Nacho114 (approved):
good

PR #289, opened 2022-04-23
browser: fix reactive training

tharvik:
* do not crash server on invalid msg
* use immutable in some other places
* browser: watch training informant correctly


Nacho114 (approved):
looks good! thanks

PR #290, opened 2022-04-25
fix eslint and strict discojs

tharvik:
I broke eslint when moving files around. This PR renables that. Also, it adds `strict` to discojs.

@s314cy I updated `discojs/src/dataset/data_loader/image_loader.ts`, please check that I didn't break anything


s314cy (approved):
LGTM

Issue #295, opened 2022-04-27
Bug fix - federated (serialized weights of wrong type)


Nacho114:
This is the bug fix I mentioned in the slack channel, it has been fixed by @tharvik and co.

Basically, the refactoring was only tested on titanic, which has a single layer, and serialization model types were define based on this, which was then incompatible with Cifar10 and other models. 

The correct code can be found in `NAN-benchmark-nacho`, we need to port these changes to develop to get everything working again.

PR #300, opened 2022-04-28
Serialization fixes

tharvik:
When creating #295, it came out that serialization was not working well enough. Fixing that and also adding back some removed files/lines.

* serialization
  * add a round-trip test
  * use "encode/decode" instead of "serialize/deserialize"
  * directly `msgpack.{en,de}code` there
  * readd `isSerializedWeights` body
*  readd discojs/aggregation & server/router/connection tests


Nacho114:
Can you remove: the connection.test.ts; I merged it with federated.test.ts (since in connection.test.ts we were just testing for federated anyway)

Nacho114 (approved):
lgtm!

PR #301, opened 2022-04-29
281 batches nacho

### LINKED ISSUE #281: RoundDuration measure: switch from epochs to batches
RoundDuration measure: switch from epochs to batches

Batches is a more natural measure for round duration. This way all clients have the same frequency of communication (if it's in terms of epochs it depends on their data size).

TODO: do these changes in the `RoundTracker`

----

What is the right measure for the round duration?

Right now it's in epochs, that means if roundDuration = .5, then every half epoch we communicate.

We could do batches instead, and say, every X batches we communicate, but in this case we need to take into account the batchSize.

I guess ideally a round of communication corresponds to a fixed amount of computation? I have not read / looked into this question.

Nacho114:
Simplify the roundTracker, now rounds are in batches, no longer depend on size of data to compute epoch length. 

fixes #281

### DIFF
--- a/browser/src/training/disco.ts
+++ b/browser/src/training/disco.ts
@@ -123,7 +123,7 @@ export class Disco {
 
     // TODO: dataset.size = null, since we build it with an iterator (issues occurs with ImageLoader) see issue 279
     await this.initTrainer()
-    await this.trainer.trainModel(data.dataset, data.size)
+    await this.trainer.trainModel(data.dataset)
   }
 
   /**
--- a/browser/src/training/trainer/round_tracker.ts
+++ b/browser/src/training/trainer/round_tracker.ts
@@ -3,25 +3,20 @@
  *
  * @remark
  * In distributed training, the client trains locally for a certain amount of epochs before sharing his weights to the server/neighbor, this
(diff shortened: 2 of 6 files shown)


martinjaggi:
looking good

small comment is that in the UI, the train&test accuracy is still reported every epoch, not every round i think. want to change that to be also every round? (or every 10 rounds if it's costly)

Nacho114:
The issue is that we only get the train accuracy from model.fit every epoch, so it's not immediate to switch to per round accuracy reporting. But yes, I agree that every n round reporting of accuracy would be better. But maybe we should keep it given that changing this will take some effort (and maybe not worth given the time it takes).

martinjaggi:
ok fine, as long as we don't need to know the data size n ourselves, which was tedious.

you're saying things already work fine in the UI both for distr and training alone, like now, right?

Nacho114:
yeah, we don't need to know the data size at all anymore.

Yes, right now the UI accuracy for train/test should be working (I have not checked with the latest PR, but it was at the end of the week). Will test again, and if it works I'll update the site.

PR #309, opened 2022-05-04
discojs supports node

tharvik:
* move `Client` implementations, many tests, tasks and training data to discojs
* discojs is built for node and webpacked in the browser
* fix training scheme selection
* add local `Client`, doing nothing
* `Disco` is now generic per scheme
* benchmark don't need browser, only server, move it there (but it's missing some files to run)


martinjaggi (approved):

discojs/src/memory/base.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  /**
+ * Adds the current working model to the model library.
+ * @param taskID the working model's corresponding task
+ * @param modelName the working model's file name
+ */
+  abstract saveWorkingModel (taskID: TaskID, modelName: string): Promise<void>
+
+  /**
+ * Removes the model model.
```
martinjaggi: minor typo, 2x.
also you didn't document the `ModelType` param in this one and some above one. but fine anyway :)

martinjaggi (approved):

martinjaggi:
impressive work, thanks a lot!

discojs/src/memory/base.ts
tharvik: Typo fixed & `ModelType` commented :)

PR #317, opened 2022-05-13
Differential Privacy

### LINKED ISSUE #303: Add differential privacy
image (1)

s314cy:
Closes #303

### DIFF
--- a/browser/src/communication/federated/federated_client.ts
+++ b/browser/src/communication/federated/federated_client.ts
@@ -11,9 +11,9 @@ import { Weights } from '@/types'
  * a specific task.
  */
 export class FederatedClient extends Client {
-  clientID: string;
-  peer: any;
-  modelUpdateIsBasedOnRoundNumber: number;
+  clientID: string
+  peer: any
+  modelUpdateIsBasedOnRoundNumber: number
 
   /**
    * Prepares connection to a centralized server for training a given task.
@@ -63,7 +63,7 @@ export class FederatedClient extends Client {
     return response.status === 200
   }
 
-  async postMetadata (metadataID, metadata): Promise<boolean> {
+  async postMetadata (metadataID: string, metadata: string): Promise<boolean> {
     const response = api.postMetadata(
       this.task.taskID,
       this.modelUpdateIsBasedOnRoundNumber,
@@ -74,7 +74,7 @@ export class FederatedClient extends Client {
     return (await response).status === 200
   }
 
-  async getMetadataMap (metadataID) {
+  async getMetadataMap (metadataID: string) {
     const response = await api.getMetadataMap(
       this.task.taskID,
       this.modelUpdateIsBasedOnRoundNumber,
--- a/browser/src/communication/federated/federated_client.ts
+++ b/browser/src/communication/federated/federated_client.ts
@@ -1,14 +1,14 @@
 import * as msgpack from 'msgpack-lite'
-import * as tf from '@tensorflow/tfjs'
 
-import { Client, Task, TrainingInformant } from 'discojs'
+import { Client, Task, TrainingInformant, privacy } from 'discojs'
 
 import { makeID } from '../authentication'
 import * as api from './federated_api'
 import { Weights } from '@/types'
+
 /**
  * Class that deals with communication with the centralized server when training
- * a specific task.
+ * a specific task in the federated setting.
  */
 export class FederatedClient extends Client {
   clientID: string
@@ -17,10 +17,10 @@ export class FederatedClient extends Client {
 
   /**
    * Prepares connection to a centralized server for training a given task.
-   * @param {String} serverURL The URL of the centralized server.
+   * @param {string} serverURL The URL of the centralized server.
    * @param {Task} task The associated task object.
    */
-  constructor (serverURL: string, task: Task) {
+  public constructor (serverURL: string, task: Task) {
     super(serverURL, task)
     this.clientID = ''
     this.modelUpdateIsBasedOnRoundNumber = -1 // The server starts at round 0, in the beginning we are behind
@@ -30,7 +30,7 @@ export class FederatedClient extends Client {
    * Initialize the connection to the server. TODO: In the case of FeAI,
    * should return the current server-side round for the task.
    */
-  async connect (): Promise<void> {
+  public async connect (): Promise<void> {
     /**
      * Create an ID used to connect to the server.
      * The client is now considered as connected and further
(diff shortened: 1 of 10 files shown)


martinjaggi (changes requested):
looking good. 2 things
- you're doing it for the weights, not the \delta w (gradients). while for adding noise that can be fine, the clipping unfortunately is not ok on models. that would need to be on gradients. we need to briefly discuss maybe if in FL we want to send gradients or models. without DP, the result in both cases is identical anyway. we might later also want to support byzantine robustness, which also likes to have clipping on gradients, not models. for the moment what you could do is keep sending models, but before you do, the 2 new ops you now added should happen only on the local gradients, not the models
- add a field to the task config where people can switch it on/off andspecify the noise level

s314cy:
> looking good. 2 things
> 
>     * you're doing it for the weights, not the \delta w (gradients). while for adding noise that can be fine, the clipping unfortunately is not ok on models. that would need to be on gradients. we need to briefly discuss maybe if in FL we want to send gradients or models. without DP, the result in both cases is identical anyway. we might later also want to support byzantine robustness, which also likes to have clipping on gradients, not models. for the moment what you could do is keep sending models, but before you do, the 2 new ops you now added should happen only on the local gradients, not the models
> 
>     * add a field to the task config where people can switch it on/off andspecify the noise level

I fully agree. At first glance, there doesn't seem to be a way to access the current epoch's gradient, which is why @Nacho114 and I decided to go with the weights. Would taking the difference W_curr - W_prev be correct? This doesn't take into account the current learning rate.

I'll add the task config field for the noise scale 👍

martinjaggi:
exactly. the diff curr-prev (over the round) is perfectly fine. then clip etc. if we do it like this, the nice thing is that it will work easily for both federated and decentralized

martinjaggi:
actually you can keep sending weights (model) instead of diffs.

it's just that locally before sending, the client needs to compute their own weight diff since beginning of the round, and then apply noise and clipping. the result can still be the weights (model)

maybe easier this way because then it will be the same in decentralized

s314cy:
> actually you can keep sending weights (model) instead of diffs.
> 
> it's just that locally before sending, the client needs to compute their own weight diff since beginning of the round, and then apply noise and clipping. the result can still be the weights (model)
> 
> maybe easier this way because then it will be the same in decentralized

Oh right, I see

browser/src/communication/federated/federated_client.ts
```diff
@@ -119,13 +123,34 @@ export class FederatedClient extends Client {
     }
   }
 
+  private addDifferentialPrivacy (currentRoundWeights: Weights): Weights {
```
martinjaggi: make state-less, by also giving `previousRoundWeights = this.previousRoundWeights` as a param.

browser/src/communication/federated/federated_client.ts
```diff
@@ -119,13 +123,34 @@ export class FederatedClient extends Client {
...
+    const weightsDiff = List(currentRoundWeights)
+      .zip(List(this.previousRoundWeights))
+      .map(([w1, w2]) => w1.add(-w2))
+
+    const normBound = weightsDiff.map((w) => w.norm().dataSync()[0]).max()
```
martinjaggi: add a comment what kind of norm you compute. why is there a max? norm is 2-norm but at the end gets maxed, like infty-norm?

browser/src/communication/federated/federated_client.ts
```diff
@@ -119,13 +123,34 @@ export class FederatedClient extends Client {
...
+
+    const noisyWeightsDiff = weightsDiff.map((w) => {
+      const clipped = w.div(Math.max(1, w.norm().dataSync()[0] / normBound))
```
martinjaggi: the clipping radius (the maximum norm the grad can be after clipping) should be a parameter in the task config as well. if set to infty, there will be no clipping

browser/src/communication/federated/federated_client.ts
```diff
@@ -119,13 +123,34 @@ export class FederatedClient extends Client {
...
+      const noise = tf.randomNormal(w.shape, 0, normBound * normBound)
```
martinjaggi: rename normBound -> norm ?  also make a configurable scalar parameter (from taskconfig) in front of this one? if zero, there will be no noise added (actually pack this one into an if-statement, as it's probably quite slow if there are millions of weights. rand generators can take quite some time.  this way the DP can be switched on / off in the task config (also add a comment to explain this in the config and here)

browser/src/communication/federated/federated_client.ts
```diff
@@ -119,13 +123,34 @@ export class FederatedClient extends Client {
...
+
   private async fetchServerStatisticsAndUpdateInformant (trainingInformant: TrainingInformant) {
     const serverStatistics = await api.getAsyncWeightInformantStatistics(this.task.taskID, this.clientID)
     trainingInformant.updateWithServerStatistics(serverStatistics.data.statistics)
   }
 
   async onRoundEndCommunication (model: tf.LayersModel, _: number, trainingInformant: TrainingInformant) {
-    await this.postWeightsToServer(model.weights.map((w) => w.read()))
+    const weights = model.weights.map((w) => w.read())
+    const noisyWeights = this.addDifferentialPrivacy(weights)
```
martinjaggi: as mentioned, better stateless with 2 params (prev and curr weights).

martinjaggi (approved):
very nice. only some cosmetic comments.
also did you try run it and observed any effect on accuracy? does it work on titanic and image tasks?

browser/src/communication/federated/federated_client.ts
s314cy: that's an error on my side! will push changes soon

browser/src/communication/federated/federated_client.ts
s314cy: I agree on the benefits of stateless-ness, however `Client` is the class actually updating the `Trainer`'s model, hence the one which has direct access to both the previous and current model. Moving the state to `Trainer` would require more effort and might not be cleaner 🤔 

EDIT: I'll try a few things

browser/src/communication/federated/federated_client.ts
s314cy: do we still want clipping when there is no noise added?

martinjaggi: these can be independent. the default in the task config would be `noiseLevel=0` and `clipping=infty`. so both deactivated. so if people want they can activate them separately (clipping gives robustness for example)

s314cy:
closing this, will open create another branch since @tharvik's latest PR moved many files

martinjaggi:
exactly. the diff curr-prev (over the round) is perfectly fine

PR #321, opened 2022-05-24
Fixes and refactoring for discojs supports node #309 

Nacho114:
In one of the last big refactor PR's the end 2 end tests were removed by accident (these need to be added again); part of the refactoring was for memory, but this broke the browser-less version, which is mostly fixed in this PR. This is done partially with the help @tharvik 

Done: 

- [x] Fixed memory class to return undefined for browser-less
- [x] Reduced server memory overhead by clearing created models 
- [x] Refactored  server app engine into functions (avoid code being executed just through imports"

TODO:

- [x] Add again end to end tests 
- [x] fix the 2 user case of benchmark.ts, current issue is related with tfjs engine "duplicate variable".


tharvik:
* add serialization for tf models
* rebase sur develop and some fixes
* use interfaces instead of classes for Task
* server: keep models in memory
* default models & tasks instead of using tasks.json
* do not yet allow to store new tasks on disk
* bunch of reordering
* split routers in various files

PR #322, opened 2022-05-25
Add testing component

### LINKED ISSUE #314: Testing page
The testing page is not working at the moment.

While fixing it, one should come up with clear/intuitive rules about what model can be tested (working models? saved models? what about working models currently being trained? etc.)

s314cy:
Closes #314. This PR brings changes to the browser part of Disco.

* Progress bar
  * Move the progress bar to a named router view, which allows for conditional render on the default router view
  * This ensures the `TaskList` and `Navigation` components share the same `ProgressBar`, which fixes an animation bug and corresponds better to the "Vue way" of doing things
* Task routes (list & nav)
  * Change the task routes from dynamic to simple routes with params
  * Fetch server information before render of the app
  * This fixes the 404 on loading the task list or a task's navigation page, and corresponds better to the "Vue way" of doing things
* Add testing
  * New "finished" step redirecting to the testing pipeline
  * New testing pipeline parallel to the original training pipeline
  * Working models are available in the model library
* Tiny model caching fix following @tharvik's PR
* Components
  * Remove redundant ID prop
  * Rename remaining components by removing the "Frame" suffix
* Color theme
  * Remove calls to and CSS related to unsupported colors
  * Remove CSS related to the dark theme. TODO: add an issue for a new dark theme following the UI updates
* UI additions & fixes I'll let you nagivate through :)


browser/src/components/Finished.vue
```diff
@@ -0,0 +1,93 @@
...
+  },
+  methods: {
+    ...mapMutations(['setTestingModel']),
+    testModel () {
+      const path = pathFor(ModelType.WORKING, this.task.taskID, this.task.trainingInformation.modelID)
+      this.setTestingModel(path)
+      this.$router.push({ path: '/testing' })
+    },
+    async saveModel () {
+      if (this.memory !== undefined) {
```
tharvik: `this.memory` is `undefined` only at the beginning. you probably want to use `instanceof` to determine if the memory is `Emtpy`

Nacho114 (approved):
lgmt!

tharvik (commented):
* "Test your model" part is empty for me, is it intended
* one can move the "Test your model" part without connecting any data

Otherwise, LGTM

s314cy:
> * "Test your model" part is empty for me, is it intended
> 
>     * one can move the "Test your model" part without connecting any data
> 
> 
> Otherwise, LGTM

* yup that's intended, this PR brings other relevant changes so better merge it asap :)
* navigating forward is currently unconditional, i left it this way for now as it does not break anything (although counter-intuitive)

browser/src/components/Finished.vue
s314cy: thanks!

PR #330, opened 2022-06-13
Fix training in browser

s314cy:


Nacho114 (approved):
lgtm

martinjaggi:
we can still rename some `feai` to `disco`

s314cy:
> we can still rename some `feai` to `disco`

ah you're right, i'll make sure the name's changed in my next PR

PR #331, opened 2022-06-13
cosmetics on descriptions & documentation

martinjaggi:


s314cy (approved):

gozgun (commented):
Everything is much more clear, thank you. I will update the text under informationCard in default.json located in the browser folder according to the changes made in the default.json located in discojs.

martinjaggi:
i think i changed that too, but maybe i didn't. there is some duplication between the two locale .json files. should we reduce that somehow, or make them identical?

also feel free to add back the 4th of the 3 cards, if it looks better. as you want

Issue #332, opened 2022-06-14
Release v1.0


tharvik:
- [x] create NPM organisation
- [x] rename `discojs` to `@[NPM-org]/discojs` in package.json and deps using it
- [x] improve readme to explain API
  - [x] add a few line on the Node API @tharvik 
- [x] test and merge https://github.com/epfml/disco/pull/333
- [x] push develop to production
- [x] tag as `v1.0.0` production's HEAD

Nacho114:
Very nice; Might also be able to add the testing frame @s314cy is working on!

Nacho114:
It would also be good to add documentation of how to use discojs via npm in the discojs readme.

s314cy:
> Very nice; Might also be able to add the testing frame @s314cy is working on!

if i understand correctly, the npm package will only include discojs and not the browser

s314cy:
btw, do we agree on version 1.0.0 and not 0.1.0? since the package has never been used publicly, i would go for 0.1.0

Nacho114:
The v1 is for the github version tag, so it includes the browser.

0.1.0 sounds good to me

s314cy:
> The v1 is for the github version tag, so it includes the browser.
> 
> 1.0.0 sounds good to me

how does the npm release version relate to the version used for the github release? how would we manage the server's docker image version then?

tharvik:
Published on NPM as v1.0.0: https://www.npmjs.com/package/@epfml/discojs :partying_face:

PR #333, opened 2022-06-14
Release v1.0

tharvik:
See https://github.com/epfml/disco/issues/332

* [x] need `secrets.NPM_TOKEN` in github actions


s314cy:
`@epfml/discojs` is great!

martinjaggi:
in the readme i think `Tasks are stored as .ts classes` is unclear. tasks can come from 3 sources:
- node API where one can load some task and modify it, or create from scratch (is that correct?)
- predefined example tasks
- new tasks created in the UI form

only the second kind is in a .ts file.

can we extend this documentation a bit, to make it easy for users to deal with it? (as it's very crucial functionality)

martinjaggi (approved):
excellent, thanks a lot!

PR #334, opened 2022-06-14
Fix testing redirection & model library

### LINKED ISSUE #315: working model also visible in model library
the working model from each task should also be visible in the model library  - this way it can be accessed for testing.

this is also to simplify the training process, as then it's clear that the 'save model' button is optional, and users still get the main functionality if they don't use that one.

one model per task is a small number so doesn't add too much confusion. maybe can optically highlight the saved models differently form the working models, but functionality is identical anyway

s314cy:
There were some remaining bugs related to IndexedDB and the Vue store. The changes are:
- finished step
  - redirects to the correct model in the "model testing" page
  - correctly initializes its memory
  - handles untrained tasks
- model library
  - displays both working and saved models (closes #315)
  - correctly deletes models
  - is consistent w.r.t. IndexedDB content
 
Additionally:
- the UI for information pages uses a single `BaseLayout` to avoid duplicate footers
- the testing page is displayed correctly when no model is available
- federated client starts at round 0 instead of -1 to avoid stalling in the beginning and thus losing a round

### DIFF
--- a/browser/src/components/Finished.vue
+++ b/browser/src/components/Finished.vue
@@ -52,25 +52,22 @@ export default {
       default: undefined
     }
   },
-  data (): { memory: Memory } {
-    return {
-      memory: undefined
-    }
-  },
   computed: {
-    ...mapState(['useIndexedDB'])
-  },
-  watch: {
-    useIndexedDB (newValue: boolean) {
-      this.memory = newValue ? new IndexedDB() : new EmptyMemory()
+    ...mapState(['useIndexedDB', 'models']),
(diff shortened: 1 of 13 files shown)


martinjaggi:
looking good!
does it also work if i start from the library page, select a model there and that one can be tested?

s314cy:
> looking good! does it also work if i start from the library page, select a model there and that one can be tested?

nice catch! it does now :)

Nacho114 (approved):
lgtm

PR #338, opened 2022-06-16
Implement model validation backend

s314cy:
Implement the testing backend, which outputs the accuracy on the given test dataset


s314cy:
the backend code is already reviewable, some UI code remains to be done (please don't merge :))

martinjaggi:
in the code, could you call the components `validation` instead of `testing/tester`? the latter one is easy to confuse in the software context. validation is fine as it can be used for both. 

by default if the user specifies no test dataset on that UI pane, then it will use the standard validation set (same as we had during training). which is another reason validation is a more clear name here. if the user wants to upload their own set, it's still ok to be called a new validation set

s314cy:
turns out the titanic model consistently yields NaN accuracies on both model training and model predictions, testing the new validation backend on `simple_task` works fine

Nacho114 (approved):
lgtm

PR #339, opened 2022-06-21
9 sec agg alyssa felix

aunell:
Branch for implementing the main secure aggregation code, first for the decentralized training mode


tharvik:
* server
  * start server via app, not ourself (allow for correct WebSocket server registration, thanks silent failure...)
  * generic `getClient`, just pass it the constructor
  * allow to set a dedicated server port for tests (way easier to play with it)
  * cleanup
* discojs/client/decentralized
  *  use `ws` protocol on `http` ones
  * do not crash when disconnecting
  * fix node/browser crosscompatiblity (`wrtc`)

martinjaggi (dismissed):
this starts to look good.

small comment: at several places you write `epoch` but you probably mean `round` as the latter is one unit of compute & communication. this is usually much shorter than an epoch. 

also can you add some more comments in the code to also make the 2 phases of each round a bit more transparent?

finally, is there any way we could reduce additional code even a bit more?

aunell:
> this starts to look good.
> 
> small comment: at several places you write `epoch` but you probably mean `round` as the latter is one unit of compute & communication. this is usually much shorter than an epoch.
> 
> also can you add some more comments in the code to also make the 2 phases of each round a bit more transparent?
> 
> finally, is there any way we could reduce additional code even a bit more?

Thanks for the feedback! We are using 'epoch' in the message interfaces because that is how it was described in isPeerMessage from the original decentralized implementation. As for comments and code reduction, we will do this once we finish passing some basic functionality tests.

martinjaggi (approved):
excellent work!

let's merge this one already, to make sure things remain aligned with dev branch. afterwards we can make some smaller PRs for polishing things further

PR #348, opened 2022-06-27
client/decentralized: fixes

tharvik:
rework discojs/client/decentralized to use WebRTC instead of tunneling message through the server

* discojs/client/dec/peer
  *  wraps simple peer for communication, with wrtc on node
  * fragment message on webrtc stream (it has a message size limit), and reorder theses on arrival (oc, it's not a stream..)
* discojs/client/dec/peer_pool
  *  to avoid reconnecting known peers, keep theses around
* discojs/client/dec/base
  *  don't send msgs to ourself
* discojs/client/dec/messages
  * are now ADTs (discriminated unions in TS, sealed case class in Scala), to have exhaustiveness checking and catch many early bugs, see https://www.typescriptlang.org/docs/handbook/2/narrowing.html#discriminated-unions
  * renamed to avoid too long names


martinjaggi (approved):

discojs/src/aggregation.ts
```diff
@@ -22,7 +22,7 @@ function applyArithmeticToWeights (peersWeights: List<Weights>, tfjsArithmeticFu
 }
 
 export function sumWeights (peersWeights: List<Weights>): Weights {
-  return applyArithmeticToWeights(peersWeights, tf.add)
+  return applyArithmeticToWeights(peersWeights.toList(), tf.add)
 }
 
 export function subtractWeights (peersWeights: List<Weights>): Weights {
```
martinjaggi: same here as well for subtraction? (though ok that method seems unused anyway and not clear)

martinjaggi (approved):

francesco98 (approved):

discojs/src/aggregation.ts
tharvik: good catch, it was dangling from a previous iteration; there is not need to convert a `List` to a `List` (:

discojs/src/client/decentralized/utils.ts
```diff
@@ -0,0 +1,23 @@
+// Time to wait between network checks in milliseconds.
+const TICK = 100
+
+// Time to wait for the others in milliseconds.
+const MAX_WAIT_PER_ROUND = 10_000
+
+// check if a given boolean condition is true, checks continuously until maxWait time is reached
+export async function pauseUntil (condition: () => boolean): Promise<void> {
```
francesco98, line 8: Is there a way to avoid this `pauseUntil` function? Maybe using some TS-oriented ways? I was trying to do that using Promises, or Events. It would be nice to drop this function because I don't really like this way of waiting until a condition is not true in TS.

server/package.json
```diff
@@ -20,7 +20,8 @@
     "express-ws": "5",
     "immutable": "4",
     "lodash": "4",
-    "msgpack-lite": "0.1"
+    "msgpack-lite": "0.1",
+    "wrtc": "0.4"
```
francesco98: This is creating some issues on M1 Mac...
https://github.com/node-webrtc/node-webrtc/issues/698

s314cy: I can confirm; they don't offer macOS M1 binaries. A possible work-around (according to the mentioned issue) would be to use Rosetta to install everything for an x64 arch. However, this conflicts with TF.js which does not work with Rosetta and requires to be installed on an arm64-arch running an arm64-node version.

discojs/src/client/decentralized/utils.ts
tharvik: I totally agree, that's quite a bad way to do async. I though about it but I didn't find an easy fix for that.. but there is a way :)

the issue is that data generation (the WS & peer connections) and data consumption (the other funcs) aren't linked. the best way would have to represent theses connections as an async stream/iterable/generator of messages, so that the consumers can await the next message and directly act on the received value, thus keeping the processing local to a function. that would also ensure that the server/peers are sending the correct sequence of messages

so I would look into turning theses connections into [async generators](https://javascript.info/async-iterators-generators#async-generators-finally), or a good old [EventEmitter](https://danilafe.com/blog/typescript_typesafe_events/)

martinjaggi: thanks, sounds like a good plan. we can make a separate issue for it, if we want to pursue it.

PR #366, opened 2022-07-07
Fix #358: Friendly home page

### LINKED ISSUE #358: Simplify the homepage (even more)
PROBLEM: The homepage currently has three options (_Start training -- Create new task -- Test your models_)
This is maybe a bit too much to parse for the new user (some questions I've had from others are: what is a task? what is a model? is a task a model? etc)

SOLUTION: 
- On the homepage, keep only the "_Start Training_" box with all its text (but change the button text to a more generic "_Get started_")
- This then redirects to a new page or a pop-up with three options using harmonized DISCO terminology 
1) **TRAIN** Join a DISCOllaborative to train an existing model
2) **TEST** Validate a DISCO-trained model
3) **CREATE** Add your own model to be trained in a DISCOllaborative

Add hover text to tell users that a DISCOllaborative is "_A group of (DIS)tributed users with similar data, (CO)llaborating to train a model_"

s314cy:
closes #358


Nacho114 (approved):
lvgtm

PR #372, opened 2022-07-12
Validation split

Nacho114:
Add validation split for images, this is straightforward since we simply split the indices of the image list. (Trickier to do for Tabular data as we stream the csv file, so it will be done in a seperate PR).

### DIFF
--- a/browser/src/components/training/Training.vue
+++ b/browser/src/components/training/Training.vue
@@ -33,7 +33,10 @@
 
     <!-- Training Board -->
     <div>
-      <TrainingInformation :training-informant="trainingInformant" />
+      <TrainingInformation
+        :training-informant="trainingInformant"
+        :has-validation-data="hasValidationData"
+      />
     </div>
   </div>
 </template>
@@ -106,6 +109,9 @@ export default defineComponent({
       }
       // default scheme
       return TrainingSchemes.LOCAL
+    },
+    hasValidationData (): boolean {
+      return this.task?.trainingInformation?.validationSplit > 0
     }
   },
   watch: {
--- a/browser/src/components/training/TrainingInformation.vue
+++ b/browser/src/components/training/TrainingInformation.vue
@@ -7,9 +7,15 @@
       x-ref="trainingBoard"
       x-show="isTraining"
     >
-      <div class="grid grid-cols-2 p-4 gap-8">
+      <div
+        class="grid p-4 gap-8"
+        :class="hasValidationData ? 'grid-cols-2' : 'grid-cols-1'"
+      >
         <!-- Validation Accuracy users chart -->
-        <div class="bg-white rounded-md">
+        <div
+          v-if="hasValidationData"
+          class="bg-white rounded-md basis-0"
+        >
           <!-- Card header -->
           <div class="p-4 border-b">
             <h4 class="text-lg font-semibold text-slate-500">
@@ -41,7 +47,7 @@
         </div>
 
         <!-- Training Accuracy users chart -->
-        <div class="bg-white rounded-md">
+        <div class="bg-white rounded-md basis-0">
           <!-- Card header -->
           <div class="p-4 border-b">
             <h4 class="text-lg font-semibold text-slate-500">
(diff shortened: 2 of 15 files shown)


discojs/src/dataset/data_loader/data_loader.ts
```diff
@@ -1,12 +1,17 @@
...
-export interface DataConfig { features?: string[], labels?: string[], shuffle?: boolean }
+export interface DataConfig {features?: string[], labels?: string[], shuffle?: boolean, valSplit?: number}
 export interface Data {
   dataset: Dataset
   size: number
 }
 
+export interface DataTuple {
+  train: Data
+  val: Data
```
s314cy: the validation set should be optional, no? btw i think giving it the full name "validation" is fine

browser/src/components/training/Training.vue
```diff
@@ -69,12 +72,14 @@ export default defineComponent({
...
-    trainingInformant: TrainingInformant
+    trainingInformant: TrainingInformant,
+    hasValidationData: boolean
     } {
     return {
       distributedTraining: false,
       startedTraining: false,
-      trainingInformant: new TrainingInformant(10, this.task.taskID, TrainingSchemes.LOCAL)
+      trainingInformant: new TrainingInformant(10, this.task.taskID, TrainingSchemes.LOCAL),
+      hasValidationData: this.task?.trainingInformation?.validationSplit > 0
```
s314cy: better off in the `computed` properties

browser/src/components/training/TrainingInformation.vue
```diff
@@ -9,7 +9,10 @@
     >
       <div class="grid grid-cols-2 p-4 gap-8">
```
s314cy: you can add the v-field `:class="hasValidationData ? 'grid-cols-2' : 'grid-cols-1'"` next to `class="grid p-4 gap-8"`

discojs/src/dataset/data_loader/data_loader.ts
```diff
@@ -1,12 +1,17 @@
 import { Dataset } from '../dataset_builder'
 import { Task } from '../../task'
 
-export interface DataConfig { features?: string[], labels?: string[], shuffle?: boolean }
+export interface DataConfig {features?: string[], labels?: string[], shuffle?: boolean, valSplit?: number}
```
s314cy: maybe name it `validationSplit` to be more expressive?

discojs/src/dataset/data_loader/tabular_loader.ts
```diff
@@ -75,15 +75,20 @@ export abstract class TabularLoader<Source> extends DataLoader<Source> {
...
+    const data = {
       dataset: config?.shuffle ? dataset.shuffle(BUFFER_SIZE) : dataset,
       // dataset.size does not work for csv datasets
       // https://github.com/tensorflow/tfjs/issues/5845
       size: 0
     }
+    // TODO: Implement validation split for tabular data (tricky due to streaming)
+    return {
+      train: data,
+      validation: data
```
s314cy: return `{ train: data }` so that the validation set is undefined and the graph is not displayed in the browser :)

s314cy (changes requested):
💯

discojs/src/dataset/data_loader/tabular_loader.ts
```diff
@@ -75,15 +75,20 @@ export abstract class TabularLoader<Source> extends DataLoader<Source> {
...
+    const data = {
       dataset: config?.shuffle ? dataset.shuffle(BUFFER_SIZE) : dataset,
       // dataset.size does not work for csv datasets
       // https://github.com/tensorflow/tfjs/issues/5845
       size: 0
     }
+    // TODO: Implement validation split for tabular data (tricky due to streaming)
+    return {
+      train: data,
+      validation: undefined
```
s314cy: better to simply omit the field, it defaults to `undefined` :) literally not defining it haha (this is what optional params and undefined mean)

s314cy (approved):

PR #380, opened 2022-07-14
Benchmark CLI

### LINKED ISSUE #261: Disco benchmark

Nacho114:
A simple CLI to easily test disco via the terminal, e.g. 

```
npm run benchmark -- --numberOfUsers 3 --epochs 2 --roundDuration 1
```

Once Decentralized is working again we can also include the training scheme as a flag.

This will make it much easier for developer / students to test out their features while they are building them.

Closes #261


s314cy:
super nice!! maybe add the new `result` folder to `.gitignore`? and remove it in a future commit to ensure it is not merged into develop :)

benchmark/README.md
```diff
@@ -31,8 +33,145 @@ npm run benchmark -- -u 4 -e 15 -r 5
 - `npm run benchmark` to run the benchmark with the default setting, to see the available flags run
 - `npm run benchmark -- --help`
 
-## Comming soon
+## Custom Tasks
+
+DiscoJS currently provides several pre-define popular tasks such as titanic, simple-face and cifar10. In order
+to understand how to add your own custom task, we will go over how we added simple-face to discojs.
+
+### Task
```
martinjaggi: this whole subsection here is very nice and useful, but it seems to be not node specific. we can put this in a more general readme how to define tasks (which can later be used both in browser and node)

benchmark/README.md
```diff
@@ -31,8 +33,145 @@ npm run benchmark -- -u 4 -e 15 -r 5
...
+
+The `Task` class is the first piece of the puzzle, this contains all the crucial information from training to mode.
+
+The task class for simple face can be found in [discojs/src/tasks/simple_face.ts](../discojs/src/tasks/simple_face.ts),
+and the contents look as follows. (For brevity we have replaced some lines with ...).
+
+Note that the preprocessFunctions are not yet supported via the Task object, we will explain after how to add a it
```
martinjaggi: instead of not supported, it's just that preprocessing must be done in/after the dataloader

benchmark/README.md
```diff
@@ -31,8 +33,145 @@ npm run benchmark -- -u 4 -e 15 -r 5
...
+
+```js
+export * as cifar10 from './cifar10'
+export * as lus_covid from './lus_covid'
+export * as mnist from './mnist'
+export * as simple_face from './simple_face' // <---- 
+export * as titanic from './titanic'
+```
+
+This allows us to access our task as via discojs:
```
martinjaggi: unclear what you here mean by access. i guess here is the point where one switches to now either use the new task in browser or in node (we can explain both). maybe also remind that one has to first build sth again, till it's usable

benchmark/README.md
```diff
@@ -31,8 +33,145 @@ npm run benchmark -- -u 4 -e 15 -r 5
...
 
-Currently only the simple-face task is supported but different tasks will be available through the CLI 
-in the future.
+```js
 
+import { tasks } from 'discojs'
+
+const TASK = tasks.simple_face.task
```
martinjaggi: not needed?

benchmark/README.md
```diff
@@ -31,8 +33,145 @@ npm run benchmark -- -u 4 -e 15 -r 5
...
+```
+
+> Note that you need to rebuild discojs every time you make changes to it (`cd discojs; rm -rf dist/; npm run build`).
+
+### Dataset
+
+The only thing missing is loading the data, we use our own data class that is a wrapper for the `tfjs` dataset class.
```
martinjaggi, line 44: make clear there is already a dataloader on the task. the node dataloader just extends that one. that's why i'm thinking both should life in the mytask.ts file, then everything would be in one place. the preprocessing is there as well (maybe add a comment there for both dataloaders too)

benchmark/README.md
```diff
@@ -31,8 +33,145 @@ npm run benchmark -- -u 4 -e 15 -r 5
...
+class NodeImageLoader extends dataset.ImageLoader<string> {
+  async readImageFrom(source: string): Promise<tf.Tensor3D> {
+    const imageBuffer = fs.readFileSync(source)
+    let tensor = tf.node.decodeImage(imageBuffer)
+    // <---- Add pre processing here!
+    // e.g: If resize needed uncomment the following
+    // tensor = tf.image.resizeBilinear(tensor, [
+    //   32, 32
+    // ])
+    tensor = tensor.div(tf.scalar(255))
```
martinjaggi, line 67: can't this be shared with the browser data-preprocessing somehow - or at least people should understand these two preprocessings should/must be the same. otherwise we can't claim it's the same task

benchmark/README.md
```diff
@@ -31,8 +33,145 @@ npm run benchmark -- -u 4 -e 15 -r 5
...
+  
+  ...
+
+  return await new NodeImageLoader(task).loadAll(files, {labels: labels})
+}
+
+
+```
+
+Example of `files` and `labels` content.
```
martinjaggi, line 84: very nice

martinjaggi (changes requested):

benchmark/README.md
Nacho114: yeah it could be, it's something we wanted to add with @s314cy, but I think we forgot about it since we didn't add it as an issue. @s314cy correct me if im wrong.
And indeed it would be not be identical with the browser.

benchmark/README.md
Nacho114: this is related to this > unclear what you here mean by access. >

benchmark/README.md
Nacho114: yes, but right now `preprocessFunctions` does not do anything, it's just a placeholder, so I wanted to make that clear.

benchmark/README.md
Nacho114: yeah i was thinking the same

Issue #363, opened 2022-07-07
Persistent models server-side


s314cy:
With the server now being compatible with both the browser and Node, there are issues with the TFJS backend which prevent the server from storing its trained models to disk, thus any server reboot/crash would mean a loss of the trained models. This also means the trained models are not reusable by the entity hosting the server other than by manually fetching the model over HTTP.

martinjaggi:
yes, but this is not a huge immediate problem as models are still properly saved on the clients, which is the main value for our users

PR #391, opened 2022-07-20
Modules import TFJS from Disco.js and use a unique TFJS backend

### LINKED ISSUE #371: load tfjs backend once
in the browser & server, importing discojs causes tfjs to be loaded twice and causes issues such as #363

in general, it is very error prone to load multiple tfjs backends (e.g. webGL + CPU) as it can mess with the memory locations of tensors and cause hard-to-fix errors

### LINKED ISSUE #379: "Variable with name conv2d_Cov2D1/kernel was already registered" error
Error when "train locally" selected after "train collaboratively" 
Note: train collaboratively did not work and thus may be dependent on issue #378

Screenshot 2022-07-14 at 00 54 55

s314cy:
closes #371 and closes #379, enables #363 for later

all modules (server, browser & benchmark) import their tfjs from discojs, and the tfjs backend is set to the corresponding module's environment (either browser or node)

this ensures each module shares the same backend as their imported instance of discojs

note that discojs now imports tfjs only, i.e. without the c++ bindings, which means using discojs on node with greater performance requires to manually import the tfjs-node backend (for lack of a better workaround)

### DIFF
--- a/benchmark/src/benchmark.ts
+++ b/benchmark/src/benchmark.ts
@@ -1,11 +1,11 @@
 import {Range} from 'immutable'
 import {Server} from 'node:http'
-import {client, ConsoleLogger, training, TrainingSchemes, EmptyMemory, informant, TrainerLog} from 'discojs'
+import {tf, client, ConsoleLogger, training, TrainingSchemes, EmptyMemory, informant, TrainerLog} from 'discojs'
+import '@tensorflow/tfjs-node'
 
 import {startServer, getClient, saveLog} from './utils'
 import {getTaskData} from './data'
 import {args} from './args'
-
 
 const NUMBER_OF_USERS = args.numberOfUsers
 let TASK = args.task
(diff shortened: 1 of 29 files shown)


tharvik:
Hum, now that I ran the tests on my side, two things:
  * first, we forgot to run the lib's test in the CI
  * second, some of its tests requires node to read files (thus are failing to compile)

Simply importing `@tensorflow/tfjs-node` doesn't help :/

s314cy:
> Hum, now that I ran the tests on my side, two things:
> 
>     * first, we forgot to run the lib's test in the CI
> 
>     * second, some of its tests requires node to read files (thus are failing to compile)
> 
> 
> Simply importing `@tensorflow/tfjs-node` doesn't help :/

I guess an easy fix for (2) would be to
```
import * as tf from '@tensorflow/tfjs-node
```
rather than
```
import tf from '@/.'
```
in tests of discojs

s314cy:
> Hum, now that I ran the tests on my side, two things:
> 
>     * first, we forgot to run the lib's test in the CI
> 
>     * second, some of its tests requires node to read files (thus are failing to compile)
> 
> 
> Simply importing `@tensorflow/tfjs-node` doesn't help :/

opened #401 for (1), guess we can merge this PR?

PR #399, opened 2022-07-22
Task filters as checkboxes and new filters

### LINKED ISSUE #395: Tasklist FILTERS (ambiguous)
Currently in the [TaskList](https://epfml.github.io/disco/#/list), `FILTERS` appears as the main heading on top. 

- This is ambiguous and could make it look like the whole page contains filters (the filters only pertain to the `Decentralized` and `Federated` buttons)
- Further, the `Decentralized` and `Federated` buttons look just like ones that usually navigate to other pages...and it is assumed it will navigate away from the page.

Could we rather 
1) Put the buttons and the `FILTERS` heading in a dedicated "Filter box" off-center
2) make the `FILTERS` text smaller
3) make the decentralized and federated options smaller...and preferably make them subtle check boxes to make the toggling more intuitive and to make them look different from navigation commands (like below)
Screenshot 2022-07-21 at 17 50 41

https://github.com/epfml/disco/blob/develop/browser/src/components/pages/TaskList.vue

s314cy:
closes #395
branched off #391 

@iglobalhealth does it correspond to what you pictured? would you prefer the filters above or below "discollaboratives"?

Screenshot 2022-07-22 at 13 53 16

### DIFF
--- a/benchmark/src/benchmark.ts
+++ b/benchmark/src/benchmark.ts
@@ -1,11 +1,11 @@
 import {Range} from 'immutable'
 import {Server} from 'node:http'
-import {client, ConsoleLogger, training, TrainingSchemes, EmptyMemory, informant, TrainerLog} from 'discojs'
+import {tf, client, ConsoleLogger, training, TrainingSchemes, EmptyMemory, informant, TrainerLog} from 'discojs'
+import '@tensorflow/tfjs-node'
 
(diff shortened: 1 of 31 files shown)


Annie-LiGHT:
Ohhh that is beautiful! 💅 
I think it would look better below the `discollaboratives` heading.

Also...could you change the text in each to be preceded by "Filter by"?

PR #404, opened 2022-07-22
Add test-lib to CI

### LINKED ISSUE #401: CI does not run tests for discojs
the current CI does not run the lib's tests

s314cy:
closes #401


tharvik (approved):
There is the same `with_server` script in `browser`, consider using it direclty (or symlink it) to avoid code driffting.
Otherwise, LGTM

PR #407, opened 2022-07-26
Secure Aggregation: simplified

tharvik:
Smaller and based on decentralized clients fixes of #339


tharvik:
After discussion with @martinjaggi: first, we merge #339, then we can apply some of the changes of this PR to `develop`.

PR #432, opened 2022-08-03
Adding mobilenetv2 for simple_face

Nacho114:
Added mobilenetv2 for simple face (currently being tested)

Trying out on Simple face (partial data set of ~ 1000 images in a 2 class setting), first in the single user setting.

-> On the the replica [model](https://towardsdatascience.com/building-mobilenet-from-scratch-using-tensorflow-ad009c5dd42c), accuracy was quite bad (Perhaps it has a mistake w.r.t. actual mobilenetv2).

Finally managed to make loadLayersModel work (however only for nodejs).


Nacho114:
Fixed in another PR

PR #443, opened 2022-08-05
Update TASK.md with descriptions of all TaskInformation parameters

aunell:
We want to clearly describe each parameter in task information so that users who do not have extensive coding knowledge can understand where they can personalize training in DisCo and the significance of the different parameters. We can describe the different parameters in [TASK.md](https://github.com/epfml/disco/blob/develop/information/TASK.md), and they will be able to specify values outside of default values as they deem fit.


Nacho114:
Felix has done this in #425

martinjaggi:
looks good, but some might still need a comment, such as `aggregateImagesById`,
`csvLabels` and `LABEL_ASSIGNMENT`

martinjaggi:
addressed in #459

PR #451, opened 2022-08-17
Less padding on mobile, Uniform padding throughout the app, Remove Lodash

### LINKED ISSUE #259: Replace lodash with immutable map functionality
@tharvik introduced [immutable](https://immutable-js.com/) that adds immutable collections to js, it also adds maps, filters, ... these are more comprehensible than the current maps we use via `lodash`. 

- Replace the remaining `lodash` usage from the code with the corresponding `immutable` and then remove `lodash` from the packages.

s314cy:
- Less padding, smaller containers and other fixes, to avoid glitchy displays on mobile
- Padding is more uniform throughout the app
- Remove the Lodash package 🎉  (closes #259)


Nacho114 (approved):
lgtm

PR #452, opened 2022-08-17
Add pre processing functionality that is easy to extend

### LINKED ISSUE #385: Preprocessing
In the task.trainingInformation object there is a `preprocessFunctions` variable, it is however not yet used.

It would be good to add a simple class that contains some basic preprocessing (e.g. normalizing, re-scaling, ...), then based on the preprocessFunctions chosen, these are applied in the data loader.

### LINKED ISSUE #393: DataLoader File Consolidation
Currently, dataloaders are created for different tasks in different spaces in the code. We need to consolidate this so that each task has its own dataloader.

Nacho114:
We add simple pre processing functionality, this reduces code duplication and it is also easy to add custom tasks.

TODO:

- [x] Update Task.md

Fixes #385 
Fixes #393 -> We consolidate all the preprocessing in a script


discojs/src/dataset/data_loader/image_loader.ts
```diff
@@ -12,15 +14,33 @@ import { DataLoader, DataConfig, Data, DataTuple } from './data_loader'
  */
 
 export abstract class ImageLoader<Source> extends DataLoader<Source> {
+  private readonly preprocessImage?: PreprocessImage = undefined
+
+  constructor (task: Task) {
+    super(task)
+    if (this.task.trainingInformation !== undefined &&
```
s314cy, line 21: `if (this.task.trainingInformation?.preprocess?.length > 0)`

discojs/src/dataset/preprocess.ts
```diff
@@ -0,0 +1,26 @@
+import * as tf from '@tensorflow/tfjs'
+import { TrainingInformation } from '../task/training_information'
+
+export type PreprocessImage = (image: tf.Tensor3D) => tf.Tensor3D
+
+export enum Preprocess {
```
martinjaggi: Preprocessing

discojs/src/dataset/preprocess.ts
```diff
@@ -0,0 +1,26 @@
...
+  Normalize = 'normalize',
```
martinjaggi: best to call them `image.normalize` or sth like this, as there might be a ton of functions in the future? btw there is still this `threshold` param on the task def which isn't saying what it does, i forgot what it's used for

discojs/src/dataset/preprocess.ts
```diff
@@ -0,0 +1,26 @@
...
+    if (info.preprocessFunctions.includes(Preprocess.Resize) &&
+      info.RESIZED_IMAGE_H !== undefined &&
+      info.RESIZED_IMAGE_W !== undefined) {
+      image = tf.image.resizeBilinear(image, [
+        info.RESIZED_IMAGE_H, info.RESIZED_IMAGE_W
+      ])
+    }
+    return image
+  }
+  return preprocessImage
```
martinjaggi: preprocessedImage (as it's now done)

discojs/src/task/training_information.ts
```diff
@@ -1,13 +1,14 @@
 import { DataExample } from './data_example'
 import { ModelCompileData } from './model_compile_data'
+import { Preprocess } from '../dataset/preprocess'
```
martinjaggi: Preprocessing

discojs/src/tasks/cifar10.ts
```diff
@@ -32,7 +32,7 @@ export const task: Task = {
     csvLabels: true,
     IMAGE_H: 32,
     IMAGE_W: 32,
-    preprocessFunctions: ['resize'],
+    preprocessFunctions: [],
```
martinjaggi: preprocessingF...

information/TASK.md
```diff
@@ -57,6 +52,99 @@ export function model (): tf.LayersModel {
   return model
 ```
 
+#### Preprocessing
+
+In the Task object we can optionally choose to add preprocessing functions. Preprocessing is defined [here](../discojs/src/dataset/preprocess.ts),
+and is currently only implemented for images. The predefined functions (resize, normalize, ...) can be found there, and it is also there
```
martinjaggi: first keep the explanation data modality independent. what to do if i want to add a new function. how do i add dataloader and preproc code, where, and which which input/output signatures? (batchsize is also a dimension or not?)

information/TASK.md
```diff
@@ -57,6 +52,99 @@ export function model (): tf.LayersModel {
...
+where you can easily create custom preprocessing.
+
+> Preprocessing is done every time an image is read, this will increase the overhead of training since image loading is lazy; for better performance use data that is already preprocessed.
```
martinjaggi: i thought it should be batch-wise, for all modalities?

information/TASK.md
```diff
@@ -57,6 +52,99 @@ export function model (): tf.LayersModel {
...
+
+Suppose we want our custom preprocessing that divides our pixels by 2. In the [preprocessing](../discojs/src/dataset/preprocess.ts) file,
```
martinjaggi: each pixel value

information/TASK.md
```diff
@@ -57,6 +52,99 @@ export function model (): tf.LayersModel {
...
+    ...
+    preprocessFunctions: [Preprocess.Custom],
+    ...
+    ...
+  }
+}
+```
+
+### Model
+
```
martinjaggi: maybe we put the model first and then preprocessing, as it will define what format we need to feed into first layer? still unclear to me here

information/TASK.md
```diff
@@ -57,6 +52,99 @@ export function model (): tf.LayersModel {
...
+  const model = tf.sequential()
+
+  // Add layers
+  model.add(...)
+
+  return model
+
+```
+
+Alternatively we can also load a pre-existing model:
```
martinjaggi: do you mean just architecture, or arch with pretrained weights? best if we think from user perspective, they might have a model with or without weights, how do they need to convert it to make it work here

information/TASK.md
```diff
@@ -57,6 +52,99 @@ export function model (): tf.LayersModel {
...
+export async function model (): Promise<tf.LayersModel> {
+  return await tf.loadLayersModel('file:./modelPath/model.json')
+}
+```
+
+the `modelPath` is relative to the root directory of the `server`.
+
+### Making the task visible to the API
+
 After adding the `simple_face.ts` task class, we also need to export it in the `index.ts` file which lives in the same folder.
```
martinjaggi: maybe we can rename it to `my_new_task.ts` or sth similar, as `simple_face` already exists so might be confusing? or do you mean to edit an existing task or duplicate to a fresh one?

martinjaggi (changes requested):

discojs/src/dataset/preprocess.ts
Nacho114: It's not used anywhere, I think this was part of sync Federated, where we had the threshold of how many fractions of users to choose from for training. I will delete it since it's deprecated.

information/TASK.md
Nacho114: I took the pre-existing code for pre processing which was image wise; I'll have a look at how batch wise would look like.

discojs/src/dataset/preprocess.ts
Nacho114: preprocessImage is a function

information/TASK.md
Nacho114: Modified it, it is now batch wise

PR #454, opened 2022-08-17
UI testing with Cypress

### LINKED ISSUE #278: Simple UI tests
One of the recent PRs passed all tests, however it broke the UI and a very simple UI test would have prevented this. It would be very helpful to have simple UI tests to make sure the page does not break with changes. 

As recommend by @s314cy, [cypress](https://www.cypress.io/) seems to be a solid library for this. Once the test is made it will be easy to add to the current CI/CD test suite of GitHub actions.

s314cy:
closes #278


browser/cypress/fixtures/example.json
```diff
@@ -0,0 +1,5 @@
+{
```
Nacho114, line 1: Maybe delete this example file? Ditto for commands.ts?

Nacho114 (approved):
lgtm

PR #456, opened 2022-08-24
Fix for loading pretrained models & storing models server side

### LINKED ISSUE #363: Persistent models server-side

### LINKED ISSUE #446: Loading pre-trained models
Due to issues with the TFJS back end, both loading and saving (#363) models from/to disk is currently not possible in discojs. In particular this means we cannot use pre-trained models.

Currently models are [built](https://github.com/epfml/disco/blob/develop/discojs/src/tasks/lus_covid.ts) and then said models are made available in the server. 

If we have a pre trained model stored as a model.json + weights.bin, we would load it as follows:

```js
await tf.loadLayersModel('file:./modelPath/model.json')
```

This however yields the following error:

```
[Tensorflow Node-js, TypeError: Only HTTP(S) protocols are supported](https://stackoverflow.com/questions/60137483/tensorflow-node-js-typeerror-only-https-protocols-are-supported)
```

As with #363, using both `tfjs` and `tfjs-node` seems to cause unexpected behavior as reported [here](https://stackoverflow.com/questions/60137483/tensorflow-node-js-typeerror-only-https-protocols-are-supported)

@tharvik Suggested to import either tfjs or tfjs-node on discojs depending on if it is used by browser or node, since the api is identical this should work in principle. Perhaps a simpler solution is possible.

Another idea ( @s314cy ) is to refactor the pertinent code from discojs back into the server.

Nacho114:
Fix for loading pretrained models by conditional import of tfjs-node

When importing `tfjs-node` two things happen:
1. The appropriate tfjs engine is loaded
2. The appropriate functions are loaded

It seems that if we import `tfjs-node` the only 1. happens, since when loading and saving files the API still thinks it's tfjs (no node version); if we import 'tfjs-node' conditionally earlier then 2. seems to happen properly allowing us to save and load models again.

Note that due to some webpack dependency checks, we need to have tfjs-node inside a string to take advantage of lazy loading to avoid weback issues.

TODO: 

- [x] model.json was not added, check on this
- [x] check if it is possible to save models in the server
- [x] Added timeout for test, was running in infinite loop if server crashed right after build.
- [x] find better way for model path (current one is relative, might not be ideal since it assumes CLI and server are at a same distance from model path).
- [x] update TASK.md

Fixes #446 
Also Fixes #363


server/src/get_server.ts
```diff
@@ -6,6 +6,10 @@ import { CONFIG } from './config'
 import { Router } from './router'
 import { TasksAndModels } from './tasks'
 
+// env variable used by discojs in order to be able to load
+// models through the file system.
+process.env.TFJSNODE = '@tensorflow/tfjs-node'
```
s314cy: why do we need to setup an env var here? seems like a lot of trouble for something that could be a simple var :p

discojs/src/tasks/simple_face.ts
```diff
@@ -39,48 +39,12 @@ export const task: Task = {
...
-  // Our last layer is a dense layer which has 10 output units, one for each
-  // output class (i.e. 0, 1, 2, 3, 4, 5, 6, 7, 8, 9).
-  model.add(tf.layers.dense({
-    units: numOutputClasses,
-    kernelInitializer: 'varianceScaling',
-    activation: 'softmax'
-  }))
+export async function model (): Promise<tf.LayersModel> {
+  // TFJSNODE is define in the server, so will only be imported there
+  if (typeof process.env.TFJSNODE === 'string') {
```
s314cy: nice 🚀 but the function would cause trouble if called within another env than node, i.e. somewhere the `process` wouldn't be defined

s314cy (changes requested):

discojs/src/tasks/simple_face.ts
Nacho114: changed it to check if window is define (since it is the only case we are outside of nodejs)

server/src/get_server.ts
Nacho114: nice

Nacho114:
Currently loading the model looks as follows:

```js
export async function model (): Promise<tf.LayersModel> {
  const file = 'file://./../models/mobileNetV2_35_alpha_2_classes/model.json'
  return await tf.loadLayersModel(file)
}
```

the issue is that `file` is relative to where the code is executed from. Note that it is either run via the server, or the benchmark. In both cases the model is at `../` distance from the cwd. So it is not very robust. One solution is to store the path in process.env and define it once in benchmark and once in server, this way the stored models path is relative not to cwd, but to each env.

Nacho114:
@tharvik @s314cy wdyt?

Nacho114:
> @tharvik @s314cy wdyt?

Found a clean solution by following the server config as suggested by @tharvik

s314cy (approved):
image

PR #463, opened 2022-09-14
Code refactoring as base for federated WS implementation

### LINKED ISSUE #226: Server communication HTTP vs Web sockets
Server communication HTTP vs Web sockets

Should we keep HTTP or move to Web sockets?

The benefit of Web sockets is that we can easily push data from server to client, e.g. "model update" or even for hyper parameter tuning #151. These features are more complex with HTTP.

The problems with web sockets are mainly the overhead it adds, although this might be negligible?

Work wise it is worth it given the benefits that we mentioned.  

Some reading:

https://blogs.windows.com/windowsdeveloper/2016/03/14/when-to-use-a-http-call-instead-of-a-websocket-or-http-2-0/

francesco98:
### Library (discojs)
Refactoring of `Federated` class. New `federated` folder created to contain a `client.ts` file that replaces the old `federated.ts` file. It will be useful to contain other files that are related to federated ws implementation.

### Server-side
Refactoring of `Decentralized` class. New parent class `Server` is now implemented and `Decentralized` extends it. The `SignalingServer` class is now removed because the WS requests are now handled directly by `Decentralized` class. It will be useful for federated ws implementation (i.e.: `Federated` class will have the same structure once implemented).

--> Also, imports of Federated class from library have been updated to comply the new class structure.

### Benchmark
The `benchmark.ts` entry file has been updated with a new line `await cli.disconnect()`. Indeed, it's important to disconnect the client once the training process is finished.

--> Also, imports of Federated class from library have been updated to comply the new class structure.

### Client-side (browser)
Imports of Federated class from library have been updated to comply the new class structure.


martinjaggi (approved):

s314cy:
excellent work :)

PR #464, opened 2022-09-14
Server and client WS Federated implementation

### LINKED ISSUE #226: Server communication HTTP vs Web sockets

francesco98:
### Server-side
`Federated` server class now implements a server-side WS, and all old HTTP endpoints have been converted to WS listener messages.

**NOTE**: `nonValidTask` test has been disabled because this functionality is not implemented yet for WS.

### Client-side
`Federated` client class now implements a client-side WS, and all old HTTP requests have been converted to WS messages.
A new `messages.ts` has been created to represent the structure of all exchanged messages over the WS channel.

### DIFF
--- a/discojs/src/client/federated/client.ts
+++ b/discojs/src/client/federated/client.ts
@@ -1,9 +1,15 @@
 import * as msgpack from 'msgpack-lite'
-import axios from 'axios'
+import isomorphic from 'isomorphic-ws'
 import { v4 as randomUUID } from 'uuid'
 
-import { privacy, serialization, informant, MetadataID, Weights } from '../..'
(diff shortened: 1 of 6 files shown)


s314cy (approved):
excellent work :)

Issue #345, opened 2022-06-24
text, SVG and GIF transparency


s314cy:
currently, common browser extensions bringing dark themes to websites work particularly well on disco

yet, a few cosmetic changes remain to be done:
- white text/SVGs should instead be transparent (`text-transparent` tailwind class)
- all three GIFs (disco logo and decentralized/federated diagrams) should have a transparent background (no idea of how this works for GIFs)
  - right now, the background color is hardcoded to match the current background of the containing page
- the MLO png logo should have an alpha channel of 0 for its white background

hleepfl:
outdated, site looks okay with dark mode addons and it has its own dark mode

PR #470, opened 2022-09-20
Changed image assets with transparent background

batuhanfaik:
Resolves two bullet points in issue #345


martinjaggi (approved):
thanks so much!

PR #471, opened 2022-09-21
Implementing byzantine-tolerant aggregator formula

francesco98:
### Library (discojs)

New byzantine-tolerant aggregator added. A new utility file has been created to contain all generic operations (norm, quantile, etc).
New parameters in the `TrainingInformation` interfaces added to include the byzantine-tolerant aggregator in the aggregation phase and to specify the tau percentile to use while applying the formula. 

### Server

Changes in Federated class to include byzantine-tolerant aggregator, if the model wants to use it (looking at TrainingInformation parameters)

### DIFF
--- a/discojs/src/aggregation.ts
+++ b/discojs/src/aggregation.ts
@@ -34,11 +34,17 @@ export function averageWeights (peersWeights: List<Weights>): Weights {
   return sumWeights(peersWeights).map((w) => w.div(numberOfPeers))
 }
 
+// Implementation of: https://arxiv.org/abs/2012.10333
 export function averageCenteredClipping (peersWeights: List<Weights>, previousWeights: Weights, tau: number): Weights {
-  const subtractedWeights: List<List<tf.Tensor>> = peersWeights.map(weights => List(weights).zip(List(previousWeights)).map(([w1, w2]) => w1.sub(w2)))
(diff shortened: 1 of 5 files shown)


martinjaggi (commented):
looks good but a bit hard to read. can you use the above functions to simplify / shorten, and add some comments (on what is the center, what will be clipped)? 
maybe the code for this itself could be in the same file as the clipping for DP, for easier comparison? as you prefer

francesco98:
> looks good but a bit hard to read. can you use the above functions to simplify / shorten, and add some comments (on what is the center, what will be clipped)? maybe the code for this itself could be in the same file as the clipping for DP, for easier comparison? as you prefer

Okay, I agree in simplifying it, but I would not put it in the same file as DP because that file is called `privacy.ts` indicating the set of functions used to add privacy. In this case we are making a new aggregator, so I think it's better to have it in the `aggregator.ts` file. What do you think?

Also, I don't think we can use the previous functions to simply the computation because the `applyArithmeticToWeights` method applies a function to a set of weights. Instead, we are using both current and previous weights for the new aggregator. However, for the final aggregation I'm still using the `averageWeights` method.

francesco98:
This PR has been replaced by: https://github.com/epfml/disco/pull/489

PR #472, opened 2022-09-27
Changing benchmark module to expect a server instance running

francesco98:
### Benchmark module

It has been modified to expect a server instance running instead of being standalone. To run `benchmark` now, you need to have a server instance running on `localhost:8080`

### Discojs library

It now includes `exampleData.ts` that has been moved from benchmark module to overcome the issue with `tfjs-node` dependency on M1 Mac.

### DIFF
--- a/.github/workflows/lint-test-build.yml
+++ b/.github/workflows/lint-test-build.yml
@@ -216,5 +216,4 @@ jobs:
       - run: npm ci 
         working-directory: ${{ env.server_dir }}
       - run: npm ci
-      - run: ./with_server npm run dev
-      - run: npm run benchmark -- -t cifar10 -u 1 -e 1
+      - run: ./with_server npm run benchmark -- -t cifar10 -u 1 -e 1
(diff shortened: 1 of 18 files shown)


s314cy:
maybe this can be dropped given #474? :)

francesco98:
> maybe this can be dropped given #474? :)

we could drop it, but maybe it still make sense that benchmark expects a server running, instead of launching its own instance. What do you think?

s314cy:
> > maybe this can be dropped given #474? :)
> 
> we could drop it, but maybe it still make sense that benchmark expects a server running, instead of launching its own instance. What do you think?

let's go with the best of the two worlds:
* launch a server on 8080 by default
* if a port is specified, expect a server running there

francesco98:
> > > maybe this can be dropped given #474? :)
> > 
> > 
> > we could drop it, but maybe it still make sense that benchmark expects a server running, instead of launching its own instance. What do you think?
> 
> let's go with the best of the two worlds:
> 
> * launch a server on 8080 by default
> * if a port is specified, expect a server running there

Okay it's not the priority now, so I close it for the moment

PR #473, opened 2022-09-27
Add weights class

### LINKED ISSUE #325: Create a `weights_math` module to collect the Weights-related functionality that is scattered across the project
## Situation:
`Weights` is a simple typedef. Since `Weights` is an array of `tf.Tensor`, even things as simple as adding together two compatible `Weights` objects are not trivial. So far, everyone has implemented whatever functionality they needed right where they needed it, leading to code duplication and time wasted as the same operation is implemented in several places and devs don't know where to search for existing implementations of the functionality they need.

## Updated proposal (cf. comments):
Write a file `weights_math.ts` instead of a `Weights` class.

## Proposed solution:
Turn `Weights` into a proper class, defined in the new file `discojs/src/weights.ts`. Go through the files listed below (and possibly others) to collect all the existing functionality and refactor it into methods of the `Weights` class.

### Where to find existing implementations of functionalities for the `Weights` type:
- `discojs/src/serialization.ts`
- `discojs/src/aggregation.ts`
- `discojs/src/secret_shares.ts`

s314cy:
closes #325

- add a convenient and friendly `WeightsContainer` class wrapping the `Weights` type and enhancing it with common operations: add, sub, etc.
- add aggregation functions (sum, diff, avg, etc.) working in synergy with the new `WeightsContainer` class
- the regular `Weights` are accessed via the `weights` getter
- propagate the changes throughout the codebase


discojs/src/client/decentralized/base.ts
```diff
@@ -227,7 +227,7 @@ export abstract class Base extends ClientBase {
       const finalWeights = await this.sendAndReceiveWeights(peers, noisyWeights, round, trainingInformant)
 
       console.debug(this.ID, 'sent and received', finalWeights.size, 'weights at round', round)
-      return aggregation.averageWeights(finalWeights)
+      return aggregation.avg(finalWeights).weights
```
morganridel: So if we have this `WeightContainer` which abstracts from the TF `Weights` (basically `tf.Tensor[]` alias if I understood correctly), shouldn't we keep it encapsulated inside the container (keep the abstraction) until we have some TF specific logic?

So for example `onRoundEndCommunication` returns `Promise<WeightsContainer>` and we keep the `WeightsContainer` up in the chain until we really needs the underlying object, in this case when the trainer needs to update the model `this.model.setWeights(aggregatedWeightsContainer.weights)`.

I guess it could be cool to have a really independant `client` folder which never reference the `tf` module, because it really only cares about communication and aggregation. After looking through the code though it seems that `Weight` is still used a lot so maybe not adequate right now.

discojs/src/weights/aggregation.ts
```diff
@@ -0,0 +1,58 @@
...
+): WeightsContainer {
+  return parseWeights(weights).reduce((acc: WeightsContainer, ws: WeightsContainer) =>
+    new WeightsContainer(acc.weights.map((w, i) =>
+      fn(w, ws.get(i) as tf.Tensor))))
+}
+
+export function sum (weights: Iterable<WeightsLike | WeightsContainer>): WeightsContainer {
+  return reduce(weights, tf.add)
+}
+
```
morganridel, line 36: It looks like most of the time, these functions are used with `Weights` as input to output a `WeightsContainer` which is immediately dereferenced again i.e.:

```
aggregation.avg(weights).weights
```

I wonder what is the expected "role" of `WeightsContainer` in the code, it looks more like some utils function because we never really keep it contained

discojs/src/client/decentralized/base.ts
morganridel: So now the aggregation functions output WeightsContainer instead of Weights but we don't use the container after?

morganridel (commented):
I like the idea of a dedicated class to handle some logic but I'm not sure I see the benefits in the current code which still use `Weights` everywhere.

My comments are more like some code design/architecture topics that I like to think about. I'm a newbie on the project and I'm not aware of all the considerations! Feel free to disregard them if it's not pertinent yet

discojs/src/client/decentralized/secret_shares.ts
```diff
@@ -3,7 +3,7 @@ import { List, Range } from 'immutable'
 import * as tf from '@tensorflow/tfjs'
 import * as crypto from 'crypto'
 
-import { Weights, aggregation } from '../..'
+import { Weights, WeightsContainer, aggregation } from '../..'
```
morganridel: So in the same idea, why have the `WeightContainer` added if we still need the `Weights`? this can become confusing to understand (*Do I need a `WeightContainer` or a `Weights`*?) depending on the scope/responsability of the current logic

martinjaggi:
thanks for the comments!
so maybe it could be simpler to consistently use `TF Weights` everywhere (but not list of TF tensor as that can be confusing with nested lists). probably not worth abstracting that yet as no alternative in sight yet... we could keep a stateless bunch of methods as in `aggregation.ts` to provide the arithmetics needed (sum of 2 or more, diff of 2, approxEqual).

s314cy:
Thanks for the most relevant comments! I see two paths here:
1. Go further and put `WeightsContainer` everywhere, only unwrapping its content when directly working with TF.js
2. Keep `Weights` everywhere and work with them with (stateless) functions

I have a personal preference for 1. (Morgan's proposal), since it will leverage `WeightContainer` to the fullest. Just need your approval @martinjaggi :)

discojs/src/client/base.ts
```diff
@@ -54,9 +51,9 @@ export abstract class Base {
...
    */
   abstract onRoundEndCommunication (
-    updatedWeights: Weights,
-    staleWeights: Weights,
+    updatedWeights: WeightsContainer,
+    staleWeights: WeightsContainer,
     round: number,
     trainingInformant: TrainingInformant
-  ): Promise<Weights> // changed from weights to void because we don't want weights autosent at end of round
+  ): Promise<WeightsContainer> // changed from weights to void because we don't want weights autosent at end of round
```
martinjaggi: comment is outdated maybe?

discojs/src/client/decentralized/secret_shares.ts
```diff
@@ -3,23 +3,23 @@ import { List, Range } from 'immutable'
...
 /*
 Return Weights in the remaining share once N-1 shares have been constructed (where N is number of ready clients)
  */
-function lastShare (currentShares: List<Weights>, secret: Weights): Weights {
+function lastShare (currentShares: List<WeightsContainer>, secret: WeightsContainer): WeightsContainer {
   if (currentShares.size === 0) {
     throw new Error('Need at least one current share to be able to subtract secret from')
   }
-  return aggregation.subtractWeights(List([secret, aggregation.sumWeights(currentShares)]))
+  return secret.sub(aggregation.sum(currentShares))
```
martinjaggi, line 16: maybe `subtract` to be extra clear?

discojs/src/privacy.ts
```diff
@@ -12,18 +9,19 @@ import { tf } from '.'
...
-    .zip(List(staleWeights))
-    .map(([w1, w2]) => w1.sub(w2))
-  let newWeightsDiff: List<tf.Tensor>
+  const weightsDiff = updatedWeights.sub(staleWeights)
+  let newWeightsDiff: WeightsContainer
 
   if (clippingRadius !== undefined) {
     // Frobenius norm
-    const norm = Math.sqrt(weightsDiff.map((w) => w.square().sum().dataSync()[0]).reduce((a: number, b) => a + b))
+    const norm = Math.sqrt(
```
martinjaggi, line 21: don't we have a difference and a norm function in the weights helper file?

martinjaggi: or does this here avoid one copy or so? (weights are large so should avoid memory duplications etc)

discojs/src/training/trainer/distributed_trainer.ts
```diff
@@ -25,8 +25,8 @@ export class DistributedTrainer extends Trainer {
    * Callback called every time a round is over
    */
   async onRoundEnd (accuracy: number): Promise<void> {
-    const currentRoundWeights = this.model.weights.map((w) => w.read())
-    const previousRoundWeights = this.previousRoundModel.weights.map((w) => w.read())
+    const currentRoundWeights = WeightsContainer.from(this.model)
```
martinjaggi, line 28: a bit confused what type `this.model` is, why does it need to be converted?

server/src/router/federated.ts
```diff
@@ -190,7 +190,7 @@ export class Federated extends Server {
 
         this.logsAppend(task.taskID, clientId, RequestType.GetAsyncRound, 0)
 
-        const weights = model.weights.map((e) => e.read())
+        const weights = WeightsContainer.from(model)
```
martinjaggi, line 193: same, why conversion?

server/tests/end_to_end_decentralized.ts
```diff
@@ -82,18 +80,18 @@ describe('end to end decentralized', function () {
     with other ready peers. The input will vary with model architecture and training data. If secure is true,
     the client will implement secure aggregation. If it is false, it will be a clear text client.
      */
-  async function makeClient (input: number[][], secure: boolean): Promise<Weights> {
+  async function makeClient (input: number[], secure: boolean): Promise<WeightsContainer> {
```
martinjaggi, line 83: (this is from older code, not related to current PR much. why is input a number type and not a model already?)

martinjaggi (changes requested):

discojs/src/training/trainer/distributed_trainer.ts
s314cy: the model held by the trainer is a raw TF.js model used for training :) this object holds all the model's info, in particular the weights which we extract with `WeightsContainer.from`

server/tests/end_to_end_decentralized.ts
s314cy: This file is a part of tests. The input here was a list of weights simplified as `number[][]` (<-> `Weights[]` <-> `tf.Tensor[][]`) since we didn't have any utilities to get a `Weights` object from numbers easily, which is convenient for tests. Now, `WeightsContainer` accepts raw numbers as input so we can directly pass those.

I left the `input[]` in `makeClient` to avoid repeatedly creating `WeightsContainer`s by leaving it to the function.

discojs/src/privacy.ts
s314cy: We don't have have any utility for the Frobenius norm, so I compute it manually from the weights container's tensors.  We already have an `add` method for adding weights containers together entry-wise, but the `reduce` here adds all entries of a single weights container together.

I moved `dataSync()` outside of the first map to avoid unnecessarily transferring tensors from GPU to CPU for each entry of the weights matrix. Now, `dataSync` is called after the reduction to a single scalar, which is much more efficient.

discojs/src/client/decentralized/secret_shares.ts
s314cy: I like short names as long as they're readable-enough, especially for such primitive operations — just like TF.js proposes `add`, `sub`, `div`, `mul`, `mod` and `pow` operations on its tensors. This is highly subjective, so I can fall back to more explicit names if you prefer :)

martinjaggi: you're right, it's better to be consistent with these TF ones

PR #474, opened 2022-10-03
Split Disco into 3 modules (core, js, node)

s314cy:
Split the DiscoJS module into:
- discojs-core, which replaces the current discojs and compiles into two distinct dists
- discojs, using the browser dist (imports `@tensorflow/tfjs`) and used by the browser module
- discojs-node, using the Node dist (imports `@tensorflow/tfjs-node`) and used by the server and benchmark modules

Users and devs should only be interacting with and using the @epfml/discojs and @epfml/discojs-node packages

This clean split fixes the backend/cross-platform problems we are encountering with TF.js


martinjaggi (commented):
looks very clean, thanks!

small thing: example training data maybe is duplicated in `-core` and `-node`?

PR #476, opened 2022-10-10
Update docs following lib split

s314cy:
update of several READMEs to reflect the architecture changes brought to Disco.js


morganridel (approved):
LGTM

PR #483, opened 2022-10-11
Ready for v2.0.0

### LINKED ISSUE #482: Disco.js v2.0.0
According to [semver](https://semver.org)'s guidelines, it seems like Disco.js is subject to a v2 release 🎉 

Version content:
<ol>
<li><code>discojs-core</code>: weights communication in the decentralized setting is fully p2p</li>
<li><code>browser</code>: fixed the task creation form</li>
<li>split <code>discojs</code> into three distinct modules:</li>
 <ul>
<li><code>discojs-core</code>: core module containing platform-agnostic code</li>
 <li><code>discojs-node</code>: Node.js module based off and extending <code>discojs-core</code></li>
 <li><code>discojs</code>: browser (native JS) module based off and extending <code>discojs-core</code></li>
</ul>
</ol>
Given the library changes mentioned in (3), three packages would be published: <code>@epfml/discojs</code>,<code>@epfml/discojs-node</code> and <code>@epfml/discojs-core</code>. The latter would be published mainly as a dependency to the other two.

### LINKED ISSUE #487: Update README guides
This readme is out of date: https://github.com/epfml/disco/blob/develop/information/TASK.md as noted by Dennis Yurkevich

s314cy:
Fixes both #482 and #487

## List of changes

- discojs, discojs-node and discojs-core exist in a bigger discojs dir ("outer" discojs), lying next to browser, server and benchmark to clearly isolate the four projects of the repo
- discojs-core is just an internal project containin code shared across discojs packages, it is thus not published to npm
- The special trick here is that discojs-core is symlinked to both `discojs/src/core/` and `discojs-node/src/core/`
- The `dist/` directories of discojs and discojs-node are published to npm together with the browser- and Node-compiled versions of discojs-core, respectively
- Local changes to discojs-core would directly reflect in discojs and discojs-node after build, as one would expect when contributing to the codebase (one less compilation step!). This is thanks to discojs-core actually being part of the discojs and discojs-node projects via the symlink.
- To ease development, global scripts are available in the outer discojs dir: `npm run build` and `npm run test`, building all the discojs libs at once and testing them, respectively.
- The `package.json` in the outer discojs is here for the discojs-core symlink to see installed `node_modules` when developing/testing, AND for the discojs and discojs-node libs to share their installed `node_modules` (all three modules got the same deps). This has the advantage to reduce download time and memory usage, while allowing for our fancy project structure (i.e. the symlink).
- discojs and discojs-node have no `node_modules` since they can see the one in the parent dir, and their `package.json` has no dependency: they are only added on publishing to the npm registry :)
- The package currently published to npm presents issues as to its usage, because of some confusion between dependencies and dev dependencies; this has been fixed.

## Why the file structure change?

The current file architecture was made to split the lib into three projects: discojs-node, discojs and discojs-core. This split made it clearer which code was executed where and fixed a long-lasting problem with TF.js, the problem being that TF.js caused conflicts when importing `@tensorflow/tfjs` and `@tensorflow/tfjs-node` at the same time. 

To enable this, discjos-core was compiled for both Node and the browser, in `dist/node/` and `dist/browser/`, each compiled dist having its own imported TF.js: either `@tensorflow/tfjs-node` or `@tensorflow/tfjs`. The discojs-node and discojs projects then install the discojs-core package locally with `npm i ../discojs-core` to have access to the `@epfml/discjos-core` package and the TF.js version adequate to the JS environment.

Although this file architecture works well in the dev environment, the *local* dependency of discojs-core in discojs-node and discojs isn't exportable when publishing to NPM. Hence, discojs-core would've had to become an NPM package of its own. This wouldn't have been ideal because:
- It's not really supposed to be used on its own
- Installing and importing discojs-core would mean installing and importing its two dists (twice the code memory)
- The two dists of discojs-core would've had to be published, meaning publishing two `dist/node/index.ts` and `dist/browser/index.ts`, meaning the package would need two different entry points—which is not possible. This meant we needed to publish discojs-core-node and discojs-core-browser: and here we ran into a dead-end.

The solution ended up being the new file structure brought in this PR, where discojs-core is symlinked into `discojs-node/src/core` and `discojs-web/src/core`. Both of the project would then compile by using their respective symlinked `src/core` which is done automatically as the code is already lying within the project, since it's symlinked. Publishing `@epfml/discojs` and `@epfml/discojs-node` now becomes straightforward.

The regroup into the parent discojs dir allowed us to go further and regroup all the three projects' dependencies into one package.json lying next to discojs-core, discojs-web and discojs-node. This means a single `npm ci` is required, and the node modules aren't duplicated like before.

Note: @morganridel and I also considered using TypeScript project references, but the few advantages (e.g. incremental compilation) were not worth the struggle.

## Dev experience
All this means the discojs architecture gained in complexity. Yet, the dev experience should be radically simpler build- and test-wise; one can install the Disco.js dependencies, build the libs and test them with `npm ci && npm run build && npm test`.
Furthermore, packages published to npm have a very clean `dist/` arch:
- `discojs/` or `discojs-node/`
  - `index.{js, d.ts}`
  - `core/`
  - `... // custom definitions`

## One last thing...
The following directories were renamed and scripts were updated accordingly:
- information -> docs
- browser -> web-client
- benchmark -> cli
- disocjs/discojs -> discojs/discojs-web


discojs/package.json
```diff
@@ -16,7 +14,11 @@
   },
   "homepage": "https://github.com/epfml/disco#readme",
   "dependencies": {
-    "@epfml/discojs-core": "file:../discojs-core/dist/browser",
+    "@epfml/discojs-node": "file:./discojs-node",
```
morganridel: Can we move this to devDependencies? If it's used only in test it doesn't need to be here

discojs/publish.ts
```diff
@@ -0,0 +1,26 @@
...
+const discojsCore = require(DISCOJS_CORE)
+const discojs = require(DISCOJS)
+const discojsNode = require(DISCOJS_NODE)
+
+const jsDeps = Map(discojsCore.dependencies)
+const nodeDeps = Map(discojsCore.dependencies)
+
+discojs.dependencies = jsDeps
+  .delete('@tensorflow/tfjs-node')
+  .delete('@epfml/discojs-node')
```
morganridel: Also don't need to remove it if it's inside devDependencies

morganridel (changes requested):

s314cy:
the CI passes; next step, update the docs :)

to actually release v2.0.0, git tagging this PR's merge commit should do the job — and trigger the github action handling the npm publication

morganridel (approved):
LGTM minus the docs! hoping that the tinkering with symlinks and prepublish won't bite us later

martinjaggi:
nice! is this PR already causing the release or can we first test some more on the dev branch once it's merged?

s314cy:
> nice! is this PR already causing the release or can we first test some more on the dev branch once it's merged?

the trigger for the npm publication is to make a github release, so we can merge this with dev and publish later :)

PR #488, opened 2022-10-13
More explicit error messages in the web client and library

### LINKED ISSUE #485: Clarify error message for data format mismatch
A user reached out to me after not understanding the following error:
error

He was attempting to upload a dummy dataset in the Titanic task.
It seems that the column names were not aligned.

SOLUTION:
Can we make the error text more explanatory and refer the user to the` data format `tab?
Something like: ⚠️ The data you would like to train on does not conform to the expected input format for the model. Please refer to XYZ"

s314cy:
See #485 for context. This closes #485.

The error handling strategy is to verify what's being fed to the `DatasetBuilder` before starting to train the model, in order to differentiate training errors from dataset errors. This is done on creation of a `{Tabular, Image}Data` object, which is what's produced by the dataset builder.

As a result, three kind of error messages might now be thrown when proceeding to train a model in the web client:
- "Please provide dataset input files": no file was provided by the user
- "Data input format is not compatible with the chosen task": invalid format of the given data files
- "An error occured during training": all other training errors

Additionally, a new `discojs-core/src/dataset/data/` dir was created to welcome the following already existing classes/types: `Dataset`, `Data`, `TabularData`, `ImageData` and `DataSplit`. Quick reminder:
- Dataset is a handy type alias for the raw TF.js dataset type
- Data wraps dataset (and provides two subclasses, TabularData and ImageData)
- DataSplit is the set of the __training__ Data and __validation__ Data

One more thing: I added two scripts, `install.sh` and `build.sh` to ease the installation and build processes of discojs-core, discojs & discojs-node :)


morganridel (approved):

install.sh
```diff
@@ -0,0 +1,6 @@
+DIR="$( cd "$( dirname "$0" )" ; pwd -P )"
+
+cd $DIR/discojs-core && npm ci && dist/ && npm run build &&
```
martinjaggi: just `dist/` or `cd dist/`?

PR #489, opened 2022-10-18
Adding byzantine-tolerant aggregator

francesco98:
### Library (discojs)
New byzantine-tolerant aggregator added. 
New parameters in the TrainingInformation interfaces added to include the byzantine-tolerant aggregator in the aggregation phase and to specify the tau percentile to use while applying the formula.

### Server
Changes in Federated class to include byzantine-tolerant aggregator, if the model wants to use it (looking at TrainingInformation parameters)

### DIFF
--- a/discojs-core/src/weights/aggregation.ts
+++ b/discojs-core/src/weights/aggregation.ts
@@ -21,7 +21,7 @@ function parseWeights (weights: Iterable<WeightsLike | WeightsContainer>): List<
   return r
 }
 
-function centerWeights(weights: Iterable<WeightsLike | WeightsContainer>, currentModel: WeightsContainer): List<WeightsContainer> {
+function centerWeights (weights: Iterable<WeightsLike | WeightsContainer>, currentModel: WeightsContainer): List<WeightsContainer> {
   return parseWeights(weights).map(model => model.mapWith(currentModel, tf.sub))
 }
 
--- a/discojs-core/src/weights/aggregation.ts
+++ b/discojs-core/src/weights/aggregation.ts
@@ -21,6 +21,26 @@ function parseWeights (weights: Iterable<WeightsLike | WeightsContainer>): List<
   return r
 }
 
+function centerWeights(weights: Iterable<WeightsLike | WeightsContainer>, currentModel: WeightsContainer): List<WeightsContainer> {
+  return parseWeights(weights).map(model => model.mapWith(currentModel, tf.sub))
+}
+
+function clipWeights (modelList: List<WeightsContainer>, normArray: number[], tau: number): List<WeightsContainer> {
+  return modelList.map(weights => weights.mapWithIndex((w, i) => tf.prod(w, Math.min(1, tau / (normArray[i])))))
+}
+
+function computeQuantile (array: number[], q: number): number {
+  const sorted = array.sort((a, b) => a - b)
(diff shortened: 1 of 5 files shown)


s314cy (approved):
excellent :)

martinjaggi:
@walidabn want to give us some feedback on this one too, before we merge? - to keep things simple and future proof as later crypto will be added to get SecAgg plus robust

discojs-core/src/weights/weights_container.ts
```diff
@@ -36,6 +36,10 @@ export class WeightsContainer {
     return new WeightsContainer(this._weights.map(fn))
   }
 
+  mapWithIndex (fn: (t: tf.Tensor, i: number) => tf.Tensor): WeightsContainer {
```
martinjaggi, line 39: document this

discojs-core/src/weights/aggregation.ts
```diff
@@ -43,6 +63,24 @@ export function avg (weights: Iterable<WeightsLike | WeightsContainer>): Weights
   return sum(weights).map((ws) => ws.div(size))
 }
 
+// See: https://arxiv.org/abs/2012.10333
```
martinjaggi, line 66: need a bit more doc here to what specifically it is done so that people don't have to read the whole paper

discojs-core/src/weights/aggregation.ts
```diff
@@ -21,6 +21,26 @@ function parseWeights (weights: Iterable<WeightsLike | WeightsContainer>): List<
   return r
 }
 
+function centerWeights (weights: Iterable<WeightsLike | WeightsContainer>, currentModel: WeightsContainer): List<WeightsContainer> {
```
martinjaggi, line 24: can we rename this whole file to sth like `weights_utils` or `weights_math` somehow? (in a separate PR). as there will be many math functions which are useful beyond aggregation only.

and add some more doc strings to avoid any confusion (always be clear if a function applies to one model or a collection of models from different clients say)

francesco98: @s314cy What do you think about it? What's the best option to organize the code?

discojs-core/src/weights/aggregation.ts
francesco98: okay I'll do that in my next PR

discojs-core/src/weights/weights_container.ts
s314cy: it's removed in #490 so less doc to write!

discojs-core/src/weights/aggregation.ts
s314cy: Technically, this file only exports functions manipulating a WeightContainer list, so I would also lean towards a rename. Indeed, only `sum` and `avgClippingWeights` are actual aggregation functions.

I think `weights_math` is good, I can add it to #490. Wdyt @martinjaggi?

martinjaggi: yes, perfect. thanks!

discojs-core/src/weights/aggregation.ts
martinjaggi: better to do it on this one before merging

martinjaggi:
this needs an end to end test at least. can you make a new PR for that?

PR #490, opened 2022-10-18
Overloading of WeightsContainer:map

s314cy:
I approved #489 before realizing the opportunity for a clean function overload on `WeightsContainer:map`

@francesco98


francesco98:
yep, we can do that!

francesco98 (approved):

PR #508, opened 2022-10-28
Fixing issue preventing tasks to be shown

francesco98:
### Discojs library

The two fields RESIZED_IMAGE_H and RESIZED_IMAGE_W have been removed because `preprocessingFunctions` is enough. 

If `ImagePreprocessing.Resize` is a preprocessing function, then the target image size after resizing it is IMAGE_H * IMAGE_W. Otherwise, we already expect to have images of size IMAGE_H * IMAGE_W.

### DIFF
--- a/discojs/discojs-core/src/task/training_information.ts
+++ b/discojs/discojs-core/src/task/training_information.ts
@@ -77,21 +77,24 @@ export function isTrainingInformation (raw: unknown): raw is TrainingInformation
     return false
   }
 
-  // interdepences on data type
-  switch (dataType) {
-    case 'image':
-      if ((IMAGE_H !== undefined && typeof IMAGE_H !== 'number') || (IMAGE_W !== undefined && typeof IMAGE_W !== 'number')) {
-        return false
-      }
-      break
-    case 'tabular':
-      if (!(Array.isArray(inputColumns) && inputColumns.every((e) => typeof e === 'string'))) {
-        return false
-      }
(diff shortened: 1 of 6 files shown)


discojs/discojs-core/src/task/training_information.ts
```diff
@@ -80,7 +80,7 @@ export function isTrainingInformation (raw: unknown): raw is TrainingInformation
   // interdepences on data type
   switch (dataType) {
     case 'image':
-      if (typeof IMAGE_H !== 'number' || typeof IMAGE_W !== 'number') {
+      if ((IMAGE_H !== undefined && typeof IMAGE_H !== 'number') || (IMAGE_W !== undefined && typeof IMAGE_W !== 'number')) {
```
s314cy: I think the best thing to do would be to enforce exactly one of `RESIZED_IMAGE_*` or `IMAGE_*` to be defined, wdyt?

francesco98: I would say only one between them, what do you think?
I cannot imagine a case where you need both, but maybe I'm wrong

s314cy: yes that's what I meant, sorry for the typo

francesco98: okay then it should be in my last commit :)

s314cy (approved):
perfect!

PR #515, opened 2022-10-31
Small fixes for `cli` README and code

batuhanfaik:
- Removed benchmark and fixed script name in README.md
- Renamed `get_training_data` file to `get_training_data.sh` and referenced it in `cli` README
- Fixed `ENOENT: no such file or directory, mkdir <path>` error when no folder named `models` exists in `cli` by recursively creating expected output directories

### DIFF
--- a/server/src/tasks.ts
+++ b/server/src/tasks.ts
@@ -41,6 +41,8 @@ export class TasksAndModels extends EventEmitter {
           console.error(e instanceof Error ? e.message : e.toString())
           return
         }
+        // Create the directory if it does not exist
+        fs.mkdirSync(modelPath, { recursive: true })
         await model.save(`file://${modelPath}`)
       }
       ret.addTaskAndModel(i.task, model)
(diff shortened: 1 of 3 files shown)


batuhanfaik:
@s314cy should I revert to `get_training_data` ?

s314cy:
> @s314cy should I revert to `get_training_data` ?

I think the .sh extension is nice, I tried to align with the `with_server` script, but both can have the .sh extension. You just have to make sure the scripts in the various `package.json` files use the correct file name. The same goes for the GitHub actions .yml config files in `.github/workflows/*.yml`, and the `Dockerfile`.

server/src/tasks.ts
```diff
@@ -41,6 +41,8 @@ export class TasksAndModels extends EventEmitter {
           console.error(e instanceof Error ? e.message : e.toString())
           return
         }
+        // Create the directory if it does not exist
```
s314cy, line 44: it's weird we're getting a "dir missing" error, since the directory has a committed `.empty` file in it and thus exists in the repo 🤷 

I think we want to keep exactly one of the two solutions: either the recursive `mkdir` or the `.empty` file (the latter you can remove)

s314cy:
superseded by #515

Issue #494, opened 2022-10-24
tasks: pretrained models via download links, no need to store on server


martinjaggi:
to keep task definitions self contained and portable, people should just provide a download link if they want to start with a pretrained model (or even for custom architectures). no need to store initial models as model files on our FL server. (intermediate weights, yes, but not models otherwise).

e.g. @francesco98 has done this with some google model hosting already for his use case example.

- [x] change this for the simple-face task
- [x] make it a general option to provide a URL when creating a new task
- [ ] when feeling fancy: provide a checksum hash on the task definition as well, to be secure against any modified or poisoned models hosted at external links. verify this after model download. task provider is responsible to provide a checksum (optional)

this would probably also solve the file reading issue which made us have different node vs web versions?

PR #525, opened 2022-11-01
Add model URL support

s314cy:
## Features

1. Tasks now support the `modelURL` string field. If defined, the Disco server will download the initial model of the task from the remote model URL. This means that tasks defining this parameter don't need to provide a `model` function!

2. The Task Creation Form now also supports model URLs. Before submitting the form, a user must provide either _all_ model files or a model URL. The form prevents the user from providing both.

3. The Task Creation Form received design improvements :)

## Linked issue

This PR is linked to #494, as it adds backend support for the `modelURL` field.


martinjaggi (approved):
looking good. maybe also do it this way in the task creation form (so tasks from there don't need to store models?)

PR #530, opened 2022-11-03
More friendly lib API before v2 publish

s314cy:
## Features

The goal of this PR is to ease the use of the lib by devs willing to integrate discojs into their app. Since this changes the discojs API and is thus a breaking change, it is ideal to include it in the v2 release.

The docs are updated accordingly.

## List of Changes

* the top-level `Disco` class
  * the `Disco` constructor
    * provides default parameters
    * the only required parameter is an object providing server information, i.e. either a URL (of type `string` or `URL`) or a client (of type `disco.Client)`
    * this drastically simplifies the initialization of a Disco client
  * the `Disco:startTraining` method
    * it was renamed to `fit` to fit the Tensorflow and PyTorch standard
    * it includes a call to `Client:connect` 
  * the `Disco:stopTraining` method has been renamed to `pause`, to better describe the behavior
  * added the `Disco:close` method that both pauses training and closes the connection to the remote server (calls both `Disco:pause` and `Client:disconnect`)
* namespaces
  * the `dataset` namespace was renamed to `data`
  * the `dataset.data_loader.*` objects are now directly accessible at `data.*`


discojs/discojs-core/src/tasks/simple_face.ts
```diff
@@ -1,4 +1,4 @@
-import { dataset, Task } from '..'
+import { data, Task } from '..'
```
martinjaggi, line 1: does this task actually need `data`? the other tasks don't

martinjaggi (approved):

discojs/discojs-core/src/tasks/simple_face.ts
s314cy: we have 3 tasks using data preprocessing (lus-covid, geotags, simple-face), but since we're within the disco lib, there are multiple ways to import the `ImagePreprocessing` enum used by all three classes

for instance, lus-covid and geotags import the enum directly from `src/dataset/index.ts`, wheras simple-face first imports the `data` namespace from `src/index.ts` and then accesses the enum

this makes no difference functionality-wise, but I'll modify geotags and lus-covid to import the enum in the standard way

PR #531, opened 2022-11-03
improve server API and prepare future publishing

morganridel:
## Changes

Changes to the server API (i.e. the server lib, not the exisiting HTTP API):
* TaskProvider interface allowing anybody to implement their own tasks
* Model can be provided via the interface, directly, or via URL
* Some update to the server README to explain how to use the server and add own task

## Example usage

We can now use the server like this:

```
const disco = new Disco()
await disco.addDefaultsTasks() // Load tasks defined locally in typescript

// add more custom task
await disco.addTask(...)
await disco.addTask(...)

// Run the server
disco.serve(port)
```

It allows a external user to add its own task before running the server without changing any source code. The other way to add tasks currently is already existing and it is directly from the client by POST call to the `/task` server endpoint

The beginning of the README contains the server example usage above, as it will be published on NPM later, so that user can directly find how to use the server.

## Unit tests

Moreover, some tests of discojs-core don't depend on discojs-node anymore. Lots of objects are now mocked instead, to respect the concepts of unit testing.


s314cy:
since the server did not provide an API beforehand, we can consider this a feature addition for v2.1.0 :)

this means v2.1.0 will include a `@epfml/disco-server` NPM package, as well as the discojs libs bumped to v2.1.0, wdyt?

morganridel:
> since the server did not provide an API beforehand, we can consider this a feature addition for v2.1.0 :)
> 
> this means v2.1.0 will include a `@epfml/disco-server` NPM package, as well as the discojs libs bumped to v2.1.0, wdyt?

Yes, however the setup of publishing the server will be done in another PR, this just updates the code

s314cy (changes requested):
awesome work :)

just a tiny detail, the current naming convention would require to rename `discojs-core/src/default-tasks` to `discojs-core/src/default_tasks`, since the middle-dash is reserved to project names whereas modules use lower_dash naming

martinjaggi:
in the PR description or issue could you describe a bit in more details what you want the API to support (now and in the future?).

also between the server readme and task.md there seems to be a bit of parallels but they don't link to each other or explain which one is for what purpose. try to start the two docs with the user need (who is it for, what will they get out of it if they follow each readme).

note for decentralized, we might aim that a single server should ideally be enough (still everyone should be able to propose new tasks there).  this would be the most common use case. no new server needed then if someone wants to add a new task, right?
while for federated, yes some people in special cases want to run their own server.

s314cy:
> in the PR description or issue could you describe a bit in more details what you want the API to support (now and in the future?).
>
> also between the server readme and task.md there seems to be a bit of parallels but they don't link to each other or explain which one is for what purpose. try to start the two docs with the user need (who is it for, what will they get out of it if they follow each readme).

The server docs are outdated, so I'll update them to clearly state the server's HTTP endpoints and how to use the — future — server's NPM package (i.e. the API brought in this PR).

> note for decentralized, we might aim that a single server should ideally be enough (still everyone should be able to propose new tasks there). this would be the most common use case. no new server needed then if someone wants to add a new task, right? while for federated, yes some people in special cases want to run their own server.

Currently, the discojs lib doesn't provide any API for pushing new tasks to the server in a programmatic way. This is only available in the web client via the task creation form, which pushes to the correct HTTP endpoint of the server—which supports this. I created issue #537.

What this PR brings is an API to add new tasks to the server in a programmatic way **as the server host** with the server lib, which is independent of users pushing new tasks via the discojs lib. So yes, no need for a new server to add a new task :)

s314cy (approved):

server/package.json
```diff
@@ -1,18 +1,21 @@
 {
   "name": "@epfml/disco-server",
-  "private": true,
-  "main": "src/index.ts",
+  "version": "2.0.0",
```
s314cy, line 3: 2.1.0 right?

s314cy: also, maybe we can already bump the version in @epfml/discojs and @epfml/discojs-node already

PR #535, opened 2022-11-07
Add digest verification of downloaded model files

### LINKED ISSUE #494: tasks: pretrained models via download links, no need to store on server

s314cy:
## Feature

When a task specifies a URL to download its initial model from, it may provide a digest to ensure the authenticity of the downloaded files. Supported algorithms are the ones supported by the OpenSSL version of the server. See the [docs](https://nodejs.org/api/crypto.html#cryptocreatehashalgorithm-options) for more info.

## Possible improvements

Don't store the downloaded files on the file system before verifying the digest. This means manually downloading the model files and verifying their content, instead of using `tf.loadLayersModel` which directly converts the downloaded files to a model object.

## Related issue

Fixes #494

### DIFF
--- a/discojs/discojs-core/src/task/digest.ts
+++ b/discojs/discojs-core/src/task/digest.ts
@@ -1,0 +1,24 @@
+export interface Digest {
+  algorithm: string
+  value: string
+}
+
+export function isDigest (raw: unknown): raw is Digest {
+  if (typeof raw !== 'object') {
+    return false
+  }
+  if (raw === null) {
+    return false
+  }
+
+  const { algorithm, value } = raw as Record<string, unknown | undefined>
+
+  if (!(
+    typeof algorithm === 'string' &&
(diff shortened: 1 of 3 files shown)


morganridel (approved):
LGTM

PR #540, opened 2022-11-14
Handy task functions

### LINKED ISSUE #537: discojs API: nice features to have
The discojs lib should:

- allow a client to easily push a locally created task to the disco server, similarly to what the web client offers with its task creation form
- not expect the user to manually fetch the list of tasks before creating the top-level `Disco` object

s314cy:
## List of changes

- new `pushTask(url: URL, task: Task, model: tf.LayersModel)` function for a client to create a new task server-side with a single POST request on `/tasks`, similar to what the task creation form offers
- new `fetchTasks(url: URL)` function to fetch the tasks stored server-side with a single GET request on `/tasks`
- updated the web-client accordingly
- fixed the node-example which broke after merging PR #531

## Related issue

Fixes #537


docs/node_example/start_server.ts
```diff
@@ -30,5 +29,5 @@ export async function startServer (): Promise<[http.Server, string]> {
     throw new Error('unable to get address to server')
   }
 
-  return [server, addr]
-}
+  return [server, new URL('', `http://${addr}`)]
+}
```
morganridel, line 33: EOF linebreak

morganridel (approved):
Nice

PR #543, opened 2022-11-15
Inference feature in the validation UI of the web client

### LINKED ISSUE #495: Inference feature in the validation UI of the web client
## Feature

In the web client, the validation UI allows users to get a prediction accuracy (and possibly other metrics #367) on given _labelled_ data for their trained models. To go further, it would be nice to get the actual label predictions for given _unlabelled_ data.

## Specifications

A user would be able to either:
- select a single unlabelled data sample and get the label prediction displayed in the Disco web client, or
- select multiple unlabelled data samples and get the label predictions as a downloadable .csv file

francesco98:
### Disco library

1. The predicted values in the assessment method were rounded by using `Math.round`, but for multi-class tasks this might lead to an array of 0s because the largest probability might still be lower than 0.5. Now, we use `tf.argMax`.
2. The assessment method now returns also an object containing the predicted values and the ground truth.
3. A new method to only making predictions has been added.
4. The displayInformation in task definition allows to define how to show labels (TEXT vs GPS). We might add other types of visualization.
5. You can now also download predictions

### Web-client

Screenshot 2022-11-24 at 15 44 59

Screenshot 2022-11-24 at 15 45 16

1. The assessment view also shows predictions vs ground truth table to check where the model is doing mistakes.

**IMAGE TASK**
Screenshot 2022-11-22 at 10 24 09

**TABULAR TASK**
Screenshot 2022-11-23 at 16 45 21

3. You can see predictions on a map (The following URL has been implemented for the specific task: https://disco-polygon.web.app/?cellIds=0,1&colors=274C78,FF0000)
Screenshot 2022-11-22 at 13 56 18

4. If the ground truth is not provided, you can only make predictions given a set of samples
Screenshot 2022-11-22 at 10 25 44


web-client/src/components/validation/Predictor.vue
```diff
@@ -0,0 +1,105 @@
+<template>
+  <div class="space-y-8">
+    <!-- test the model -->
+    <ButtonCard
+      class="mx-auto mt-10 lg:w-1/2"
+      :button-placement="'center'"
+      @action="inferModel()"
+    >
+      <template #title>
+        Infer your model
```
martinjaggi: just a language thing: let's change to sth like `Make predictions using your model`. i would also change the code names (doing inference is fine, but infering the model in statistics means training the model, so this will be confusing later on)

martinjaggi (changes requested):

web-client/src/components/validation/Predictor.vue
francesco98: Okay, sure!

martinjaggi (approved):

s314cy (approved):
excellent work @francesco98 ;)

PR #545, opened 2022-11-15
Download and infer from remote models

### LINKED ISSUE #497: Downloadable models in the validation panel
## Feature

In the web client, the validation panel allows a user to select a previously Disco-trained model to validate against data. This means, for a model to be validated with the web client, it has to be loaded in IndexedDB via the training panel first. This corresponds to the following steps:
- go to the task list
- select your task (the model is then downloaded)
- train the task for at least one round
- go to the validation panel
- select the model

It would be more intuitive to allow users to directly download models from the validation UI, and directly proceed to validation. This would result in the following steps:
- go to the validation panel
- select the model to download (the model is then downloaded)

## Specifications

The validation UI would list downloadable pre-trained models next to the ones saved in IndexedDB. The rest the validation pipeline would remain the same.

## Linked issues

The concept of downloadable models has to be thought through, as issue #494 also needs this concept in Disco.js.

s314cy:
## List of changes

* server
  * prints prettier console messages
* web-client/validation 
  * displays a list of the latest models available on the remote server (one for each task)
  * allows the user to directly validate one of such remote models
  * uses the composition API
* discjos/memory
  * supports saved model duplicates

## Related issue

Fixes #497

### DIFF
--- a/web-client/src/components/validation/Validation.vue
+++ b/web-client/src/components/validation/Validation.vue
@@ -24,60 +24,129 @@
       </div>
     </div>
     <div v-show="validationStore.step === 0">
-      <div
-        v-if="memoryStore.models.size > 0"
-        class="grid gris-cols-1 md:grid-cols-2 lg:grid-cols-3 items-stretch gap-8 mt-8"
-      >
-        <div
-          v-for="[path, metadata] in memoryStore.models"
-          :key="path"
-          class="contents"
-        >
-          <ButtonCard
-            :click="() => selectModel(path)"
-            :button-placement="'left'"
-          >
+      <div class="flex flex-col gap-16">
+        <div v-if="memoryStore.models.size > 0">
+          <IconCard>
             <template #title>
-              {{ taskTitle(metadata.taskID) }}
+              Model Library — <span class="italic">Locally Available and Ready to Test</span>
             </template>
-            <template #text>
-              <div class="grid grid-cols-2 justify-items-between">
-                <p class="contents">
-                  <span>Model:</span>
-                  <span>{{ metadata.name.substring(0, 20) }}</span>
-                </p>
-                <p class="contents">
-                  <span>Date:</span>
-                  <span>{{ metadata.date }} at {{ metadata.hours }}</span>
-                </p>
-                <p class="contents">
-                  <span>Size:</span><span>{{ metadata.fileSize }} kB</span>
-                </p>
+            <template #content>
+              Test any model below against a validation dataset.
+              The models listed were downloaded from the remote server.
+              Perhaps you even contributed to their training!
+              Note that these models are currently stored within your browser's memory.
+              <div class="grid gris-cols-1 md:grid-cols-2 lg:grid-cols-3 items-stretch gap-8 mt-8">
+                <div
+                  v-for="[path, metadata] in memoryStore.models"
+                  :key="path"
+                  class="contents"
+                >
+                  <ButtonCard
+                    :click="() => selectModel(path)"
+                    :button-placement="'left'"
+                    class="shadow shadow-disco-cyan"
+                  >
+                    <template #title>
+                      {{ taskTitle(metadata.taskID) }}
+                    </template>
+                    <template #text>
+                      <div class="grid grid-cols-2 justify-items-between">
+                        <p class="contents">
+                          <span>Model:</span>
+                          <span>
+                            {{ metadata.name.slice(0, 20) }}
+                            <span v-if="metadata.version !== undefined && metadata.version !== 0">
+                              ({{ metadata.version }})
+                            </span>
+                          </span>
+                        </p>
+                        <p class="contents">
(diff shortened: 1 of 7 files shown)


Annie-LiGHT:
Wow. This is beautiful! 💅 

Some minor comments

- Change "Registered models" --> "Selected models"
- There is "placeholder" text that says "placeholder". not sure if that remains visible on the platform?
- It is slightly unclear that the "registered model" box _includes_ the models cards underneath (a box is usually enclosing its content, whereas here it is acting as a heading). Is there a way to either put the model cards in the box or remove the box and put free-floating heading with a line separating them?

s314cy:
Thanks for the feedback!

 >  Change "Registered models" --> "Selected models"

I would go with "Models Stored Locally" and "Remote Models"? Both texts will have an explanation below anyway.

>  There is "placeholder" text that says "placeholder". not sure if that remains visible on the platform?

It won't appear on the platform :)

>  It is slightly unclear that the "registered model" box _includes_ the models cards underneath (a box is usually enclosing its content, whereas here it is acting as a heading). Is there a way to either put the model cards in the box or remove the box and put free-floating heading with a line separating them?

I'll try and include the models in their boxes!

Annie-LiGHT:
hmmm "remote model" might imply that it is working remotely.


Maybe 
- "Your model library _(downloaded and ready to train)_" 

- "Select a model from the DISCO repository to add to your library"

s314cy:
> hmmm "remote model" might imply that it is working remotely.
> 
> Maybe
> 
>     * "Your model library _(downloaded and ready to train)_"
> 
>     * "Select a model from the DISCO repository to add to your library"

What do you think?

Screenshot 2022-11-16 at 15 24 19

Annie-LiGHT:
NICE!
Micro comment: "DISCO model repository" sounds a bit better

web-client/src/components/validation/Validation.vue
```diff
@@ -24,60 +24,122 @@
...
-          <ButtonCard
-            :click="() => selectModel(path)"
-            :button-placement="'left'"
-          >
+      <div class="flex flex-col gap-16">
+        <div v-if="memoryStore.models.size > 0">
+          <IconCard>
             <template #title>
-              {{ taskTitle(metadata.taskID) }}
+              Model Library — <span class="italic">Downloaded and Ready to Test</span>
```
martinjaggi: should not be called 'downloaded'. they grow and stay locally. `locally available and ready to test`?

web-client/src/components/validation/Validation.vue
```diff
@@ -24,60 +24,122 @@
...
-                <p class="contents">
-                  <span>Date:</span>
-                  <span>{{ metadata.date }} at {{ metadata.hours }}</span>
-                </p>
-                <p class="contents">
-                  <span>Size:</span><span>{{ metadata.fileSize }} kB</span>
-                </p>
+            <template #content>
+              Test any model below against a validation dataset.
+              The models listed were downloaded from the remote server.
```
martinjaggi, line 35: clarify that this is only for federated tasks, for decentralized this is not possible.

BTW for the future, do we also want to offer sharing of models via providing an URL in general (i can email it to you e..g.), even outside FL?

martinjaggi (changes requested):

web-client/src/components/validation/Validation.vue
s314cy: > clarify that this is only for federated tasks, for decentralized this is not possible.

I changed it so it only displays models for federated tasks 👍 

> BTW for the future, do we also want to offer sharing of models via providing an URL in general (i can email it to you e..g.), even outside FL?

do you mean for the initial model, on task creation?

web-client/src/components/validation/Validation.vue
```diff
@@ -24,60 +24,130 @@
...
+            <template #title>
+              <span class="font-disco font-normal text-xl">DIS</span>
+              <span class="font-disco font-normal text-xl text-disco-blue">CO</span>
+              Model Repository — <span class="italic">Download and Test</span>
+            </template>
+            <template #content>
+              Select any model below to download it. For federated tasks only.
+              The models listed are not currently stored in your browser's memory,
+              but are available and downloadable from the remote Disco server; the same server
+              used during federated and decentralized training.
```
martinjaggi: i'd remove the last sentence on fl and dec.

web-client/src/components/validation/Validation.vue
martinjaggi: future model sharing: can be at any stage of training. the interface is always my local model library. i'll save some good ones when accuracy is good, and later i might want to share some of these checkpoints with you, for your purposes of doing inference for example, or to continue training/finetuning from my checkpoint instead of from the initial model of the task. the library already supports saving to local disk, but not import yet (except what you did now with FL server)

anyway this is not time sensitive

PR #550, opened 2022-12-01
clarify TASK docs

morganridel:
I clarified some part of the tasks docs to better reflect the task workflow after the refactoring (`TaskProvider` interface which wraps `Task` and `tf.LayersModel`).

I also tried to remove some duplicated information and fix broken links.

### DIFF
--- a/docs/TASK.md
+++ b/docs/TASK.md
@@ -48,13 +48,11 @@ I am a user who wants to define my custom task and bring my model to Disco, with
 
 ## 3) Procedure for adding a custom task
 
-In order to add a new custom task to Disco.js, we first need to create and export a new instance of the `Task` class, defined [here](../discojs/discojs-core/src/task/task.ts).
-Then, we must also export a function called `model` that specifies a model architecture for the task.
-We have to remember to export the task and the function in the `index.ts` [file which lives in the same folder](../discojs/discojs-core/src/tasks/index.ts).
-Finally, we need to rebuild Disco.js: `cd discojs/ && npm run build`
+In order to add a new custom task to Disco.js, we need to have defined a `TaskProvider` which need to implement two methods:
+   * `getTask` which returns a `Task` as defined [here](../discojs/discojs-core/src/task/task.ts), the `Task` contains all the crucial information from training to the mode
+   * `getModel` which returns a `Promise<tf.LayersModel>` specifying a model architecture for the task
 
-
-
+You can find examples of `TaskProvider` currently used in our Disco server in `discojs/discojs-core/src/default_tasks/`. These tasks are all loaded by our server by default.
 
 ### Task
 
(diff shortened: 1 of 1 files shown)


docs/TASK.md
```diff
@@ -138,80 +153,69 @@ to use preprocessing functions to resize the data (we also describe how to add c
...
+        validationSplit: 0.2,
+        batchSize: 10,
+        preprocessFunctions: [],
+        learningRate: 0.001,
+        modelCompileData: {
+          optimizer: 'sgd',
+          loss: 'categoricalCrossentropy',
+          metrics: ['accuracy']
+        },
+        modelID: 'my_new_task-model',
```
martinjaggi, line 187: a bit confused why there are two tasks here, `simple-face` and `my_new_task`?

docs/TASK.md
```diff
@@ -48,87 +48,102 @@ I am a user who wants to define my custom task and bring my model to Disco, with
...
 ### Task
 
 For the task creation, we consider the main use case which does not go through the user interface : 
 
 **I am a developper who wants to define my own task**
 
-In this case, your model and task will be uploaded and stored on our DISCO servers. You will have to make the task visible to the API. For your custom model, the JSON model architecture is necessary, but the .bin weight file is optional : if you include the weights file, your model will be loaded with the passed weights. If a weights file is not specified, the weights for the model will be initialized randomly.
+If you want to add a new task to our production DISCO server you have two possibilities:
+  * using the user interface as described above
+  * exporting your own `TaskProvider` from `discojs/discojs-core/src/default_tasks/`  and adding a new default task by contributing to the code.
```
martinjaggi, line 65: maybe say first one is without any coding required, second one is with describing the ML task in your own code

docs/TASK.md
```diff
@@ -48,87 +48,102 @@ I am a user who wants to define my custom task and bring my model to Disco, with
...
+To export a new task in the code, make sure to export the `TaskProvider` in the `discojs/discojs-core/src/default_tasks/index.ts` file as follows:
 
+```js
+export { cifar10 } from './cifar10'
+export { lusCovid } from './lus_covid'
+export { mnist } from './mnist'
+export { titanic } from './titanic'
+export { simpleFace } from './simple_face'
+export { myNewTask } as my_new_task from './my_new_task' // <---- including our new custom task!
+export { geotags } from './geotags'
```
martinjaggi, line 76: or new one last?

docs/TASK.md
```diff
@@ -48,87 +48,102 @@ I am a user who wants to define my custom task and bring my model to Disco, with
...
+```
 
-### Making the task visible to the API
+> Note that `discojs-core` must only contain platform-agnostic code that works both in the browser and on Node.js.
+> Thus, if your task requires reading some file from your local file system, you need to define the task in `discojs-node` only: `discojs/discojs-node/src/default_tasks/my_new_task.ts`
```
martinjaggi, line 80: could you comment on the easiest case first, where is the model usually read from (and mention can be initial model, or pre-trained model)

docs/TASK.md
```diff
@@ -48,87 +48,102 @@ I am a user who wants to define my custom task and bring my model to Disco, with
...
-Start by creating a function in `my_new_task.ts` called `model` that returns the `tf.LayersModel`. In this example we programatically 
-define our model.
 
-After exporting the Task, you need to also export a function called `model` that returns the layers model. If you use a 
-pre-trained model, you can simply load and return said model in the function via `tf.loadLayersModel(modelPath)`.
+### Model
 
-> Note, it's important to add the `export` tag to the model as well as to the subsequent objects that we define.
+The interface let you load your model however you want, as long as you return a `tf.LayersModel` at the end. If you use a 
+pre-trained model, you can simply load and return said model in the function via `tf.loadLayersModel(modelPath)`.
```
martinjaggi, line 124: (same comment as above, is the default recommendation some URL, or file-system access on the phone, or server?)

docs/TASK.md
```diff
@@ -48,87 +48,102 @@ I am a user who wants to define my custom task and bring my model to Disco, with
...
-
-```js
-const simpleFaceModelPath = path.join(CONFIG.modelsDir, 'mobileNetV2_35_alpha_2_classes', 'model.json')
-
-// TODO, to add a custom model for a task, add the path here
-const MODEL_PATH = Map<string, string>()
-  .set(defaultTasks.simple_face.task.taskID, simpleFaceModelPath)
-
-```
+At runtime, the models are stored in `disco/server/models/`, and it is also in the server side that we let disco know where exactly they are saved. In particular,
```
martinjaggi, line 146: do you mean the initial models for each task, or even the freshly trained ones? (i guess it's true for FL but not for decentralized?)

martinjaggi (changes requested):
thanks a lot, this is super helpful.  i just added some minor comments above from the perspective of a naive user :)

PR #557, opened 2022-12-12
Implementing end-to-end test with Byzantine-Robust Aggregator for cifar10 task

### LINKED ISSUE #556: Missing Unit Test for Byzantine-robust aggregator
We recently added the byzantine-robust aggregator, but we did not have the chance to truly test it, so we should add a unit test to show that it has an effect. 

The idea is to add 2 unit-test:
1. We simulate all good actors and we prove that the accuracy is almost the same when using the standard aggregator vs the byzantine-robust aggregator.
2. We simulate a bad actor (e.g.: flipping label attack) and we probe that the accuracy is better when using a byzantine-robust aggregator.

francesco98:
### Discojs

- Changed cifar10 task definition to comply the input size required by the network. Indeed, MobileNet is using an input size of 224x224x3, while the example data (and the previous specified input shape) was 32x32x32.
- Preprocessing function to normalize tabular data has been added

### Server
Changed the end-to-end test case for the federated setting to use both standard aggregator and byzantine-robust one. Also, the test is now considering the presence of multiple users, splitting the training data among them and validating the model computing the accuracy after the training phase. As last step, an assertion has been added to expect a reasonable accuracy after training.


discojs/discojs-core/src/client/base.ts
```diff
@@ -52,6 +52,7 @@ export abstract class Base {
   abstract onRoundEndCommunication (
     updatedWeights: WeightsContainer,
     staleWeights: WeightsContainer,
+    updatedMomentum: WeightsContainer,
```
martinjaggi, line 55: not sure if we need to send around global momentum. better to only use local momentum to determine the clipping centers

discojs/discojs-core/src/client/decentralized/base.ts
```diff
@@ -165,6 +165,7 @@ export abstract class Base extends ClientBase {
   async onRoundEndCommunication (
     updatedWeights: WeightsContainer,
     staleWeights: WeightsContainer,
+    _: WeightsContainer, // TODO: Implement Byzantine-Robust Aggregator in Decentralized setting
```
martinjaggi, line 168: is it needed?

discojs/discojs-core/src/dataset/data/index.ts
```diff
@@ -2,4 +2,4 @@ export { DataSplit } from './data_split'
 export { Data } from './data'
 export { ImageData } from './image_data'
 export { TabularData } from './tabular_data'
-export { ImagePreprocessing } from './preprocessing'
+export { ImagePreprocessing, TabularPreprocessing } from './preprocessing'
```
martinjaggi, line 5: do we have tabular as a data modality in core disco now already? if yes can we re-base this PR?

discojs/discojs-core/src/training/trainer/distributed_trainer.ts
```diff
@@ -23,18 +24,25 @@ export class DistributedTrainer extends Trainer {
    * Callback called every time a round is over
    */
   async onRoundEnd (accuracy: number): Promise<void> {
+    const beta = 0.5 // TODO: Add parameter to task definition
```
martinjaggi, line 28: agree with the TODO (and say it's only for Byz robust, not used if robustness is turned off, right?)

martinjaggi (changes requested):
added some comments. one goal to keep in mind is to make the Byzantine robust clipping such that it can later also be used in decentralized. so let's try to define the clipping center based on local info only (and try to minimize additional communication if any).

we should also write down some thoughts&docu on how it will work with async (whenever the buffer size is reached, we do an aggregation with that group, so that's a round)

discojs/discojs-core/src/client/base.ts
francesco98: this is basically the `m_i` in the formula (i.e.: the momentum computed by the client `i`) which is then sent to the server. It's not the global momentum, but refers to a specific local client.

discojs/discojs-core/src/client/decentralized/base.ts
francesco98: The parent class of `decentralized/base.ts` defines a third parameter `updatedMomentum` that is not used in decentralized setting, but we cannot avoid to put it in the signature because it is inherited from the parent class. However, it will be useful when implementing the Byzantine aggregator also in the decentralized setting.

discojs/discojs-core/src/training/trainer/distributed_trainer.ts
francesco98: yes, true it is only for Byzantine.

discojs/discojs-core/src/client/base.ts
martinjaggi: so i think this one doesn't need to be sent. it's computed locally and remains on the client, no?

discojs/discojs-core/src/training/trainer/distributed_trainer.ts
martinjaggi: TODO

discojs/discojs-core/src/client/base.ts
francesco98: The server uses all the agents' momentum vectors to compute the global shared momentum. Thus, it needs to know all agents' values, right? At least in my implementation, I use these values on the server to first computer the centered part and then the clipping factor while computing the updated global shared momentum. Depending on whether the model use or not the Byzantine-Robust Aggregator, each client sends the updatedWeights (potentially with the privacy-preserving computations) or updatedMomentum.

s314cy (approved):

PR #558, opened 2022-12-28
Fixing up assess(...) function

francesco98:


discojs/discojs-core/src/validation/validator.ts
```diff
@@ -18,32 +18,48 @@ export class Validator {
...
     if (batchSize === undefined) {
       throw new TypeError('batch size is undefined')
     }
 
-    const labels: string[] | undefined = this.task.trainingInformation?.LABEL_LIST
-    const classes = labels?.length ?? 1
-
     const model = await this.getModel()
 
+    let features: Array<number | number[] | number[][] | number[][][] | number[][][][] | number[][][][][]> = []
```
s314cy: the `xs` vector is always a batch of samples, hence we usually directly cast it as `number[][]` to avoid carrying TF.js's bad typing

discojs/discojs-core/src/validation/validator.ts
```diff
@@ -18,32 +18,48 @@ export class Validator {
...
         const xs = e.xs as tf.Tensor
-        const ys = (e.ys as tf.Tensor).dataSync()
 
-        const pred = (model.predict(xs, { batchSize: batchSize }) as tf.Tensor)
-          .dataSync()
-          .map(Math.round)
+        const ys = this.getLabel(e.ys as tf.Tensor)
+        const pred = this.getLabel(model.predict(xs, { batchSize: batchSize }) as tf.Tensor)
+
+        const currentFeatures = xs.arraySync()
```
s314cy, line 49: you might have to cast the result of `arraySync` to `number[][]` for the reason mentioned above :)

discojs/discojs-core/src/validation/validator.ts
```diff
@@ -18,32 +18,48 @@ export class Validator {
...
+
+        if (Array.isArray(currentFeatures)) {
+          features = [...features, ...currentFeatures]
```
s314cy: `a = a.concat(b)`

discojs/discojs-core/src/validation/validator.ts
```diff
@@ -18,32 +18,48 @@ export class Validator {
...
+        } else {
+          throw new TypeError('features array is not correct')
+        }
+
+        groundTruth.push(...Array.from(ys))
```
s314cy, line 57: maybe `getLabel` could call `arraySync` instead of `dataSync` (and cast) to avoid manually constructing an array

s314cy (changes requested):
thanks very much for the changes and bugfix :)

s314cy (approved):

PR #565, opened 2023-02-06
add documentation in TASK for nlp and lstm task

### LINKED ISSUE #491: add NLP usecase
add a new task as an example for NLP such as text classifiers. can probably use the `.csv` dataloaders for it.

 - Plan A: MobileBert
Preprocessing pipeline and pretrained model for TFjs is already available here:
https://github.com/tensorflow/tfjs-models/tree/master/qna
the question is if the model is compiled graph format (inference only) or could be available as layered model as well for training.  see also huggingface for more [pretrained model infos], potentially converting to keras-style layered model (https://huggingface.co/docs/transformers/model_doc/mobilebert), and the google [blog post](https://blog.tensorflow.org/2020/03/exploring-helpful-uses-for-bert-in-your-browser-tensorflow-js.html) on the TFjs one

- Plan B: LSTM
If finetuning mobileBert as above doesn't work in js yet, we can try simpler LSTM models as for example here: https://blog.ldtalentwork.com/2020/01/10/tensorflowjs-how-to-create-a-language-translator/

In either case, what's nice is that we have existing tokenizers/preprocessing code already available so don't need to start from scratch

what do others think?

walidabn:
closes #491 

Simple nlp task with lstm to do sentiment analysis. We document in TASK.md the details of this NLP task, along with our tokenization preprocessing function.


martinjaggi:
can you rename the PR as it's not just documentation but also the code for NLP tasks?

in terms of code, the different nodes currently don't use the same vocabulary, because not based on the a shared tokenizer. the tokenizer should be specified in the task definition. we can wait for #572, then this should be simple afterwards

martinjaggi:
not closing yet

s314cy:
as discussed with @walidabn, this PR is superseded by #577 

the `discojs-core/src/default_tasks/sentiment.ts` task might be included later on as a classic NLP use-case, in addition to LLM's

PR #567, opened 2023-02-13
Add converter script to keras

walidabn:
Fixes issue @480 in the following sense : 

This PR is not meant to be merged on develop, but essentially brings the DeepBreath model to JS. In more details : 
  - The converted DeepBreath model in JS  is in /tfjs_model
  -  The Tensorflow saved_model is in /saved_model
  - Three Python scripts were written, the first one called convertDB.py is a conversion from Pytorch to ONNX, a second one called onnx2TensorFlow.py is a rewritting from the ONNX format to tensorflow/TensorflowJS, but by comparing it to the original Pytorch model, some layers had to be dropped due to incompatibilites between Pytorch and ONNX. The third script is an accurate rewritting from the Pytorch model to Tensorflow/TFJS, called torchToTFModel.py. Torch is not used in it, inside we write a TF/Keras model and then save it.

A few notes : When converting from Pytorch to Tensorflow/TFJS, unless the model is trivial, we should not rely on automatic conversion tools like pytorch2keras, or onnx2keras, as these libraries are out of date and don't support many layers/essential functionalities. Redevelopping the model manually on Tensorflow is the best way to go for these kind of tasks.

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -28,19 +28,15 @@ which should output something similar to `arm64`. Then, `npm i @tensorflow/tfjs`
 
 4. Execute `npm run dev` and you are done!
 
-### Problem: Importing models or weights from different frameworks like Pytorch to TensorflowJS
+### Problem: Importing models or weights from Pytorch to TensorflowJS
 
 ### Solution:
 
-The pipeline for any conversion scheme that takes a model in a framework X (imagine Pytorch as a guiding example) goes usually as follows :
 
-Framework X -> ONNX -> Keras -> TensorflowJS
+There exist several libraries that try to perform automatic conversion between frameworks, which we do not recommend as most of the tools have compatibility issues if the model contains components which differ strongly in implementation between the two frameworks.
+The simplest way to obtain a TensorflowJS model is to first obtain a Tensorflow/Keras model, stored as a .h5 file, and then convert it using TensorflowJS's converter tool, which transforms a Tensorflow/Keras model to TensorflowJS. To obtain the Keras/Tensorflow model from we strongly recommend to directly develop the model in Keras : most of Pytorch components have their equivalent counterpart in Tensorflow/Keras, and translating model architectures between these two frameworks can be done in a straightforward way. One caveat is that, weights currently cannot be converted from the Pytorch `.pth` format to the Keras `.h5` format, so the model must be retrained.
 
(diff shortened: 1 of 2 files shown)


martinjaggi:
in the description/readme could you clarify the 3rd way in some more details? i guess you mean to write a model in python TF/keras and then saving. this is in `torchToTFModel.py` while torch is not used in it, right? 

can you link to our documentation somehow how to take python TF/keras models and convert it to TFjs?

how about the preprocessing pipeline in this case? i guess it remains in python but how did you abstract it?

were you able to test the resulting JS model if it does the same as the pytorch python one? maybe with and without preprocessing?

morganridel:
Did the 70MB+ files "output" needed to be pushed in Git? @s314cy worked a lot to clean the Git history to get a light repo

s314cy:
> Did the 70MB+ files "output" needed to be pushed in Git? @s314cy worked a lot to clean the Git history to get a light repo

yes, such heavy static files should not be pushed to the repo, even if not merged in dev/prod, since the repo will keep track of the files as long as they exist in some commit in some remote branch (so it won't be enough to remove them and commit)

a good solution would be for the branch to only contain the __source code__ that generated the static model files that were pushed; the latter can be gitignored (note that logs and other situational files should not be pushed either, thus they can be gitignored as well)

technically, this means adding all the log, onnx and bin files to the gitignore, git rm them, commit, squash the git rm commit with the commit that added the files, and finally push force

walidabn:
- For Martin's first point, it's exactly that. I edited the message to clarify that point.

- I added and clarified the documentation for model conversion in FAQ.md and adapted the one in TASK.md. I think it's a bit clearer now.

- Adding some details on the preprocessing pipeline : 
The preprocessing pipeline goes as follows : 
convert the raw audio file into a text file using the python preprocessing pipeline abstracted in preprocess.py into a text file with numeric values. In order to do so, we used the following preprocessing pipeline that can be found [here] (https://github.com/epfl-iglobalhealth/DeepBreath-App-Giorgio/blob/main/models/train_models_for_mobile/train_model_disease_classification.py).

Running the script yields a .txt file containing a certain amount of floating point values, which will be a multiple of 32. In order to use the TensorflowJS model on the data, read values from the text file by loading them in a float tensor of size [32,-1]. 

- Comparing performance : As weights cannot be transfered from Pytorch to TensorflowJS, assessing and comparing the performance of the two models cannot be done.

s314cy:
@walidabn should I merge this once the `.py` files are removed? or would you prefer to add the deepbreath task in this PR as well?

s314cy (approved):

PR #569, opened 2023-02-20
Dataset metadata storage UI

morganridel:
UI part isolated for review.

Based and dependent on merge of https://github.com/epfml/disco/pull/568

We can discuss how we should present this feature to the user (naming, interface...)


JulienVig:
Closing along with #568.

PR #577, opened 2023-04-20
Add support for text data & tokenization

### LINKED ISSUE #491: add NLP usecase

### LINKED ISSUE #572: support modern tokenizers for NLP
if we can support recent standard tokenizers (byte-pair, sentencepiece) etc such as used for modern transformer models, we'll have much better compatibility with current NLP models. (and the tokenizers can later be used for any kind of small/big word2vec/lstm/transformer etc models)

the only thing needed at first will be the ability to load and use a predefined (pretrained) tokenizer, such as the ones for say GPT-style models, or from huggingface.

here is a suitable codebase in typescript
https://github.com/botisan-ai/gpt3-tokenizer

let's evaluate if we can use it and if it can load existing pretrained tokenizers (a nice vocabulary size is 2^15 = 32k, but can start smaller for a proof of concept)

s314cy:
closes #572 and closes #491 

add support for text data & tokenization:
- [x] tokenize samples
- [x] load labelled text data in the browser/node
- [ ] load unlabelled text data in the browser/node
- [ ] lazily load text data in node

this PR includes a rework of the data preprocessing pipeline, which is much more modular and makes it easy to add new preprocessing functions!

it also fixes the CI by:
- making the github actions data cache run-specific
- ensuring the data download script bypasses the gbucket cache
- replacing the example data's archive from a BSD tar to a GNU tar (macOS vs. linux) which caused issues in the CI


discojs/discojs-core/src/dataset/preprocessing/text_preprocessing.ts
```diff
@@ -0,0 +1,30 @@
...
+export interface TextTensorContainer extends tf.TensorContainerObject {
+  xs: tf.Tensor1D
+  ys: tf.Tensor1D | number | undefined
+}
+
+export enum TextPreprocessing {
+  Tokenize = 'tokenize'
+}
+
+export function getPreprocessImage (task: Task): PreprocessText {
```
martinjaggi: should this one be called image?

martinjaggi: also mind adding a comment if you will output a stream of token ids?

for LLMs, we can then also support datasets without any label being needed

also let's say where/how people could load different tokenizers (task config or hardcoded either is fine)

martinjaggi:
very cool, thanks for getting this started!

discojs/discojs-core/src/dataset/preprocessing/text_preprocessing.ts
s314cy: I'll make sure that the PR follows your comments once it's out of the "draft" stage!

PR #578, opened 2023-05-03
aggregator and client refactor

LucasTrg:
In the following PR, we separate the aggregation logic to the communication one by splitting `Clients` into `Client`,  `Aggregator` and `Preparator`. 
This will allow to develop more advanced aggregation techniques, different encryption scheme, etc... as well as different network topologies in the future.


server/tests/client/decentralized.spec.ts
```diff
@@ -61,5 +62,5 @@ function test (
   })
 }
 
-test('cleartext', clients.decentralized.ClearText)
-test('secure', clients.decentralized.SecAgg)
+test('cleartext', clients.decentralized.Base)
+test('secure', clients.decentralized.Base)
```
martinjaggi: do both tests still work if called on the same?

server/tests/client/decentralized.spec.ts
```diff
@@ -32,7 +33,7 @@ function test (
         getClient(Client, server, TASK),
         getClient(Client, server, TASK)
       ]))
-
+      const preparators = users.map(u => new ClearPreparator(u, TASK))
```
martinjaggi: add some comments / docstring what preparators are and why needed

discojs/discojs-core/src/default_tasks/index.ts
```diff
@@ -4,3 +4,4 @@ export { mnist } from './mnist'
 export { titanic } from './titanic'
 export { simpleFace } from './simple_face'
 export { geotags } from './geotags'
+export { mnistBandit } from './mnist_bandit'
```
martinjaggi: would you mind having the bandit later in a separate PR, after the refactor (the refactor is bigger)

discojs/discojs-core/src/client/decentralized/base.ts
```diff
@@ -157,46 +159,131 @@ export abstract class Base extends ClientBase {
     this.connected = false
   }
 
-  async onTrainEndCommunication (_: WeightsContainer, trainingInformant: TrainingInformant): Promise<void> {
+  public getPeers (): Set<PeerID> {
+    return this.peers ?? Set()
+  }
+
+  async onTrainEndCommunication (_: Map<PeerID, WeightsContainer>, trainingInformant: TrainingInformant): Promise<void> {
```
martinjaggi: why do we need on trainEndComm being different from the end of round comm?

martinjaggi: why do we need on trainEndComm being different from the end of round comm?

discojs/discojs-core/src/aggregator/mean_agg.ts
```diff
@@ -0,0 +1,36 @@
+import { Map } from 'immutable'
+import { Base as Client } from '../client/base'
+
+import { Base as Aggregator } from './base'
+import { Task, TrainingInformant, WeightsContainer, WeightsMapping, aggregation, tf } from '..'
+
+/**
+ * Aggregator that computes the mean of the weights received from the peers.
```
martinjaggi: is this one used both for secure and cleartext? somehow i don't see the refactor of the SecAgg parts well yet (and trying to keep aggregators compatible with both federated and decentralized as much as possible)

also the robust one by francesco, will it also build on this same class here, or how do they fit together?

server/tests/client/decentralized.spec.ts
LucasTrg: I should actually delete the 'secure' one for the time being. As @swaglid explained to us, the current sec_agg is not working properly.  My goal is to get this PR merged quick enough for him to begin re-implementing secure aggregation.

martinjaggi: the secure decentralized one should be fine, it has unit tests and end2end tests. would be better to keep it or migrate it during the refactor (maybe walid can help?)

did you check with him how it would have to be modified to become compatible with the new aggregator concept here (does the one here include decentralized and federated actually?)

server/tests/client/decentralized.spec.ts
LucasTrg: I already did a tiny bit in `discojs/discojs-core/src/preparator/base.ts`, but plan on describing it in the README.MD.
I think a new section to help future collaborator to add new aggregation mechanism or encryption scheme, etc... could be a great addition, as the point of the refactor is mostly to allow such development !

discojs/discojs-core/src/default_tasks/index.ts
LucasTrg: That's a good idea. It'll make the testing a little bit easier and reduce the scope of the PR a little

discojs/discojs-core/src/client/decentralized/base.ts
LucasTrg: I think it's an oversight from me

discojs/discojs-core/src/aggregator/mean_agg.ts
LucasTrg: As mentionned above. Secure aggregation will be brought by @swaglid. Maybe he can push to this PR to not break any already existing feature, or it can be brought by a separate PR.

I have not caught up with Francesco's work. With @s314cy, we've agreed to first focus on getting this PR to pass tests locally, and then to resolve the pretty big conflict with Francesco's PR.

server/tests/client/decentralized.spec.ts
walidabn: If I recall correctly, after a discussion with Hugo and Lucas, we figured that splitting aggregation strategies into a preparator/aggregator maked more sense than having a unique aggregator in order to reuse code between decentralized and federated. Preparator prepares the extra processing that has to be done on the client so that Aggregator only handles the aggregation (or norm checks in case of clipping).

If we take secure aggregation in federated for example, the preparator would handle all client-side pre-computation (shares of 0 generation and adding them to the weights), while the aggregator would handle the server-side computation, so in this case, just averaging.

In decentralized, preparator/aggregator are both part of the client, while in federated the preparator/aggregator are split between client/server. That enables code reuse of preparators between decentralized and federated.  (correct me if I'm wrong on the idea here @LucasTrg )

discojs/discojs-core/src/aggregator/bandit_agg.ts
```diff
@@ -0,0 +1,65 @@
...
+      totalRelationships += this.relationships?.get(peerId) ?? 1
+    })
+
+    return aggregation.sum(weightedUpdates).map((ws) => ws.div(totalRelationships))
+  }
+
+  update (peerId: PeerID, losses: tf.Scalar[]): void {
+    const oldRelationship = this.relationships?.get(peerId) ?? 1
+    const newRelationship = List(losses).zip(List(this.currentLosses))
+      .map(([loss, currentLoss]) => (loss.sub(currentLoss).div(currentLoss)).exp())
```
s314cy: it would be less costly to exponentiate the sum rather than multiplying the exponentials

discojs/discojs-core/src/aggregator/base.ts
```diff
@@ -0,0 +1,21 @@
...
+  constructor (
+    protected readonly client: Client,
+    private readonly task: Task,
+    protected model?: tf.LayersModel | undefined) {
+  }
+
+  abstract onRoundEndAggregation (weightsToSend: WeightsMapping, currentWeights: WeightsContainer, roundNumber: number, trainingInformant: TrainingInformant): Promise<WeightsContainer>
+  abstract onTrainEndAggregation (weights: WeightsContainer, trainingInformant: TrainingInformant): Promise<void>
+
+  applyDifferentialPrivacy (weightsToSend: WeightsContainer, staleWeights: WeightsContainer): WeightsContainer {
```
s314cy: (moved to preparator)

discojs/discojs-core/src/client/decentralized/base.ts
```diff
@@ -157,46 +159,131 @@ export abstract class Base extends ClientBase {
...
+      } else {
+        throw new Error('weightsToSend is not a Map')
+      }
     } catch (e) {
       let msg = `errored on round ${round}`
       if (e instanceof Error) { msg += `: ${e.message}` }
       console.warn(this.ID, msg)
 
-      return updatedWeights
+      return Map() as WeightsPayload
```
s314cy: useless cast

discojs/discojs-core/src/training/trainer/distributed_trainer.ts
```diff
@@ -27,12 +30,19 @@ export class DistributedTrainer extends Trainer {
...
 
-    const aggregatedWeights = await this.client.onRoundEndCommunication(
+    // It might make sense to move the preparator to the aggregator since they are often closely related
+    const weightsToSend = this.preparator.prepareForRound(
       currentRoundWeights,
       previousRoundWeights,
       this.roundTracker.round,
       this.trainingInformant
     )
+    const aggregatedWeights = await this.aggregator.onRoundEndAggregation(
```
s314cy: I'm not sure to understand, shouldn't the aggregator be used client-side in the decentralized setting only?

discojs/discojs-core/src/types.ts
```diff
@@ -9,3 +11,7 @@ export type Weights = tf.Tensor[]
 export type MetadataID = string
 
 export type Features = number | number[] | number[][] | number[][][] | number[][][][] | number[][][][][]
+
+export type WeightsPayload = WeightsContainer | Map<PeerID, WeightsContainer> | Map<ClientID, WeightsContainer> | Map<number, WeightsContainer>
```
s314cy:
```suggestion
export type WeightsPayload = WeightsContainer | WeightsMapping
```

discojs/discojs-core/src/types.ts
```diff
@@ -9,3 +11,7 @@ export type Weights = tf.Tensor[]
...
+
+export type WeightsMapping = Map<PeerID, WeightsContainer> | Map<ClientID, WeightsContainer> | Map<number, WeightsContainer>
```
s314cy:
```suggestion
export type NodeID = ClientID | PeerID
export type WeightsMapping = Map<NodeID, WeightsContainer>
```

server/tests/end_to_end/decentralized.spec.ts
```diff
@@ -2,13 +2,14 @@ import fs from 'fs/promises'
...
 import { Range } from 'immutable'
-
 import {
-  node, informant, Task, Disco, client as clients, WeightsContainer, defaultTasks
+  node, informant, Task, Disco, client as clients, WeightsContainer, defaultTasks, WeightsMapping
 } from '@epfml/discojs-node'
 
 import { getClient, startServer } from '../utils'
 import { assertWeightsEqual } from '../../../discojs/discojs-core/src/weights/aggregation.spec'
+import { MeanAggregator } from '@epfml/discojs-node/dist/core/aggregator'
```
s314cy: should be imported from the lib rather than the compilation result

s314cy (dismissed):

s314cy:
thanks very much for the changes! let's try and fix the CI this week so that we can tackle the merge conflicts the next one :)

there still remains a few changes to make server-side, right? such as making use of the aggregator classes

discojs/discojs-core/src/aggregator/bandit.ts
LucasTrg, line 1: I think we should remove this one al together and add in a separate PR

discojs/discojs-core/src/aggregator/base.ts
```diff
@@ -0,0 +1,239 @@
+import { client, Task, tf, AsyncInformant } from '..'
+
+import { EventEmitter } from 'events'
+
+import { Map, Set } from 'immutable'
+
+export enum AggregationStep {
```
LucasTrg, line 7: Maybe some doc here could be appreciated so that we better understand what the framework is, and how we can modify it if needed

discojs/discojs-core/src/aggregator/base.ts
```diff
@@ -0,0 +1,239 @@
...
+    })
+  }
+
+  /**
+   * The aggregation result can be awaited upon in an asynchronous fashion, to allow
+   * for the receipt of contributions while performing other tasks. This function
+   * gives access to the current aggregation result's promise, which will eventually
+   * resolve and contain the result of the very next aggregation step, at the
+   * time of the function call.
+   * @returns The promise containing the aggreation result
```
LucasTrg:
```suggestion
   * @returns The promise containing the aggregation result
```

discojs/discojs-core/src/aggregator/base.ts
LucasTrg, line 1: I think some more information is needed on the difference between communication rounds and aggregation rounds for the future devs that will work on this :)

discojs/discojs-core/src/aggregator/index.ts
```diff
@@ -0,0 +1,14 @@
+import { BanditAggregator } from './bandit'
+import { MeanAggregator } from './mean'
+import { RobustAggregator } from './robust'
+import { SecureAggregator } from './secure'
+
+export { Base as AggregatorBase, AggregationStep } from './base'
+export { MeanAggregator } from './mean'
+export { BanditAggregator } from './bandit'
```
LucasTrg:
```suggestion

```

discojs/discojs-core/src/aggregator/index.ts
```diff
@@ -0,0 +1,14 @@
...
+export { RobustAggregator } from './robust'
+export { SecureAggregator } from './secure'
+
+export { getAggregator, AggregatorChoice } from './get'
+
+export type Aggregator = MeanAggregator | BanditAggregator | RobustAggregator | SecureAggregator
```
LucasTrg:
```suggestion
export type Aggregator = MeanAggregator | RobustAggregator | SecureAggregator
```

discojs/discojs-core/src/aggregator/robust.ts
LucasTrg, line 1: Should we include this in a separate PR or do you want to leave it as a draft ?

discojs/discojs-core/src/async_informant.ts
```diff
@@ -1,85 +1,64 @@
...
-
-    // DEBUG
-    console.log('After update')
+    if (this.round === 0 || this.round < this.aggregator.round) {
+      this._round = this.aggregator.round
+      this._currentNumberOfParticipants = this.aggregator.size
+      this._averageNumberOfParticipants = this.totalNumberOfParticipants / this.round
+      this._totalNumberOfParticipants += this.currentNumberOfParticipants
+    } else {
+      this._round = this.aggregator.round
```
LucasTrg, line 22: Not quite sure of this edge case, or could we get a number of round bigger than the number of aggregation round ? 
Maybe some comments in the code could help

discojs/discojs-core/src/async_informant.ts
LucasTrg, line 1: This is making me think  that there is a small distinction that we did not capture for participants. Namely, the participation graph can be directed ("I give you my model, but you did not give me yours"), and I don't think this current implementation would fit. 
It's not a hard requirement yet, more of a food or thoughts comment

discojs/discojs-core/src/default_tasks/index.ts
LucasTrg:
```suggestion

```

discojs/discojs-core/src/default_tasks/mnist_bandit.ts
LucasTrg, line 1: Maybe we could delete this one

LucasTrg (commented):
Overall LGTM, looking forwrd to test the edge case we might run into wth the bandit !

discojs/discojs-core/src/aggregator/base.ts
s314cy: I think I'll add the corresponding docs in the ongoing PR #580, wdyt @martinjaggi?

discojs/discojs-core/src/aggregator/base.ts
s314cy: similarly, maybe I can add this in #580?

discojs/discojs-core/src/aggregator/robust.ts
s314cy: probably another PR :)

discojs/discojs-core/src/async_informant.ts
s314cy: you are right, the current code allows one to define "what" kind of contribution the client should send to its neighbors, but not "who" it should send its contributions to

similarly, the client will expect a contribution (possibly many) from every active neighbor, with an eventual non-failing timeout in the case where no contribution was received, which is is not ideal because it adds peer idleness

both issues can be fully handled by the aggregator by keeping lists of desirable nodes to receive from/send to (subsets of active nodes), in order to differ from the set of nodes usually kept in sync between the client and aggregator

do you need these changes for bandit? would you like to make these changes yourself? if not, I'll make the changes in a separate PR

LucasTrg: I think for the moment I can keep up with this discrepancy by sending empty payloads to the nodes that were not selected for the current round. 
We can clean it up if the bandit happens to be a promising project

discojs/discojs-core/src/async_informant.ts
s314cy: this is here because the `update` method is called once per communication round, meaning possibly multiple times during a single training round, but the stats are over training rounds

but yes the code is unclear 😁

discojs/discojs-core/src/aggregator/base.ts
```diff
@@ -0,0 +1,239 @@
...
+export abstract class Base<T> {
+  /**
+   * Contains the ids of all active nodes, i.e. members of the aggregation group at
+   * a given round. It is a subset of all the nodes available in the network.
+   */
+  protected _nodes: Set<client.NodeID>
+  /**
+   * Contains the contributions received from active nodes, accessible
+   * by node id. It defines the effective aggregation group, which is
+   * possibly a subset of all active nodes, depending on the aggregation
```
martinjaggi: subset of all network nodes. but not subset of aggr group, right

discojs/discojs-core/src/aggregator/base.ts
```diff
@@ -0,0 +1,239 @@
...
+  /**
+   * Emits the aggregation event whenever an aggregation step is performed.
+   * Triggers the resolve of the result promise and the preparation for the
+   * next aggregation round.
+   */
+  private readonly eventEmitter: EventEmitter
+
+  protected informant?: AsyncInformant<T>
+  /**
+   * The result promise which, on resolve, will contain the current aggregation result.
```
martinjaggi, line 34: want to say if it's a model (params) or model difference, or could be either?

discojs/discojs-core/src/aggregator/base.ts
```diff
@@ -0,0 +1,239 @@
...
+        console.log(`Buffer is full. Aggregating weights for round (${this.communicationRound}, ${this.round})\n`)
+    }
+  }
+
+  setModel (model: tf.LayersModel): void {
+    this._model = model
+  }
+
+  /**
+   * Adds a node's id to the set of "nodes". A node represents an active neighbor
```
martinjaggi: to the aggregation group

discojs/discojs-core/src/aggregator/base.ts
```diff
@@ -0,0 +1,239 @@
...
+  setNodes (nodeIds: Set<client.NodeID>): void {
+    this._nodes = nodeIds
+  }
+
+  resetNodes (): void {
+    this._nodes = Set()
+  }
+
+  /**
+   * Sets the aggregator's round. To be used whenever the aggregator is out of sync
```
martinjaggi: round number

discojs/discojs-core/src/aggregator/base.ts
```diff
@@ -0,0 +1,239 @@
...
+
+  /**
+   * The aggregator's round.
+   */
+  get round (): number {
+    return this._round
+  }
+
+  /**
+   * The aggregator's current size, defined by its amount of contributions.
```
martinjaggi, line 221: good. maybe can say this is the size of the aggregation group, even if a bit redundant

discojs/discojs-core/src/aggregator/get.ts
```diff
@@ -0,0 +1,27 @@
...
+  switch (task.trainingInformation.aggregator) {
+    case AggregatorChoice.MEAN:
+      return new aggregator.MeanAggregator(task)
+    case AggregatorChoice.ROBUST:
+      throw new Error('not implemented')
+    case AggregatorChoice.BANDIT:
+      throw new Error('not implemented')
+    case AggregatorChoice.SECURE:
+      if (task.trainingInformation.scheme !== 'decentralized') {
+        throw error
```
martinjaggi: error message that we only support secagg for decentralized currently?

discojs/discojs-core/src/aggregator/base.ts
martinjaggi: oh i see for the buffer case this is growing. there when we reach the desired size (best to call that the aggr group size), then the aggregation is triggered

discojs/discojs-core/src/aggregator/base.ts
```diff
@@ -0,0 +1,239 @@
...
+  }
+
+  /**
+   * Logs useful messages during the various aggregation steps.
+   * @param step The aggregation step
+   * @param from The node which triggered the logging message
+   */
+  log (step: AggregationStep, from?: client.NodeID): void {
+    switch (step) {
+      case AggregationStep.ADD:
```
martinjaggi, line 98: add might be maybe misinterpreted as addition potentially, but here you mean more like `save` or `register`?

discojs/discojs-core/src/aggregator/base.ts
```diff
@@ -0,0 +1,239 @@
...
+        console.log(`> Adding contribution from node ${from ?? '"unknown"'} for round (${this.communicationRound}, ${this.round})`)
+        return
+      case AggregationStep.UPDATE:
```
martinjaggi, line 101: update meaning?

martinjaggi (changes requested):
amazing work, and excellent tests added also. thanks a lot. only minor comments above

discojs/discojs-core/src/aggregator/base.ts
s314cy: it can be anything really, since the class is generic, but in the case of the subclasses, it can either be the model weights or model weights difference

currently, I'm pretty sure the weights passed to the client are the entire model, which will go back to model difference in the future polishing PR(s): write docs, support a directed communication graph, re-include DP (includes model diff & clipping), re-write byzantine-robustness, etc.

discojs/discojs-core/src/aggregator/base.ts
s314cy: in this case, it is rather the set of active nodes, by opposition to the aggregation group defined by the contributing nodes (subset of active nodes)

I'll make sure that the "active nodes" and "aggregation groups" keywords are used consistently throughout the docs

discojs/discojs-core/src/aggregator/base.ts
s314cy: this docstring is outdated so I updated it in #580 (and it's clearer too)

discojs/discojs-core/src/aggregator/base.ts
s314cy: it corresponds to the aggregator's `add` method and will have a dedicated docstring in #580 so it'll make more sense then :)

discojs/discojs-core/src/aggregator/base.ts
s314cy: it corresponds to a node overriding its previous contribution with a new one, and will be clearer in #580 with its dedicated docstring

martinjaggi (approved):

PR #580, opened 2023-06-29
Add API docs for disco

s314cy:
### Description

Provides discojs (`discojs-{core,node,web}`) with centralized docs, under the form of HTML pages accessible at some path from `epfml.github.io/disco`. To do so, we use the [typedoc](https://typedoc.org) pkg which generates HTML API docs based on code documentation.

### Additions & Changes
- [x] add and config the `typedoc` pkg
- [x] add code documentation
- [ ] ensure the docs tree provides the right tree structure (done either by adding typedoc configs or by slightly modifying the discojs exports tree)
- [ ] expose the HTML docs on a path from `epfml.github.io/disco`


s314cy:
While the HTML docs can be built using `npm run docs` within any Disco subproject (`core`, `web`, `node`), exposure to the web is not a priority and thus is left for a future PR.

Issue #607, opened 2024-01-11
DISCO training fails silently


JulienVig:
**Clicking on the "Train" button in the pre-defined tasks of the DISCO frontend, both in local and on the [website](https://epfml.github.io/disco/#/) doesn't start the training.**

The error is only triggered when using the web client, local tests all pass.

1. Go to any pre-defined tasks such as the [MNIST task](https://epfml.github.io/disco/#/mnist), either on the website or on your local instance running the `develop` branch
2. Click on 'NEXT' two times until reaching the training step. (Uploading data doesn't change the outcome)
3. Click on 'Train Alone' or 'Train collaboratively'
4. Inspect element and check the console logs:
```
TypeError: this.datasetBuilder.isBuilt is not a function
    startTraining index.js:95
    Lc index.js:7
    Lc index.js:3
    startTraining index.js:93
    0 templateLoader.js:29
    d runtime-core.esm-bundler.js:155
    p runtime-core.esm-bundler.js:164
    t runtime-dom.esm-bundler.js:369
```

**Expected behavior**
Clicking on the Train button should start the training successfully or return an error message to the user specifying there is a problem, such message could be: "You didn't uplaod any data" or "The service is currently unavailable".

**Screenshots**
<img width="1094" alt="Screen Shot 2024-01-11 at 16 43 50" src="https://github.com/epfml/disco/assets/33122365/b76ca7d6-4fda-480a-b2ec-757f7d7aba34">

**Desktop (please complete the following information):**
 - MacOS
 - Firefox

JulienVig:
Edit: the error arises both on the website and on a local web-client running the `develop`branch

JulienVig:
Update: uploading data before training triggers the same error

JulienVig:
[This commit](https://github.com/epfml/disco/commit/6e85f6f1f8b772e15c5e03cb5b55ef4030db1894#diff-04761da9988636b38bf48b80fb0fcab0a5b47c7bba0254dc48233159441c8613) changes the method `dataset_builder.isBuilt()` into the getter function `get built()` without updating the frontend call to `datasetBuilder.isBuilt()` in `Trainer.vue`.
Getter function calls don't use parentheses, so I changed the call to `datasetBuilder.built` in `Trainer.vue. I found the same mistake for `datasetBuilder.size`, we should be on the lookout for others.

PR #608, opened 2024-01-11
Refactoring created mismatched function calls

JulienVig:
After [this commit](https://github.com/epfml/disco/commit/6e85f6f1f8b772e15c5e03cb5b55ef4030db1894#diff-04761da9988636b38bf48b80fb0fcab0a5b47c7bba0254dc48233159441c8613) changed the method `isBuilt()` to be the getter `get built()`, the front end was raising the error `datasetBuilder.isBuilt() is not a function`.

I changed the getter call from `datasetBuilder.isBuilt()` to `datasetBuilder.built` ([getters are called without parentheses](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get)

Same thing for `get size()`

### DIFF
--- a/web-client/src/components/testing/Tester.vue
+++ b/web-client/src/components/testing/Tester.vue
@@ -335,7 +335,7 @@ async function predictUsingModel (): Promise<void> {
 }
 
 async function assessModel (): Promise<void> {
-  if (props.datasetBuilder?.size() === 0) {
+  if (props.datasetBuilder?.size === 0) {
     return toaster.error('No file was given')
   }
 
--- a/web-client/src/components/training/Trainer.vue
+++ b/web-client/src/components/training/Trainer.vue
@@ -141,7 +141,7 @@ export default defineComponent({
     async startTraining (distributedTraining: boolean): Promise<void> {
       this.distributedTraining = distributedTraining
 
-      if (!this.datasetBuilder.isBuilt()) {
+      if (!this.datasetBuilder.built) {
         try {
           this.dataset = await this.datasetBuilder.build()
         } catch (e) {
@@ -157,7 +157,7 @@ export default defineComponent({
         await this.disco.fit(this.dataset)
         this.startedTraining = false
       } catch (e) {
-        this.$toast.error('An error occured during training')
+        this.$toast.error('An error occurred during training')
         console.error(e)
         this.cleanState()
       }


EduardDurech:
Generating Documentation...

web-client/src/components/testing/Tester.vue
```diff
@@ -335,7 +335,7 @@ async function predictUsingModel (): Promise<void> {
 }
 
 async function assessModel (): Promise<void> {
```
EduardDurech, line 337: **Suggestion:** Proposed documentation
```suggestion
/**
 * This function assesses the model. It first checks if the size of the datasetBuilder is zero.
 * If it is, it returns an error. Otherwise, it gets the validator and assigns it to the validator value.
 */
async function assessModel (): Promise<void> {
```

web-client/src/components/training/Trainer.vue
```diff
@@ -141,7 +141,7 @@ export default defineComponent({
     async startTraining (distributedTraining: boolean): Promise<void> {
```
EduardDurech, line 141: **Suggestion:** Proposed documentation
```suggestion
    /**
     * This method starts the training process. It first checks if the datasetBuilder is built.
     * If it is not, it tries to build the dataset. If an error occurs during this process, it cleans the state.
     * If the dataset is successfully built, it tries to fit the dataset using the disco object.
     * If an error occurs during this process, it also cleans the state.
     */
    async startTraining (distributedTraining: boolean): Promise<void> {
```

EduardDurech:
I'm just testing Codium for Doc creation, let me know if the bots responses were accurate above

martinjaggi:
fixes issue #607

JulienVig:
> I'm just testing Codium for Doc creation, let me know if the bots responses were accurate above

The doc is accurate but IMO it doesn't add much, it's a very literal description of the "if/else"'s and probably doesn't help someone with the basic programming skills to better understand what's happening. It's better than nothing but I would prefer a higher-level description that addresses potential confusions and design choices (and potentially human if need be).

martinjaggi:
want to mark it as ready to review, and merge?
 
(if possible i would avoid big restructurings on the docs side for now and aim for the minimal changes needed to get things working again, and to clarify some things in the markdown docs if it helps new users getting started?)

JulienVig:
Yes, so far I've only been updating markdown files and fixing typos in the code as I see them

JulienVig:
@EduardDurech is busy this week, @martinjaggi do you have time to review it?

EduardDurech (approved):

EduardDurech:
Small commit so able to check

JulienVig:
Fixed #607

PR #614, opened 2024-01-30
Update DISCO documentation

### LINKED ISSUE #549: task documentation outdated for custom tasks
in `docs/TASK.md`, parts of the custom task creation is already outdated:

such as `make sure to include its reference in the discojs/discojs-core/src/tasks/index.ts file as follows:`
it seems to me nothing has to be changed anymore in that file, right?

would be good if someone quickly goes through if things work to add a new task like described there

### LINKED ISSUE #606: Update and improve documentation
Most markdown documentation is outdated, for example with broken links, outdated commands or different naming conventions (DISCO, DisCo, Disco, Disco 🔮), reference to outdated github project and milestones.

Additionally, I think the documentation can be improved by fixing typos, creating a documentation table of contents, adding further installation instructions and explanations.

JulienVig:
The DISCO documentation has become outdated over time and the recent refactor. I also improve the instructions from a newcomer's point of view and included information from the onboarding video as well as what I had to figure out by myself. Notably, I created an installation script `install.sh` to install and build the different project modules.

To review:
- [x]  `README.md`
- [x]  `DEV.md`
- [x]  `docs/ONBOARDING.md`
- [x]  `docs/CONTRIBUTING.md`
- [x]  `docs/DISCOJS.md`
- [x]  ~`docs/ARCHITECTURE.md`~ → dispatched content into relevant files
- [x]  `docs/PRIVACY.md`
- [x]  ~`web-client/src/components/README.md`~ → `docs/VUEJS.md`
- [x]  `docs/TASK.md`
        - Did most of it, will come back to it when I need to implement a custom task
- [x]  `docs/FAQ.md`

Will address the following after other PRs are merged (fixing the CLI, integrating npm workspaces or Buns) as they may affect the doc instructions
- [ ]  `docs/node_example`
- [ ]  `web-client/README.md`
- [ ]  `server/README.md`
- [ ]  `cli/README.md`
- [ ]  `discojs/README.md`
- [ ]  `discojs/discojs-core/README.md`
- [ ]  `discojs/discojs-node/README.md`
- [ ]  `discojs/discojs-web/README.md`

Closes #549, Closes #606


web-client/package.json
```diff
@@ -12,7 +12,7 @@
     "serve17": "export NODE_OPTIONS=--openssl-legacy-provider && vue-cli-service serve"
   },
   "dependencies": {
-    "@epfml/discojs": "file:../discojs/discojs-web",
+    "@epfml/discojs": "^2.1.1",
```
JulienVig, line 15: Made this change according to what Hugo said in the onboarding video at 14:00, package.json should not be pointing to the local file.

DEV.md
```diff
@@ -13,44 +8,165 @@
...
+After installation, you should be able to run
+```
+nvm -v
+0.39.7 # my nvm version at the time
+```
+2. Install Node.js version 16
+```
+nvm install 16
+```
+You can now choose which Node.js version to use:
```
tharvik, line 58: is there a minimal node js version one need to use? if so, that'll be the right place to state it.

install.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
```
tharvik: it's not using any bash specifics

```suggestion
#!/bin/sh
```

DEV.md
```diff
@@ -13,44 +8,165 @@
...
-  - [discojs-node](./discojs/discojs-node/README.md) contains the Node.js implementatino of the Disco.js library. It is available on NPM as `@epfml/discojs-node`.
-  - [discojs-core](./discojs/discojs-core/README.md) contains code common to `discojs-web` and `discojs-node`. Little source code changes between the two implementations, hence the need for a "core" directory.
-- [server](./server/README.md) contains the Node.js server for Disco, enabling the decentralized (with peer-to-peer coordination) and federated learning of Disco.js nodes. It uses `discojs-node` under the hood.
-- [web-client](./web-client/README.md) contains a web client to interact with Disco.js and the Disco server via a user-friendly UI. This basically emulates a Disco.js node in the network. It uses `discojs-web` as backend.
-- [cli](./cli/README.md) contains the CLI tool for Disco.js. Just like `web-client`, it emulates a Disco.js node in the network. Furthermore, it can actually emulate multiple nodes and even manages its own Disco server instance. It uses `discojs-node` under the hood.
-
-## Example
+4. Run the installation script
+```
+sh install.sh
```
tharvik: no need to actively run it with `sh`, the [shebang](https://en.wikipedia.org/wiki/Shebang_(Unix)) takes care of that

```suggestion
./install.sh
```

install.sh
```diff
@@ -0,0 +1,30 @@
...
+set -e # Exit script on error
+DIR="$( cd "$( dirname "$0" )" ; pwd -P )" # Fetch current directory
+
+echo '\n>>> Installing disco.js dependencies and building the library\n'
```
tharvik: `echo` varies greatly between shells. for example, using `bash` actually prints "\n>>> Installing [...] library\n" instead of line returns as wanted, and `dash` actually prints it in three lines. `printf` is better if you want to interpret "\n" as line returns consistently. or you can use multiple echo.

install.sh
```diff
@@ -0,0 +1,30 @@
...
+
+echo '\n>>> Installing the CLI dependencies\n'
+cd $DIR/cli
+npm ci
+
+echo '\n>>> Installing ts-node\n'
+npm install -g ts-node
+
+echo '\n>>> Downloading and extracting sample training data\n'
+sh get_training_data
```
tharvik: same idea as before, you can add a shebang to `get_training_data` instead.

DEV.md
```diff
@@ -13,44 +8,165 @@
...
+```
 
-A full -- self-contained -- examples of the Disco.js API running two federated users can be found [here](./docs/node_example). This runs on Node.js outside of any browser, using the `@epfml/discojs-node` NPM package and the `server` module of this repo. A Disco server is run from the script itself and the data is already available in the repo.
+<details>
+  <summary><b>What does <code>install.sh</code> do?</b></summary>
+  </br>
+  The installation script installs the dependencies required by the different parts of the project, which are described in the Structure section.
```
tharvik: worth a link to the named section (or don't worry about it if moving to a code)

DEV.md
```diff
@@ -13,44 +8,165 @@
...
 
-The following command lines will install the required dependencies, build Disco.js and start the Disco server (on localhost:8080) and the Disco web client (on localhost:8081). We recommend using (and assume you are) [nvm](https://github.com/nvm-sh/nvm) (the Node Version Manager).
+## Structure
 
+The DISCO project is composed of multiple parts. At the root level, there are four main folders: `discojs`, `server`, `web-client` and `cli`.
+
+- `discojs`, or Disco.js, is the JavaScript library that contains federated and decentralized learning logic. The library allows to train and use machine learning models in a distributed fashion. The library itself is composed of the `disco-node` and `disco-web` modules, both of them extending the platform-agnostic code in `disco-core`. In other words, `disco-core` contains most of the implementation but can't be used by itself, while `disco-web` and `disco-node` allow using `disco-core` via different technologies. To some extents, you can think of `disco-core` as an abstract class extended by `disco-web`and `disco-node`.
```
tharvik:
```suggestion
- `discojs`, or Disco.js, is the TypeScript library that contains federated and decentralized learning logic. The library allows to train and use machine learning models in a distributed fashion. The library itself is composed of the `disco-node` and `disco-web` modules, both of them extending the platform-agnostic code in `disco-core`. In other words, `disco-core` contains most of the implementation but can't be used by itself, while `disco-web` and `disco-node` allow using `disco-core` via different architectures. To some extents, you can think of `disco-core` as an abstract class extended by `disco-web`and `disco-node`.
```

DEV.md
```diff
@@ -13,44 +8,165 @@
...
+  Install the CLI dependencies:
+  ```
+  cd ../cli
+  npm ci
+  ```
+  Install `ts-node` globally. `ts-node` lets us compile and run TypeScript code in a single command.
+  ```
+  cd ..
+  npm install -g ts-node
+  ```
```
tharvik: installing ts-node is not really required for running disco, is it? it's more of a dev tool. consider moving it out of the install script and add it as a quick note.

DEV.md
```diff
@@ -13,44 +8,165 @@
...
-
-## Example
+4. Run the installation script
+```
+sh install.sh
+```
 
-A full -- self-contained -- examples of the Disco.js API running two federated users can be found [here](./docs/node_example). This runs on Node.js outside of any browser, using the `@epfml/discojs-node` NPM package and the `server` module of this repo. A Disco server is run from the script itself and the data is already available in the repo.
+<details>
+  <summary><b>What does <code>install.sh</code> do?</b></summary>
```
tharvik: details are so fancy, I love it! but consider moving this section as comments in the install script as to easily keep code and explanation in sync

JulienVig:
Read to merge this PR, I will update the remaining docs after other PRs which are susceptible to change the doc content are merged.

discojs/test.sh
```diff
@@ -1,5 +1,20 @@
-DIR="$( cd "$( dirname "$0" )" ; pwd -P )"
+set -e # Exit script on error
```
tharvik, line 2: similar comments as for `install.sh`

docs/TASK.md
```diff
@@ -307,7 +285,7 @@ export interface TrainingInformation {
   // batchSize: batch size of training data
```
tharvik: we are using `typedoc` in a few places, is the doc pushed somewhere? can we instead reference the documented `TrainingInformation` to avoid out-of-sync doc & code?

tharvik (commented):
impressive documentation, thanks! I added a few comments, some not really related to the PR but for my own understanding

web-client/package.json
tharvik: yep, it won't matter anymore with the use of #617 as `npm` takes care of that for us

DEV.md
JulienVig: I'm not well versed in shell scripts but I think using sh ensures we can run the script without having to chmod +x, am I wrong? At least on my mac I can't run ./install.sh directly

docs/TASK.md
JulienVig: I have no idea about typedoc I didn't read anything about it so far. And yes good idea I'll point to TrainingInformation.

DEV.md
JulienVig: I don't know, however I saw there were different npm run scripts for node 17

DEV.md
JulienVig: Unfortunately, I think you're right. I got blinded by my desire for fancy details. I'm moving it to a new Section.

DEV.md
JulienVig: That makes me realize that there is no clear distinction between installation for dev users and contributors. Similarly, the script should probably not link local modules for dev users as it currently does. Should I change that now or will the npm --workspaces take care of it @tharvik ?

JulienVig: I'm adding contributor specific instructions in `ONBOARDING.md` but I'm leaving the npm links in `install.sh` for now

DEV.md
tharvik: alright, we'll see if/when it breaks then

DEV.md
tharvik: indeed, only executable files can use a shebang; git actually tracks file modes so you can `chmod +x` and that'll work for fresh install also
in our case, it is not really relevant that it is, or not, a shell script that we run, it's only a way to install deps. I think having it as executable makes it more obvious that it is meant to be run

DEV.md
tharvik: the links won't be needed when using workspaces as it'll be automatically created using `npm i`. I'll update the doc as necessary via #617

install.sh
EduardDurech: Expanding on @tharvik's comment, to make sure echo respects escape characters use `echo - e`, but agree using, `printf` is better

EduardDurech (dismissed):

install.sh
EduardDurech: Sorry I'm on mobile and submitted the comment wrong, just ignore the "approved changes"

tharvik (approved):

PR #616, opened 2024-02-01
Fix training related failures

### LINKED ISSUE #610: Train Collaboratively raises error during training
From a localhost web-client, uploading titanic data and clicking on Train Collaboratively raises recurrent errors (inspect page and check the console logs)
```
...
> Adding contribution from node 410578ec-1b93-425c-ab2f-56e33800b657 for round (0, 2)
********************************************************************************
Buffer is full. Aggregating weights for round (0, 2)

Error: timeout
    at Timeout._onTimeout (/Users/julien/Documents/DISCO/repo/discojs/discojs-node/dist/core/client/utils.js:13:64)
    at listOnTimeout (node:internal/timers:559:17)
    at processTimers (node:internal/timers:502:7)` 
...
```

Additionally, the training doesn't seem to ever stop.

### LINKED ISSUE #612: Fix local test failure
Running server tests fails with a timeout on `1) end-to-end federated  "two cifar10 users reach consensus"`, the user accuracies don't converge. (At the time of writing, on the `develop` branch commit `c0f27cf`)

### LINKED ISSUE #615: Training on Titanic has a NaN loss
When training on the titanic task (and potentially other tasks), the loss is NaN at every epoch. The output of `model.predict` is only composed of NaN values. Therefore are no accuracy improvements throughout the epochs.
This happens for every training format: from the browser UI, local, federated etc.

JulienVig:
Fixes multiple relatively small bugs happening when training with DISCO.

* Fixes #612
One local test was failing with a timeout on my laptop. It happens that the test timeout was too short for my humble machine to finish the test. I simply increased the timeout value.

* Fixes #615
The training Loss was NaN when training the Titanic task. There were two issues:
  1. the tabular preprocessing was empty after a preprocessing (cf. issue for more details). I added a placeholder preprocessing to handle missing values and ensure there is no NaN in the inputs
  2. the default learning rate was too high for the small amount of data and led the weights to diverge. Decreased it from 0.01 to 0.001.
Before: training accuracy constant at 0.61 and NaN Loss, now the training accuracy increases throughout the epochs (as well as the validation) while the loss decreases monotonically, both from a Node.js script and from the web-client.

* Fixes #610 
See issue for detailed description of bug and fix. While the client used to send the weights and then request the new weights separately, the server now automatically sends back the new weights after receiving the local contribution. This change saves 1 RTT and importantly, ensures that events happen in the right order, which was previously causing timeouts.
Now training collaboratively doesn't timeout anymore. The training accuracy now increases monotonically to 70% while it was previously staying constant at 61% with NaN loss.


discojs/discojs-core/src/client/federated/base.ts
```diff
@@ -19,14 +19,6 @@ export class Base extends Client {
...
    */
   public static readonly SERVER_NODE_ID = 'federated-server-node-id'
-  /**
-   * Most recent server-fetched round.
-   */
-  private serverRound?: number
-  /**
-   * Most recent server-fetched aggregated result.
-   */
-  private serverResult?: WeightsContainer
```
JulienVig, line 29: Only used once in a single method, made them local variables rather than attributes

discojs/discojs-core/src/client/federated/base.ts
```diff
@@ -226,13 +214,11 @@ export class Base extends Client {
       throw new Error('local aggregation result was not set')
     }
 
-    // Send our contribution to the server
-    await this.sendPayload(this.aggregator.makePayloads(weights).first())
-    // Fetch the server result
-    await this.receiveResult()
+    // Send our local contribution to the server
+    // and receive the most recent weights as an answer to our contribution
+    const serverResult = await this.sendPayloadAndReceiveResult(this.aggregator.makePayloads(weights).first())
```
JulienVig, line 219: Since now the server directly returns the new weights after receiving the client's weights, I moved both the sending and receiving in one method to ensure that both occur right after the other (and not miss the new weights)

docs/node_example/start_server.ts
JulienVig, line 1: Currently doesn't work after the last refactor, removed it for now and launch a server instance manually

server/src/router/federated/server.ts
JulienVig, line 1: Moved logic in private functions outside the `handle` method for modularity.

discojs/discojs-core/src/dataset/data/tabular_data.ts
```diff
@@ -21,7 +21,8 @@ export class TabularData extends Data {
     try {
       await dataset.iterator()
     } catch (e) {
-      throw new Error('Data input format is not compatible with the chosen task')
+      console.error('Data input format is not compatible with the chosen task.')
+      throw (e)
```
tharvik, line 25: good idea to actuall keep the error we got; we can now even [wrap the received error in the thrown `Error` via cause](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause). now it can serve as context around the received error (as in most modern languages). also, avoiding console.* let users of the library choose how/when to print errors (such as a line in a webapp instead of the console). 

```suggestion
      throw new Error('Data input format is not compatible with the chosen task.', { cause: e })
```

docs/node_example/tsconfig.json
```diff
@@ -14,7 +14,7 @@
 
     "declaration": true,
 
-    "typeRoots": ["node_modules/@types", "discojs-core/types"]
+    "typeRoots": ["node_modules/@types", "../../discojs/discojs-core/types"]
```
tharvik, line 17: good catch; we should add a GA pass for the examples

discojs/discojs-core/src/client/federated/base.ts
```diff
@@ -92,41 +84,37 @@ export class Base extends Client {
...
-  async receiveResult (): Promise<void> {
-    this.serverRound = undefined
-    this.serverResult = undefined
-
-    const msg: messages.MessageBase = {
-      type: type.ReceiveServerPayload
-    }
-    this.server.send(msg)
-
+  async receiveResult (): Promise<WeightsContainer|undefined> {
```
tharvik: tell me if I'm wrong but I don't think it's used anywhere outside this class

```suggestion
  private async receiveResult (): Promise<WeightsContainer|undefined> {
```

discojs/discojs-core/src/dataset/data/preprocessing/tabular_preprocessing.ts
```diff
@@ -9,7 +10,25 @@ export enum TabularPreprocessing {
   Normalize
 }
 
+interface TabularEntry extends tf.TensorContainerObject {
```
tharvik, line 13: hoo cool, I didn't knew about this TensorContainerObject, that's useful

tharvik (approved):
small non-blocking comments, merge at will :rocket:

discojs/discojs-core/src/dataset/data/tabular_data.ts
JulienVig: Ah great I googled a bunch for that and didn't find this syntax thanks

docs/node_example/tsconfig.json
JulienVig: What's a GA pass?

discojs/discojs-core/src/client/federated/base.ts
JulienVig: You're right! I left it as it was previously without thinking

discojs/discojs-core/src/dataset/data/tabular_data.ts
JulienVig: Okay so lib ES2022 is needed to compile an error with cause and I couldn't do it without raising new issues. I'll let you change it if you want as I'm not sure of what upgrading to ES2022 entails.

docs/node_example/tsconfig.json
tharvik: a github action step to actually try to build/run the examples, as to automatically catch theses out-of-sync issues

discojs/discojs-core/src/dataset/data/tabular_data.ts
tharvik: ho, I didn't check which ECMA version it was definied in, too bad; yeah, I'll try to update it later then, thanks for point this out.

PR #617, opened 2024-02-02
*: use npm workspaces

tharvik:
* no need for `npm link` anymore
* allow to build all projects via `npm -ws run build` (or `test`, or any script really)
* allow to easily import packages (simply import @epfml/disco-core) by using a common node_modules for shared deps
* watch files changes and run npm script in most packages, for eg `npm -w ./discojs/discojs-core run watch build`
* change much of the github action to better reflect the stages dependencies (we can't lint discojs-core without build discojs-node for example)


JulienVig:
Amazing! Can you update `DEV.md` with an updated install.sh description?
Edit: never mind, I need to merge the documentation branch beforehand

tharvik:
> Amazing! Can you update `DEV.md` with an updated install.sh description?

Thanks, yep, will do!

I will also first discuss it with @peacefulotter as it probably conflicts with the bun integration PR.

tharvik:
I've also added a way to watch for changes and build/lint/test accordingly. One can use it via `npm run watch build` in every pkg.

discojs/publish.ts
JulienVig, line 1: Why is `publish.ts` not useful anymore?

docs/ONBOARDING.md
```diff
@@ -39,19 +38,18 @@ ts-node your_script.ts
...
+npm -w ./discojs run build
 ```
 
-* When cloning the repo, the server and the web-client points to the [@epfml/disco-node](https://www.npmjs.com/package/@epfml/discojs-node) and the [@epfml/disco-web](https://www.npmjs.com/package/@epfml/discojs) packages respectively published on the `npm` remote repository. In a development environment, we want to use the local implementations in the `discojs/discojs-node` and `discojs/discojs-web` folders for changes to take effect immediately. To do so, we need to link the local folders as the actual dependencies:
-```
-cd server 
-npm link ../discojs/discojs-node
-cd ../web-client 
-npm link ../discojs/discojs-web
+If you want to automate the building phase, you can use the watch script in every project that run the given script when changes are detected.
```
JulienVig:
```suggestion
If you want to automate the building phase, you can use the `watch` script to rebuild a module whenever changes are detected.
```

docs/ONBOARDING.md
```diff
@@ -39,19 +38,18 @@ ts-node your_script.ts
...
+As each of theses calls are non-terminating (theses are watching indefinitely), you'll need to run each in a different terminal.
```
JulienVig:
```suggestion
As each of those calls are non-terminating (they're watching indefinitely), you'll need to run each of them in a different terminal.
```

JulienVig (approved):
Very nice! This is going to make things so much smoother, thanks! I left a few non-blocking questions and suggestions.

discojs/publish.ts
tharvik: it was to recreate the `package.json` files of discojs-{node,web}, but that's not needed anymore as discojs-core is now a real dependency, not bundled in both. note that, despite its name, it doesn't actually publish the packages.
hopefully, we will simply need to use `npm publish` to release a version to NPM.

PR #620, opened 2024-02-03
Added a NextJS frontend for training GPT on wikitext and executing some web text dataset related tests

peacefulotter:
This PR adds another frontend to Disco, this time written in NextJS. The reason is simply because I do not know Vue at all while I am quite familiar with NextJS.
In any case, I believe the Vue frontend should in turn entirely move to NextJS as it appears to be much better (and modern) in general, but this is more of a personal opinion.


JulienVig:
Closing as there is no foreseeable need to support another frontend framework and the PR still requires some work before being able to be merged

peacefulotter:
Noooo NextJS >>> 
Are there plans to change the frontend framework or no need in the end?

JulienVig:
This is currently no plan to integrate another one. Migrating away from vusjs to nextjs would require re-implementing all the UI features again which is definitely not a priority, though you're welcome to work on this if you want!

peacefulotter:
If you have some ideas on the future frontend components and use cases, I could try to see if it's worth migrating to NextJS and even contribute for sure. Right now, the UI does not look too professional imho but I don't know if this will ever be an important criteria

JulienVig:
Having a good-looking frontend is important! I would really appreciate your feedback on what's not looking professional right now.

PR #626, opened 2024-02-13
TMP trigger CI

tharvik:
I'm having issue w/ wrtc segfaulting (see #617), trying out with smaller commits


tharvik:
hum, it seems to be working again, I'm unable to trigger the issue anymore on #617; well, let's hope that it won't happen again.

PR #630, opened 2024-02-13
Fix flaky titanic validator test

### LINKED ISSUE #629: Titanic validator test is flaky
The `validator.spec.ts` test that validates the titanic task is flaky, it requires a minimum accuracy of 0.5 and fails around half of the time.

How to reproduce: 
1. build discojs: `cd discojs; npm run build`
2. launch a server instance `cd server; npm run dev`
3. run the titanic test: `cd discojs/discojs-core; npx mocha -g "works for titanic"`

JulienVig:
Fixes #629 

When uncommenting the titanic validator test (which was previously commented because the loss was NaN), I increased the accuracy minimum without noticing that the accuracy was tested at initialization, i.e., without any training.
Therefore, I lower the validator accuracy expectations (from 0.5 to 0.3) but I add assertions on minimum training and validation accuracy (both > 0.6) in the end-to-end test.

### DIFF
--- a/discojs/discojs-core/src/validation/validator.spec.ts
+++ b/discojs/discojs-core/src/validation/validator.spec.ts
@@ -32,9 +32,9 @@ describe('validator', () => {
 
     const data: data.Data = (await new node.data.NodeImageLoader(simplefaceMock)
       .loadAll(files.flat(), { labels })).train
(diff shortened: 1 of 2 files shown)


tharvik (approved):
hoo nice, you found out why, thanks!

Issue #638, opened 2024-02-28
don't reexport discojs-core and tfjs


tharvik:
currently, discojs-node & discojs-web are reexporting discojs-core, which looks weird to me: users are mixing imports from both packages & commonization of deps is non-trivial.
having every project importing directly from discojs-core and only platform specifics from discojs-{node,web} makes more sense IMO.

tfjs is also reexported everywhere and that might produce subtle bugs when dealing with mixed versions.

removing this tight coupling would help both dev and users of disco.

PR #636, opened 2024-02-22
Update doc according npm workspace

JulienVig:
* Fixes #635 
  Update each module's instruction now that npm workspaces has been integrated

I reorganized the documentation with 3 different types of DISCO users in mind. What they should be able to do are concentric subsets:
1. non-technical users → want to use the UI to train and create their own custom tasks without coding
2. technical users → additionally code in Node.js to train and create their own custom tasks with discojs-node and discojs-server
3. contributors → additionally need to be able to build, test, modify and learn in-depth information

As such, I organized the doc by following these successive levels:
1. README.md for everyone and especially non-technical users
2. DEV.md for technical users and contributors, it now contains instructions for the different ways to use DISCO
3. CONTRIBUTING.md for contributors, with build, test, coding conventions and in-depth information needed to modify the code base.

Additionally:
- Removed redundant module readme (discojs, disco-core/node/web, web-client) as instructions are now centralized in DEV.md and CONTRIBUTING.md
- Merged ONBOARDING.md into CONTRIBUTING.md and deleted the former. I didn't find the difference between the two files self-explanatory enough and contributors have to go through both of them anyway so I merged the two into one file.
- Created a new example for adding a custom task to the server in `docs/node_example`, which I renamed `docs/examples` with the objective of adding further examples in the future.

Out of scope: #639 Adding custom tasks is currently very convoluted and needs a rework to make it easier for external users. Information on how to add custom tasks is not clear and not accessible enough for any of the three user categories I mentioned.

* Also fixes #609 
  I expect it got fixed by the previous PRs.

PR #640, opened 2024-02-28
don't reexport tfjs/discojs-core

tharvik:
implementation of #638


JulienVig:
Very nice!! 
Once the PR #636 is merged can you remove the doc information on importing different tfjs versions ? 
It's in docs/CONTRIBUTING.md, Section #discojs-core-discojs-node-and-discojs-web:

> Note that, if you end up making calls to the Tensorflow.js API, you must import it from the root index. This is to ensure the right version of TF.js is loaded (depending on the compilation dist/), and only once. The only exception occurs in unit tests, which should import TF.js from the (local) @epfml/discojs-node, since those run on Node.js.

JulienVig (approved):
Nothing to say, everything seems to work well and everything seems much better

PR #643, opened 2024-02-29
add fondations for LLM

tharvik:
merging LLM (from #619) requires some reshaping of the code, here goes some changes splitted from the original PR.

* add a `Model` abstracting away tensorflow specifics
  * still incomplete as it input/output `tf.Tensors` for now, and one can extract the `tf.LayersModel`
  * remove `Task.training.modelCompileData` as it's now contained inside the `models.TFJS`
* add typesafe event emitter to safely register/call callbacks
* rename `Task.taskID` to `Task.id`
* while at it: drop unused training_functions, update axios, use async file read


martinjaggi (approved):

cli/src/cli.ts
```diff
@@ -1,41 +1,45 @@
...
 
-const infoText = `\nStarted federated training of ${TASK.taskID}`
+const infoText = `\nStarted federated training of ${TASK.id}`
 console.log(infoText)
 
 console.log({ args })
 
-async function runUser (task: Task, url: URL, data: data.DataSplit): Promise<TrainerLog> {
+async function runUser (task: Task, server: Server, data: data.DataSplit): Promise<TrainerLog> {
+  const client = await getClient(clients.federated.FederatedClient, server, task, new aggregators.MeanAggregator(TASK))
```
JulienVig, line 21: What's the point of passing a whole client to Disco rather than the url?

server/src/tasks.ts
```diff
@@ -93,34 +93,26 @@ export class TasksAndModels {
...
 
+    let tfModel: Model
     if (model === undefined) {
       tfModel = await this.loadModelFromTask(task)
-    } else if (model instanceof tf.LayersModel) {
+    } else if (model instanceof Model) {
       tfModel = model
     } else if (model instanceof URL) {
-      tfModel = await tf.loadLayersModel(model.href)
+      tfModel = new models.TFJS(await tf.loadLayersModel(model.href))
```
JulienVig, line 110: I feel that the `models.TFJS` object is quite confusing, what does it represent? 
If it's only used here can we expose a model builder API rather than exposing the TFJS native constructor? Like `Model.fromURL(...)`

JulienVig: Similarly, exporting `TFJS` in `disojs-core/models/index.ts` feels weird. Is the object a Disco TF.js wrapper? What about renaming it TFjsAdapter after the adapter pattern, or something more self-explanatory?

JulienVig: Ah I think I understand better, TFJS is the TF.js implementation of the generic abstract Model class. What about renaming it TFjsModel? Or are you relying on the module name `models.TFJS`  to make it clear?

discojs/discojs-node/src/data/image_loader.spec.ts
```diff
@@ -16,24 +16,38 @@ const DIRS = {
...
 const cifar10Mock: Task = {
-  taskID: 'cifar10',
+  id: 'cifar10',
   displayInformation: {},
   trainingInformation: {
+    modelID: 'cifar10',
+    epochs: 1,
+    roundDuration: 1,
+    batchSize: 1,
+    dataType: 'tabular',
```
JulienVig, line 26: CIFAR10's `dataType` is probably 'image' rather than 'tabular'

discojs/discojs-node/src/data/image_loader.spec.ts
```diff
@@ -16,24 +16,38 @@ const DIRS = {
...
 const mnistMock: Task = {
-  taskID: 'mnist',
+  id: 'mnist',
   displayInformation: {},
   trainingInformation: {
+    modelID: 'mnist',
+    epochs: 1,
+    roundDuration: 1,
+    batchSize: 1,
+    dataType: 'tabular',
```
JulienVig, line 43: Same dataType should be 'image' for MNIST

web-client/src/components/task_creation_form/TaskForm.vue
```diff
@@ -254,14 +248,14 @@ const onSubmit = async (rawTask: any, { resetForm }): Promise<void> => {
 
   // format the flat form entries to a nested task object
   const specialSections = [
-    'generalInformation', 'modelCompileData', 'privacyParamters', 'modelFiles'
```
JulienVig, line 257: Why are you removing the compilation parameters?

JulienVig:
For what it's worth very good stuff! I only left comments for what I didn't understand but it's a very nice refactoring

cli/src/cli.ts
tharvik: getting the URL out of a server is non-trivial, as you can see from `getClient` impl. but now that you mention it, passing the server in full here disallow connecting to an existing server. I'll update it so that `startServer` actually returns the URL to connect to and get rid of `getClient`.

discojs/discojs-node/src/data/image_loader.spec.ts
tharvik: indeed, good catch!

discojs/discojs-node/src/data/image_loader.spec.ts
tharvik: double catch, thanks

server/src/tasks.ts
tharvik: > TFJS is the TF.js implementation of the generic abstract Model class.

indeed, the idea is to have multiples implementation of `Model`, one would be wrapping a tfjs model, one would be GPT specific, some other. that allow to keep the old `tf.LayersModel` contained and only providing generic accessor, to better support the coming up models.

> What about renaming it TFjsModel? Or are you relying on the module name `models.TFJS` to make it clear?

I find the indirect import (`models.TFJS`) more explicit, and it's volontarely more cumbersome to write as users should depend on `Model` and not on a specific implementation.

web-client/src/components/task_creation_form/TaskForm.vue
tharvik: it looked to me to be very model specific (and not fit for GPT for eg), I inlined it inside `getModel`. I don't see it as decoupled; and it is needed to compile the model with it to be able to train so expose a `Model.compile` which looked weird.

I'm trying to untangle this omninous global object that `Task` is, cutting fields, not depending on it where I can. I think that the model specific fields are not fit to be there at all, it should be only in `getModel`. but at some point, we will need to discuss a bit more on what we want from that (the no-code platform is an illusion and not useful IMO)

tharvik:
> For what it's worth very good stuff! I only left comments for what I didn't understand but it's a very nice refactoring

héhé, thanks :smile:

followup of fixes from your remarks #645

Issue #641, opened 2024-02-28
first LLM prototype


martinjaggi:
we already had the PRs but no issue yet, so adding that for completeness

objective: add a first prototype implementation based on the [GPT-tfjs](https://github.com/zemlyansky/gpt-tfjs) code adaption by @peacefulotter and @s314cy

related
https://github.com/peacefulotter/disco/tree/NAN-llm-s314cy

is addressed partially in those PRs it seems:
#619 and #600

Issue #632, opened 2024-02-14
upgrade to ES2022


tharvik:
currently, we are using EcmaScript 5 (ES5) which is quite old (released in 2009). as the JS ecosystem is moving forward, and quite fast theses years (one new per year), we need to move to a newer version. TypeScript allow uses to use some of theses features by transpiling to older ES via helpers, but it's only partial support, and using theses helpers instead of browser implementations do have size and runtime costs. here goes a list of features that will benefit the project
* async/await support & standard modules (ES6/ES2015)
* async iterator/generator (ES2018)
* `Array.flatMap` (ES2019)
* null coalesing & optional chaining (ES2020)
* `Error.cause` (ES2022)

~94% of browsers [do support ES2022 (except static field block init)](https://caniuse.com/?feats=mdn-javascript_builtins_array_at,mdn-javascript_builtins_regexp_hasindices,mdn-javascript_builtins_object_hasown,mdn-javascript_builtins_error_cause,mdn-javascript_operators_await_top_level,mdn-javascript_classes_private_class_fields,mdn-javascript_classes_private_class_methods,mdn-javascript_classes_static_class_fields), that looks wide enough support to me.

peacefulotter:
I know bun and a bunch of other packages are using esnext. I have heard that it was recommended to set it to esnext but I am not sure about its support coverage

PR #644, opened 2024-02-29
discojs-core/models: add gpt

tharvik:
add GPT model, tracked in #641

* add text loader from Hugo
* add "wikitext" task, the LLM reference
* add `models.GPT`, wrapping @peacefulotter code
* do *not* add real tokenization as #646 takes care of that
* add /datasets in place of /example_training_data
  * a script takes care of getting the data, it should be replaced by submodules & LFS, see #648

this is a prototype that is not being tested (as we need tokenization to get meaning out of it), only that the training of the model is reducing loss


discojs/discojs-core/src/models/gpt/index.ts
```diff
@@ -0,0 +1,136 @@
...
+
+      const xs = tf.buffer([batchSize, GPT.blockSize], 'int32')
+      const ys = tf.buffer([batchSize, GPT.blockSize, GPT.vocabSize], 'int32')
+
+      for (let i = 0; i < batchSize; i++) {
+        for (let j = 0; j < sampleSize; j++) {
+          const idx = (i * sampleSize + j) * 2
+          const low = buffer.get(idx)
+          const high = buffer.get(idx + 1)
+          const token = toUInt16(low, high)
```
JulienVig, line 92: Do you know how `toUInt16` works? I'm particularly confused by the arguments `low` and `high` which should be numbers from the function signature but `buffer` contains text sentences.
More importantly, it currently always return 0, so the training data is a matrix filled with 0.
@peacefulotter if you have any insight it would be welcomed!

discojs/discojs-core/src/models/gpt/index.ts
```diff
@@ -0,0 +1,136 @@
...
+    validationData?: Dataset,
+    epochs = 1,
+    tracker = new Sink()
+  ): AsyncGenerator<EpochLogs, void> {
+    for (let i = 0; i < epochs; i++) {
+      let logs: tf.Logs | undefined
+
+      await this.model.fitDataset(
+        this.convertCharDataset(trainingData), {
+          epochs: 1,
```
JulienVig, line 113: Also, do you know why we need to manually iterate over the epochs rather than using the `epochs` parameter?
When I remove the for-loop and use the `epochs` parameter, the training stops after the first epoch
(@peacefulotter)

discojs/discojs-core/src/models/gpt/index.ts
peacefulotter: The raw dataset is a text file which is tokenized into 2 bytes tokens saved in a binary file. When reading the tokenized file, chunks of bytes need to be converted back to tokens. This is what the toUint16 does: it groups 2 bytes (uint8) into a single number (uint16). If the output of toUint16 is always 0, the file you are reading from contains zeros.

JulienVig: Thanks for your answer! Could you send us your version of wikitext-103? The [link](https://github.com/peacefulotter/disco-experiments/blob/main/install-wikitext.sh) in your disco-experiments is broken.

discojs/discojs-core/src/models/gpt/index.ts
peacefulotter: I am not sure why training stops after one epoch. If I remember correctly, I had to bypass the batchSize parameter because the text loader creates the batch itself (without using gf dataset.batch(batchSize). This was done for improving performance. However, I can't see why it would have link with epochs, but it might be a start

discojs/discojs-core/src/models/gpt/index.ts
peacefulotter: What? This is really weird, the url is supposed to be permanent.. I will check tonight

JulienVig: I'm getting "access denied"

discojs/discojs-core/src/models/gpt/index.ts
martinjaggi: whichever is easiest...
in LLM terminology, one datapoint (=element of the batch) would usually be the whole sequence (of fixed context length, say 500, with padding the non-used tokens). batch-sizes larger than 1 would be nice to support, but it has to be its own tensor dimension independent of the sequence length

see also https://github.com/epfml/llm-baselines  for example

peacefulotter: I am not sure I made myself clear but batches are implemented and work. The only thing is that the text loader doesn't use the traditional .batch(batchSize) but rather constructs it itself

discojs/discojs-core/src/models/gpt/index.ts
peacefulotter: Ok yeah me too, check out maybe on this link: https://huggingface.co/datasets/wikitext
In general, any wikitext-103-raw dataset should be fine

JulienVig: Yes I've been working with the ones I found but none of them have the byte format you described

peacefulotter: The byte format is obtained by running the preprocessing script included under experiment/ I believe. Everything is documented in the experiment/README.md or even better in my repo disco-experiments

peacefulotter: To add a bit more on top of that, preprocessing is done outside of disco to avoid unecessary (repetitive?) computation. But the preprocessing I included can definitely be moved as a disco preprocessing step! Just needs a tokenizer with tokens encoded as 2 bytes max. If, for some reasons, you wish to use a different format (1 or 4 bytes for instance) then you need to modify preprocessing slightly and the toUint16 function

JulienVig: I see thanks! Did you consider using the gpt-tokenizer library at some point?

discojs/discojs-core/src/models/gpt/index.ts
tharvik: `Model` is using generators instead of callbacks, to be able to stop it from the outside and to get logs from every epoch. as such, each `fitDataset` call is configured to do a single epoch, but as the weights are not reset, `Model` is effectively running multiple epoch.

there are issue when using the whole dataset (some out of bound access), so the current code is only using the first `config.maxIter` batches.

DEV.md
```diff
@@ -75,8 +75,13 @@ npm --version
...
 **3.** Clone the repository
 
+As we are using some huge files in the repository, you'll need to install [`git-lfs`](https://git-lfs.com).
+Be sure to run `git lfs install` after installing it.
+
+We are using git submodules, so we need to also clone theses.
+
 ```
-git clone git@github.com:epfml/disco.git
+git clone --recurse-submodules git@github.com:epfml/disco.git
```
JulienVig: Why do we need submodules now?

datasets/README.md
```diff
@@ -0,0 +1,9 @@
+# Raw datasets
+
+This directory contains a selection of raw datasets.
+
+You'll need [`git-lfs`](https://git-lfs.com) to get some of theses.
```
JulienVig: So if I understood correctly git-lfs is not used yet, is this right?

discojs/discojs-core/src/aggregator/base.ts
```diff
@@ -54,10 +54,6 @@ export abstract class Base<T> {
   protected _communicationRound = 0
 
   constructor (
-    /**
-     * The task for which the aggregator should be created.
-     */
-    public readonly task: Task,
```
JulienVig, line 60: Was the aggregator's `task` attribute unused?

discojs/discojs-core/src/informant/training_informant/base.ts
```diff
@@ -71,6 +73,17 @@ export abstract class Base {
     return this.validationGraphInformant.accuracy()
   }
 
+  /** add loss for current round */
+  // eslint-disable-next-line accessor-pairs
+  set loss (loss: number) {
+    this._losses = this._losses.push(loss)
```
JulienVig, line 78: I think this syntax is quite misleading and leads to inconsistent getter - setter:
```js
trainingInformant.loss = logs.val_loss // actually pushes val_loss in a list  rather than set the value
console.log(trainingInformant.loss) // prints undefined
console.log(trainingInformant.losses.last() // this is the right syntax the value I just assigned, which I find inconsistent
```
This snippet explains why an undefined loss is being printed during training.
What do you think?

discojs/discojs-core/src/models/gpt/index.ts
JulienVig: Ah thanks! I'm surprised this is necessary to be able to stop and get training statistics though.

discojs/discojs-core/src/default_tasks/wikitext.ts
```diff
@@ -0,0 +1,40 @@
+import type { Model, Task, TaskProvider } from '..'
+import { TrainingSchemes, models } from '..'
+
+export const wikitext: TaskProvider = {
+  getTask (): Task {
+    return {
+      id: 'wikitext-103',
+      displayInformation: {
+        taskTitle: 'Wikitext 103 Raw',
```
martinjaggi: Let's call it `language modelling on wikitext`

discojs/discojs-core/src/default_tasks/wikitext.ts
```diff
@@ -0,0 +1,40 @@
...
+        summary: {
+          preview: 'In this challenge, we ask you to do next word prediction on a dataset of Wikipedia articles.',
+          overview: 'Wikitext-103-raw is a dataset comprising unprocessed text excerpts from Wikipedia articles, designed for tasks related to natural language processing and language modeling.'
```
martinjaggi, line 12: add a comment what tokenizer is used (type, and pretrained tokenizer downloaded from ...)

discojs/discojs-core/src/default_tasks/wikitext.ts
```diff
@@ -0,0 +1,40 @@
...
+      },
+      trainingInformation: {
+        dataType: 'text',
+        modelID: 'wikitext-103-raw-model',
+        validationSplit: 0.2, // TODO: is this used somewhere? because train, eval and test are already split in dataset
+        epochs: 10,
+        // constructing a batch is taken care automatically in the dataset to make things faster
+        // so we fake a batch size of 1
+        batchSize: 1,
+        scheme: TrainingSchemes.DECENTRALIZED,
```
martinjaggi: in the CI test you do federated. does it already work for all modes, train alone, federated, decentralized?

discojs/discojs-core/src/models/gpt/index.ts
```diff
@@ -0,0 +1,139 @@
...
+  vocabSize: number
+  maxEvalBatches: number
+}
+
+export class GPT extends Model {
+  private readonly model: GPTLMHeadModel
+
+  private static readonly batchSize = 4
+  private static readonly blockSize = 128
+  private static readonly vocabSize = 50258
```
martinjaggi, line 33: comment somewhere on where the tokenizer came from? (from the task definition maybe?)

server/src/router/federated/server.ts
```diff
@@ -317,13 +316,13 @@ export class Federated extends Server {
 
   /**
    * Appends a request to the logs.
-   * @param taskId The task id for which the request was made
+   * @param task The task id for which the request was made
    * @param nodeId The node id who made the request
```
JulienVig, line 320: I personally would have preferred naming it `taskId` to be consistent with `nodeId` and to not confuse a task object and a task id when a variable is called `task`. Especially if the function argument you input is `task.id`, then I would be surprised to see the function signature calling it `task` rather than `taskId`.
But I'm nitpicking so I'll let you choose, I saw you renamed it to `task` everywhere so up to you :)

discojs/discojs-core/src/models/gpt/index.ts
```diff
@@ -0,0 +1,139 @@
...
+  }
+
+  override set weights (ws: WeightsContainer) {
+    this.model.setWeights(ws.weights)
+  }
+
+  private convertCharDataset (dataset: Dataset): Dataset {
+    const batchSize = 4
+    const sampleSize = GPT.blockSize + 1
+    const chunkSize = sampleSize * batchSize * 2
```
martinjaggi, line 65: comment that 2 bytes will be one token id? BTW is it much more mem overhead to use an array (or stream) of int (to represent the token stream input) than the 2 bytes here?

discojs/discojs-core/src/models/gpt/model.ts
```diff
@@ -0,0 +1,542 @@
+import tf, { LayersModel, layers, serialization } from '@tensorflow/tfjs'
```
martinjaggi, line 1: comment that this whole file (or most?) is from `gpt-tfjs`? (or is it?)

discojs/discojs-core/src/default_tasks/wikitext.ts
martinjaggi: and also a comment on what is to expect (like if you train alone for 5mins or 5h, you should expect this train/test loss value roughly, and that the resulting model will somehow sound english?)

martinjaggi (commented):
amazing work, thanks!

i just left minor comments

martinjaggi (approved):

JulienVig (approved):
Great work thank you Valérian! I left some questions and opinions but you can proceed with the merge

discojs/discojs-core/src/models/gpt/index.ts
peacefulotter: I a not sure if my code is wrong but it is supposed to pull the exact number of bytes it needs for one batch. So chunkSize is the number of bytes in the read buffer not number of integers (4 bytes)

discojs/discojs-core/src/models/gpt/index.ts
peacefulotter: This is literally the library I used for preprocessing datasets. Specifically busing the davinci-003 tokenizer as it had a smaller vocabulary (50k instead of 100k) which saves a lot of memory since, at least to my knowledge, the ys need to be one hot encoded with size = vocabulary size

discojs/discojs-core/src/models/gpt/index.ts
peacefulotter: Ah yes.. I haven't really tested the loaders when reaching the end of the dataset - there might be some issues with my code alone in that regards. If so, I can try to fix it

discojs/discojs-core/src/models/gpt/model.ts
peacefulotter: Indeed it is, everything under gpt/ is a modification or straight up copy paste from gpt-tfjs.

discojs/discojs-core/src/default_tasks/wikitext.ts
tharvik: there are no tests to ensure that but it should.

discojs/discojs-core/src/models/gpt/model.ts
tharvik: good idea, I've also added the license

discojs/discojs-core/src/default_tasks/wikitext.ts
tharvik: there isn't really tokenization happening: `gpt-tokenizer` is unable to be compiled in the webapp (probably due to the lack of #632). currently, it takes wikitext as a stream of characters and pass it trough `GPT.convertCharDataset` to have the correct shape (don't know what its purpose is but it was required to make it work).

on the duration of training, I can only say that the loss is reducing. I hope that it'll write some correct english but I've no test to prove it.

hopefully, #646 will help us make it work and to general try it out.

DEV.md
tharvik: hum, indeed, we don't, it's some remains from a previous iteration; removing it

datasets/README.md
tharvik: correct, we don't need it currently, getting rid of it

discojs/discojs-core/src/default_tasks/wikitext.ts
peacefulotter: I think running gpt-micro on Shakespeare for 1 epoch should already produce some "shakespearesque" output

discojs/discojs-core/src/aggregator/base.ts
tharvik: pretty much yeah, it was easy to separate where it was used (server/router/federated mostly). I'm trying to reduce where it is used as to make it less of a global object.

discojs/discojs-core/src/models/gpt/index.ts
peacefulotter: I wonder if the tokenizer should be part of the task definition or be abstracted away. Is there really a use case for Disco where one might change this parameter? (Or even has the knowledge about what a tokenizer is)

discojs/discojs-core/src/informant/training_informant/base.ts
tharvik: hum, yeah, I got lost while developping there. I thought that users only want to refere to the whole history of losses and not the most recent one. re-adding the getter

server/src/router/federated/server.ts
tharvik: I see why it might be confusing. in dynamic typed languages, naming variable with its type is common and help with knowing what to do with the value. but in statically typed languages (such as strict TS), I don't find that it is helping as the type of a variable is known.
in a given function, when using `task` rather than `taskID`, I find it makes the code a bit more readable, as we don't really need to know that it's a full object or an ID, but rather what we'll do with it.
but yeah, that's clearly a matter of taste :)

peacefulotter: While I agree with Valerian, to be even more picky about this, I would personally name it just id, so that when a variable is called task it's always the actual Task object. This is also useful when looking up the word task over the whole codebase

PR #645, opened 2024-03-05
#643 fixes

tharvik:
I merged #643 a bit fast without waiting for @JulienVig review. here goes the fixes related to it.


cli/src/cli.ts
```diff
@@ -30,12 +29,12 @@ async function runUser (task: Task, server: Server, data: data.DataSplit): Promi
 }
 
 async function main (): Promise<void> {
-  const server = await startServer()
+  const [server, url] = await startServer()
```
JulienVig, line 32: What about making the url a server attribute in the future?

docs/examples/package.json
```diff
@@ -2,17 +2,26 @@
...
+    "lint": "npx eslint --ext ts --max-warnings 0 .",
+    "test": "npm run train"
   },
   "license": "ISC",
   "dependencies": {
     "@epfml/disco-server": "*",
+    "@epfml/discojs-core": "*",
     "@epfml/discojs-node": "*"
   },
+  "devDependencies": {
```
JulienVig, line 18: Should we add an instruction for installing dependencies in the docs/examples/README.md ? Is docs/examples included in the npm -ws commands?

docs/examples/training.ts
```diff
@@ -52,62 +67,26 @@ async function loadSimpleFaceData (task: Task): Promise<data.DataSplit> {
...
-    throw new Error('unable to get address to server')
-  }
-
-  return [server, new URL('', `http://${addr}`)]
-}
-
-
 // You can run this example with "npm start" from this folder
-main().catch(console.error)
+main('titanic').catch(console.error)
```
JulienVig: I feel this parameter can be easily missed, I think a constant in capslock at the top of the file would be more visible but less clean for sure

JulienVig (approved):
Great job, no change required!

cli/src/cli.ts
tharvik: there is two type of server exposed currently, `node:http.Server` & our `Disco` server. wrapping node's one fully inside `Disco` would be better, and maybe adding url attribute can be useful.
note that this field would `URL | undefined`, as the server has two states, started or not. using a function to start the server which return an URL ensures that the only way to get the value is by starting it. I find it quite elegant rather than to check `server.url !== undefined` to know that.

docs/examples/package.json
tharvik: hum, good question; yes? I'm not really sure that I understand the audience for this `docs/examples`.
 * if it is for devs to have a starter project based on disco, I think linter, how to deps and such are out of scope IMO as they would uses their own coding pratices
   * in this case, it won't be inside `workspaces` as it needs to be very independant
 * if it is for devs **of** disco, to understand how it is used, yeah, linter, how to deps, … does make sense
   * but why not directing them to directly read tests? that way, they can directly jump in code, it also pushes us to better write theses

docs/examples/training.ts
tharvik: clearly, I lost myself in wanting to scope everything :sweat_smile:
adding a screaming constant at the begin _of the function_ would be perfect
pushing one last change w/ it :)

cli/src/cli.ts
JulienVig: Indeed makes sense!

docs/examples/package.json
JulienVig: I was thinking of these examples for external users so yea I don't think making it a workspace is a good idea

Issue #639, opened 2024-02-28
Make custom tasks clearer and more accessible


JulienVig:
#690 should be addressed first.

Adding a custom task is currently very convoluted, and the process is not documented enough. 

For example, the no-code UI for adding a custom task requires users to choose parameter values of learning rate, DP sensitivity, gradient clipping etc which is certainly beyond the grasp of users without technical knowledge. Furthermore, users have to upload their own model which requires coding to some extent.

- `web-client`
    - [ ] Create a step-by-step guide to add a custom task from the UI, including how to create an initial tfjs model. Reference it from the UI custom task page.
    - [ ] Add more explanations for the custom task training and privacy parameters (e.g. on hover)
- `docs`
    - [ ] Most of the doc is currently in docs/Task.md, think about splitting information depending on how users are (non-technical, technical users, contributors) into different documents and reference them appropriately.
    - [ ] Add clear links to the repo documentation such that a non-technical user can easily find how to add a custom task from the UI
- `server` and `discojs`
    - [ ] @peacefulotter mentioned that the Task implementation could be improved, think about a potential rework with the ease of use and ease of adding user's custom task in mind.
- `discojs-node`
    - [ ] Currently, the only way to add a custom task from a script is to add it to the server before startup, which requires being the one handling the server. I think that being able to add a task as a user (i.e. using only `discojs-node`) should be possible.

tharvik:
> a custom task requires users to choose parameter values of learning rate, DP sensitivity, gradient clipping

that's clearly an issue, creating a `Task` is way too complex, there are so much fields, some very technicals, some that are only related to image or tabular, some that are unused even. it should be as easy as `Task { id: 'titanic', model: ... }` for the basic cases, with a maximum of fields having sane defaults.

> Furthermore, users have to upload their own model which requires coding to some extent.

I don't think we can really avoid that but a guide will indeed help a lot to do so.

peacefulotter:
Thanks for tagging me, would love to brainstorm / contribute for the task refactoring if needed

JulienVig:
Related to #647

tharvik:
> Related to #647

indeed, moving my (now deleted) response from there

> what's the purpose of tasks in Disco and what they are,

clearly agree, it's currently used as an global context object everywhere, with way too many fields. you can see some very early draft of a split of TrainingInformation @ https://github.com/epfml/disco/blob/647-split-tasks-tharvik/discojs/discojs-core/src/task/training_information.ts
[...]

PR #651, opened 2024-03-18
Add tokenization and prompting API to GPT models

### LINKED ISSUE #646: Add tokenization support to Disco LLMs
Previous and current works on LLM integration to DISCO relies on pre-tokenized datasets and doesn't account for token decoding after inference.

Full tokenizer support would allow:
- [x] To input natural text which is tokenized in DISCO
- [x] To perform Token decoding after model inference
- [x] To be able to use a pre-trained LLM in Disco first requires converting the weights to a format compatible with TF.js or other JS libraries. But additionally, we will also need to use the model's "pre-trained" tokenizer, and will also need to convert the tokenizers to JavaScript.
- [ ] Train a tokenizer with a particular algorithm (e.g. SentencePiece BPE) on a custom dataset, use it when training a model, and save it along with the model
- [ ] To save and store tokenizer alongside the model with which it was used

JulienVig:
Fixes #646 

* Adds tokenization support, allowing to choose any pre-trained tokenizer from Transformers.js
* Implement text generation with a trained model
* Extend text processing with shuffling and left padding
* Create an example of language model training in `docs/examples/wikitext.ts`
* I encountered a lot of issues and limitations which I listed in #654


JulienVig:
@tharvik @peacefulotter I'd be happy to hear your take on how to integrate tokenization and prompting into the Disco architecture.
The initial commit is a POC of a pipeline from text training data (so not already preprocessed) to textual output (cf `docs/example/wikitext.ts`)
I wanted to make the tokenizer as a `trainingInformation` field in the infamous TaskProvider but it seems that complex objects (like functions) are removed from the task object when communicated between the server and the client...
My next step is to migrate away from `gpt-tokenizer` to use [`Transformers.js`](https://huggingface.co/docs/transformers.js/api/tokenizers) such that we can specify which pre-trained tokenizer we want to use as a string in the trainingInformation

peacefulotter:
Congrats on the LLM POC milestone!

Firstly, I am wondering if having a custom tokenizer is even really relevant in the first place. Is there really a use case for it in Disco?

If so, instead of passing the tokenizer object, you could pass a string representing the tokenizer to use. Then an object could map tokenizer ids / names to the corresponding instance.

peacefulotter:
Secondly, since it appears tokenization needs to be done on the fly (before or during training) and not as a completely separate step. It could be done as a preprocessing step like Hugo planned to do on his branch

JulienVig:
@peacefulotter yes I want to do both points you mentioned! The second one is already implemented, the tokenization is part of the preprocessing.
And for the first I want to migrate to Transformers.js for that reason, it allows loading different pre-trained tokenizer from a string (rather than an import like gpt-tokenizer)

tharvik:
> I'd be happy to hear your take on how to integrate tokenization and prompting into the Disco architecture.

I don't like `Task`, it's the goto place for every context variable in Disco. here are some questions before adding it to task
  * is it applicable to every model? ie, can images be tokenized?
  * would someone want to change the tokenizer of wikitext?

after a bit more thinking, I don't think it really matters for now, as long as #639 is not done

> I wanted to make the tokenizer as a `trainingInformation` field in the infamous TaskProvider but it seems that complex objects (like functions) are removed from the task object when communicated between the server and the client...

indeed, functions can't be represented by msgpack/json so it gets removed. it'll be nice to have a serialization process to ensure that we get function back from the network (a string such as `gpt-tokenizer/davinci` in the serialized object that get mapped to a known function)

JulienVig:
Transformer.js requires Node >= 18, I'm waiting on #653 to be merged

web-client/.gitignore
```diff
@@ -1,3 +1,4 @@
 # local env files
 .env.local
 .env.*.local
+node_modules
```
JulienVig, line 4: A node_modules folder kept getting created, I think when running the web-client's tests. I don't know if it's me that messed up with something or the cypress upgrade

discojs/discojs-core/package.json
```diff
@@ -22,6 +22,7 @@
   "dependencies": {
     "@tensorflow/tfjs": "4",
     "@types/msgpack-lite": "0.1",
+    "@xenova/transformers": "^2.16.1",
```
tharvik: `npm i --save` likes to add the full version, but I do prefer to use the package-lock.json to keep track of the actual version we're supporting and keep package.json will only the major version

```suggestion
    "@xenova/transformers": "2",
```

I just noticed that "gpt-tokenizer" is also a dep, but was never used. can you also remove it?

discojs/discojs-node/src/data/text_loader.ts
```diff
@@ -1,14 +1,15 @@
...
-  async loadDatasetFrom (source: string): Promise<data.Dataset> {
-    // TODO sure, good idea to load the whole dataset in memory #irony
-    const content = await fs.readFile(source)
-    const file = new tfData.FileDataSource(content)
-
-    return new tfData.TextLineDataset(file)
+  loadDatasetFrom (source: string): Promise<data.Dataset> {
+    // TODO: reads all the file at once, 
+    // inputting the file path to FileDataSource isn't supported anymore
+    const inputFile = fs.readFileSync(source)
```
tharvik: better to use async `readFile` from 'node:fs/promises', as it allow for the other async methods to run while the file is loading.

```suggestion
    const inputFile = await fs.readFile(source)
```

discojs/discojs-core/src/task/training_information.ts
```diff
@@ -50,6 +50,14 @@ export interface TrainingInformation {
   // aggregator:  aggregator to be used by the server for federated learning, or by the peers for decentralized learning
   // default is 'average', other options include for instance 'bandit'
   aggregator?: AggregatorChoice
+  // tokenizerName (string). For example: 'Xenova/gpt2'. The name should match a Transformers.js tokenizer available on HuggingFace's hub. 
+  tokenizerName?: string
+  // tokenizer (object). The actual tokenizer. It is initialized according to the tokenizerName the first time it is needed for the subsequent tokenizations
+  // This field is not expected to be filled when initializing a field
+  tokenizer?: object | null
```
tharvik: I feel like both are linked and multually exclusive, did you try having it a single field?
also the only type of tokenizer we're currently supporting is a `PreTrainedTokenizer` (`object` is way too generic).
and I do recommend the use of `undefined` where possible instead of `null` (both are bottom types, why JavaScript..), as it's more "in the language" (partially init object, default args, … are undef)
smth like sounds nice: `tokenizer?: string | PreTrainedTokenizer`

discojs/discojs-core/src/dataset/data/preprocessing/text_preprocessing.ts
```diff
@@ -1,57 +1,84 @@
...
+      tokenizer = await AutoTokenizer.from_pretrained(tokenizerName)
+      task.trainingInformation.tokenizer = tokenizer
+    }
+    const maxLength = task.trainingInformation.maxSequenceLength ?? tokenizer.model_max_length as number
+    // Should never happen because tokenization truncates inputs
+    if (xs.size > maxLength) {
+      xs = xs.slice([0], [maxLength])
+    } else if (xs.size < maxLength) {
+      const paddingToken = tokenizer.pad_token_id
+      xs = xs.pad([[Math.max(0, maxLength - xs.size), 0]], paddingToken)
```
tharvik, line 41: ~~note there is [support for padding in transformer.js' tokenizers](https://github.com/xenova/transformers.js/blob/main/src/tokenizers.js#L2498). from what I saw in code, it's not allowing to left pad, only to right pad (still in a TODO state). consider adding a line mention to know where to look for updates.~~

and I read a bit further now, you already thought of that :)

discojs/discojs-core/src/models/gpt/evaluate.ts
```diff
@@ -1,50 +1,58 @@
...
 
+interface DataPoint extends tf.TensorContainerObject {
+  xs: tf.Tensor2D,
+  ys: tf.Tensor3D,
+}
+
 export default async function evaluate (
   model: tf.LayersModel,
-  dataset: tf.data.Dataset<{ xs: tf.Tensor, ys: tf.Tensor }>
+  dataset: tf.data.Dataset<{ xs: tf.Tensor, ys: tf.Tensor }>,
```
tharvik: no need to cast anymore in here, making the function safer

```suggestion
  dataset: tf.data.Dataset<DataPoint>,
```

discojs/discojs-core/src/models/gpt/evaluate.ts
```diff
@@ -1,50 +1,58 @@
...
   const acc: [number, number] = [0, 0]
 
-  await dataset.map(({ xs, ys }) => {
+  let iteration = 1
+  const iterator = await dataset.iterator()
+  let continueIterating = true
+  while (continueIterating) {
+    const next = await iterator.next()
+
+    const { xs, ys } = next.value as DataPoint
```
tharvik: the iterator is not infinite, you will [read an invalid `next.value`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#the_iterator_protocol) at the end of it (when `done` is true).
I don't really see the need of moving to a while loop instead of the stream/function interface: the latter takes care of the iteration for you, stopping when needed (use `dataset.take(maxEvalBatches)`).

discojs/discojs-core/src/models/gpt/evaluate.ts
```diff
@@ -1,50 +1,58 @@
...
+    const lossValue = await lossTensor.array()
+    if (typeof lossValue !== 'number') {
       throw new Error('got multiple loss')
     }
 
     const accTensor = tf.metrics.categoricalAccuracy(ys, logits)
     const accSize = accTensor.shape.reduce((l, r) => l * r, 1)
     const accSum = accTensor.sum()
-    const accSummed = await accSum.array()
+    const accSummed = accSum.arraySync()
```
tharvik: but why? async is highly concurrent when doing IO (which might happen depending on the running platform)

docs/examples/wikitext.ts
tharvik, line 1: can you add a `script` in package.json for it? and run it in the `test` script

docs/examples/wikitext.ts
```diff
@@ -0,0 +1,52 @@
...
+  const url = new URL('http://localhost:8080')
+  const tasks = await fetchTasks(url)
+  const task = tasks.get('wikitext-103')
+  if (task === undefined) { throw new Error('task not found') }
+  const dataset = await loadWikitextData(task)
+  const aggregator = new aggregators.MeanAggregator()
+  const client = new clients.federated.FederatedClient(url, task, aggregator)
+  const trainingInformant = new informant.FederatedInformant(task, 10)
+  const disco = new Disco(task, { scheme: 'federated', client, aggregator, informant: trainingInformant })
+  await disco.fit(dataset)
```
tharvik: it's a huge block of code which is not very understable to newcomers: can you add a few comments, new lines to make it least daunting plz?

web-client/.gitignore
tharvik: yeah, I got it too, it's for a cache of some build files (as if we don't have temporary dir dedicated to that), it's not triggering anymore with the bump of vue, let's add it for now

tharvik (approved):
great work! a bit of nitpicking here and there but nothing really vital

discojs/discojs-core/src/models/gpt/evaluate.ts
JulienVig: Yeah so I took the sad decision to use a while loop instead of the functional way because there was a memory leak when streaming over the dataset. I am still not exactly sure of where it happened but I suspect the iterator values were not correctly disposed. The while loop allows me to ensure it is done.
The check for `next.done` is done at the end of the loop, but it assumes at least one datapoint in the dataset, I'm adding an initial check on next.done before the while

discojs/discojs-core/src/models/gpt/evaluate.ts
JulienVig: Good catch, it was not intentional 👍

discojs/discojs-core/src/task/training_information.ts
JulienVig: Yes good idea I'll try that now! The null was because of a linting error I wasn't sure how to solve but it should not be needed anymore

discojs/discojs-core/src/models/gpt/evaluate.ts
JulienVig: Actually I thought you were talking about the training loop which I also converted from functional stream to an iterator loop and where the memory leak was occurring. 
There is still a growing number of allocated tensors (#654), especially in the evaluation loop, that's why I also tried changing the evaluate loop but it doesn't solve the issue. I'm reverting the evaluation loop to be functional.

discojs/discojs-core/src/models/tokenizer.ts
JulienVig, line 1: @tharvik I've created a new `models/tokenizer.ts` file because code to load/initialize the tokenizer was getting duplicated. Let me know if see anything to improve!

JulienVig, line 1: @tharvik I've created a new `models/tokenizer.ts` file because code to load/initialize the tokenizer was getting duplicated. Let me know if you see anything to improve!

discojs/discojs-core/src/models/gpt/evaluate.ts
```diff
@@ -43,8 +49,46 @@ export default async function evaluate (
...
+
+  //   datasetSize += 1
+  //   totalLoss += lossValue
+  //   acc[0] += accSummed
+  //   acc[1] += accSize
+
+  //   tf.dispose([xs, ys, logits, accTensor, accSum, lossTensor, next.value])
+  //   iteration++
+  //   continueIterating = next.done !== true && iteration <= maxEvalBatches
+  // }
```
tharvik: let's not keep commented code around, it'll get out of sync

discojs/discojs-core/src/dataset/data/preprocessing/text_preprocessing.ts
```diff
@@ -1,57 +1,74 @@
...
-const gpt3Tokenizer = null as unknown as { encode: (_: string) => { bpe: number[]; text: string[] } }
+/**
+ * We are currently only implementing left padding for text generation 
+ * https://huggingface.co/docs/transformers/en/llm_tutorial#wrong-padding-side
+ * The function can easily be extended to support right padding once the need arise
+ */
+const leftPadding: PreprocessingFunction = {
+  type: TextPreprocessing.LeftPadding,
+  apply: async (x: Promise<tf.TensorContainer>, task: Task): Promise<tf.TensorContainer> => {
+    let { xs } = await x as TokenizedEntry
```
tharvik, line 28: just to be sure that people use it correctly, can you check that the `tf.TensorContainer` is really containing a field `xs`, of type `tf.Tensor` with `rankType === Rank.R1`?

discojs/discojs-core/src/dataset/data/preprocessing/text_preprocessing.ts
```diff
@@ -1,57 +1,74 @@
...
+ * Tokenize and truncates input strings
+ */
 const tokenize: PreprocessingFunction = {
   type: TextPreprocessing.Tokenize,
-  apply: (x: tf.TensorContainer) => {
-    const { xs, ys } = x as TextEntry
-
-    const tokenized = gpt3Tokenizer.encode(xs[0]).bpe
+  apply: async (x: Promise<tf.TensorContainer>, task: Task): Promise<tf.TensorContainer> => {
+    const xs = await x as string // tf.TextLineDataset yields strings
```
tharvik, line 63: same idea of checking that we indeed get `string` as input

tharvik (approved):
> I've created a new models/tokenizer.ts file

LGTM!

a few more comments (I'm a never ending stream of critics, please stop me)

discojs/discojs-core/src/models/gpt/evaluate.ts
JulienVig: Good catch thanks! that was completely unintentional

Issue #634, opened 2024-02-21
packages hierarchy should be a tree


tharvik:
currently, the build graph is a tree, but the testing graph isn't
```mermaid
flowchart TD
    DC[discojs-core] --> |build| DN[discojs-node]
    DC --> |build| DW[discojs-web]
    DN --> |build & test| S[server]
    DW --> |build & test| W[web-client]
    DN --> |build| C[CLI]
    DN --> |test| DC
    S --> |test| DC
```
that is quite cumbersome to know what is the origin of a bug, as when a test in discojs-core fails, the code one need to check is the union of discojs-core, discojs-node & server. it's also weird to need to build discojs-node to compile test for discojs-core.

Issue #633, opened 2024-02-19
mv from `wrtc` to `node-datachannel`


tharvik:
`wrtc` is an abandonned project, a bunch of people are actually building it by themself, publishing binaries around but there isn't a concerted effort. it'll be nice to actually get rid of it a some point and use `node-datachannel` that is more modern.

it currently causes [some spurious segfaults](https://github.com/WonderInventions/node-webrtc/issues/7), that sometimes triggers on exit, that's quite annoying #heisenbug

Issue #597, opened 2023-09-11
Byzantine robust aggregator


s314cy:
Following the aggregator refactor in #578, byzantine-robust aggregation in the federated setting does not work anymore as it remains to be re-implemented with the new architecture in mind.

JulienVig:
The work pre-refactoring has been left in the code unused, we are removing it in #652. For future use, it can be found in `discojs/discojs-core/src/aggregator/robust.ts` at commit [b9db50f](https://github.com/epfml/disco/blob/b9db50f3e777d8f9cfa343db5ae4c8887314862f/discojs/discojs-core/src/aggregator/robust.ts)

tharvik:
* the initial implementation was done in #489
* based on [a MLO paper](https://arxiv.org/abs/2012.10333)
* byzantine robust in this case is robustness against "bad weights"
* seems quite douable (added to v4 milestone)

thanks @JulienVig for the code de-zombification

PR #652, opened 2024-03-19
*: upgrade to es2022

tharvik:
move to the current ECMA standard of modules #632

in addition
* replace `wrtc` segfaulting dep by `node-datachannel` #633
* reenable cypress (web-client browser tester) which hasn't been for a while
* fix a **bunch** of eslint issues
* simplify locales (languages) which was splitted around and somehow still written in JavaScript
* ensure completness of Task/TrainingInformation/... [type guards](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates)
* remove test dep cycle #634

from a quick test, we can now support importing and using newer library, such as `gpt-tokenizer` or Transformer.js


martinjaggi (approved):

tsconfig.base.json
```diff
@@ -0,0 +1,15 @@
+{
+  "compilerOptions": {
+    "module": "node16",
```
JulienVig, line 3: Why node16 rather than node20?

discojs/discojs-core/src/aggregator/robust.ts
```diff
@@ -1,5 +1,7 @@
-import { Base as Aggregator } from './base'
-import type { client, Model, WeightsContainer } from '..'
+// TODO remove as nothing is implemented
```
JulienVig: Do you want to remove the file completely? I linked the current state of work in the relevant issue #597

discojs/discojs-core/src/client/federated/base.ts
```diff
@@ -144,25 +146,6 @@ export class Base extends Client {
...
-      taskId: this.task.id,
-      nodeId: this.ownId,
-      round: this.aggregator.round,
-      key,
-      value
-    }
-
-    this.server.send(msg)
-  }
-
```
JulienVig, line 165: Was it unused?

discojs/discojs-core/src/training/disco.ts
```diff
@@ -56,26 +47,26 @@ export class Disco {
...
           break
-        case TrainingSchemes.DECENTRALIZED:
+        case 'decentralized':
           options.client = new clients.decentralized.DecentralizedClient(options.url, task, options.aggregator)
           break
-        default:
+        case 'local':
           options.client = new clients.Local(options.url, task, options.aggregator)
           break
       }
```
JulienVig: Should we add a fallback error?

discojs/discojs-core/src/training/disco.ts
```diff
@@ -56,26 +47,26 @@ export class Disco {
...
           break
-        case TrainingSchemes.DECENTRALIZED:
+        case 'decentralized':
           options.informant = new informants.DecentralizedInformant(task)
           break
-        default:
+        case 'local':
           options.informant = new informants.LocalInformant(task)
           break
       }
```
JulienVig: Same here

discojs/discojs-core/src/weights/aggregation.ts
```diff
@@ -1,8 +1,8 @@
 import { List } from 'immutable'
-import tf from '@tensorflow/tfjs'
+import * as tf from '@tensorflow/tfjs'
```
JulienVig, line 2: I see you've changed the way tf is imported in a lot of places, why is it needed?

web-client/src/components/training/Trainer.vue
```diff
@@ -106,33 +106,29 @@ export default defineComponent({
...
           break
-        case TrainingSchemes.DECENTRALIZED:
+        case 'decentralized':
           this.trainingInformant = new informant.DecentralizedInformant(...args)
           break
-        default:
+        case 'local':
           this.trainingInformant = new informant.LocalInformant(...args)
           break
       }
```
JulienVig: Same here, no default case

JulienVig (approved):
Very good job thanks for your work!! No blocking comments

JulienVig:
Closes #632, closes #633, closes #634

JulienVig:
The PR is merging into 633-bump-node-tharvik rather than develop, is it intentional?

tsconfig.base.json
tharvik: that's a bit confusing but correct: [`module` defines the module system](https://www.typescriptlang.org/tsconfig#module) (which supports ESM since node v16 so called "node16", versus only supporting CommonJS called "node"), not the target runner (node v20 in our case, or browsers).

discojs/discojs-core/src/client/federated/base.ts
tharvik: I didn't saw any reference to it when looking, which looked weird. it seems that messages where implemented but never used nor tested. do you think it's worth keeping it?

JulienVig: It seems it's from before the aggregator refactoring. I think we can remove it but then we should also remove it from the message type enumeration and from the server message type switch in server.ts

discojs/discojs-core/src/training/disco.ts
tharvik: in the potential `default` case, the type of `options.scheme` is actually `never`, thanks to [exhaustiveness checking](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#exhaustiveness-checking) there is no way that it'll happen (iff types are correct), so it'd be dead code.
but yeah, let's add a compile time error to ensure that's it's indeed not happening (might be the case when changing types)

discojs/discojs-core/src/aggregator/robust.ts
tharvik: I'm all for cleanup, thanks for linking on the issue

discojs/discojs-core/src/weights/aggregation.ts
tharvik: it was needed while during the implementation of ESM, but it is not needed per se anymore.

the semantic of `import` changes with ESM, before that, it's a mix of TypeScript automagic importer, based on the type of module imported (CommonJS, AMD, ...), with helpers. if you want, you can read the whole issues around it on the [module interop page](https://www.typescriptlang.org/docs/handbook/modules/appendices/esm-cjs-interop.html).

in theses cases, the default import (`import ... from ...) for tfjs (and many other modules) is quasi-equivalent to the namespace import (`import * as ... from ...`) but that's not always the case. I prefer to be explicit with such modules, to show more clearly how the import will be used. there are also probably some benefits with tree shaking thus reduce build time & size.

tharvik:
> The PR is merging into 633-bump-node-tharvik rather than develop, is it intentional?

it is! it was based on it and now that #633  is merged, the base branch is automatically changed to "develop".
that's a way to have some kind of linearity in the PRs that I found quite useful.

PR #653, opened 2024-03-21
bump node

tharvik:
currently, we're running with node 16, which is not LTS anymore. some packages requires newer nodejs, [`wrtc` is dead](https://github.com/epfml/disco/issues/633) (and not compatible with node 16 anyway), so upgrading is needed.


tharvik:
node-datachannel requires esm modules, moving it #652.
hopefully, bumping to nodejs 20 won't segfault too much

DEV.md
```diff
@@ -98,13 +87,13 @@ Or you can select a specific workspace via `--workspace=$name-or-path-to-package
 npm -ws run build
```
JulienVig, line 87: Running `npm -ws run build` fails with
```
> @epfml/discojs-core@2.1.1 build
> tsc

sh: tsc: command not found
npm ERR! Lifecycle script `build` failed with error:
npm ERR! Error: command failed
npm ERR!   in workspace: @epfml/discojs-core@2.1.1
npm ERR!   at location: /Users/.../discojs/discojs-core
```
Do we require installing typescript globally?

JulienVig: It may be a MacOS-specific issue: https://stackoverflow.com/a/61360373

JulienVig (commented):
Looks good!

I've quickly tried to import and use Transformers.js but I get an import error.
It's the same as in this [stackoverflow post](https://stackoverflow.com/questions/76883048/err-require-esm-for-import-with-xenova-transformers/76902405#76902405), which is answered by Transformers.js's author
Following the answer I can circumvent the issue like that:
```js
const TransformersApi = Function('return import("@xenova/transformers")')()
const { AutoTokenizer } = await TransformersApi
const tokenizer = await AutoTokenizer.from_pretrained('Xenova/bert-base-uncased')
const { tokens } = await tokenizer('I love transformers!')
```
The Function import triggers a linting error and honestly looks quite hacky
What do you think? I'm bringing this up here in case we need to change things before merging

DEV.md
tharvik: hum, it seems that `tsc` (from the `typescript` dep) is not installed. check that you don't have any node_modules around, and `npm ci` from the root, that should do it.

tharvik:
> I've quickly tried to import and use Transformers.js but I get an import error.

haa, welcome to the ESM (ECMAScript Module) world! this error means that node is trying to read an ESM module as a CommonJS one (old node-specific module system). to be able to load it, we need ESM support (which requires to change typescript target & module to be at least ES2015). and that's what I'm trying to fix with #652.

> The Function import triggers a linting error and honestly looks quite hacky What do you think? I'm bringing this up here in case we need to change things before merging

this is indeed quite hacky, does it work at all?
anyway, if it helps you progress, keep it for now, and (as I'll probably forget), remember me to remove it when reviewing #652

JulienVig (approved):
All good! Comments are resolved by #652

Issue #534, opened 2022-11-07
Linting mismatch


s314cy:
Linting matches in all projects (`discojs`, `server`,`cli`) but `web-client`, meaning the `web-client` project enforces different coding styles.

To go further, one would want to integrate linting specific to Vue 3's composition API with the `setup` syntactic sugar https://vuejs.org/api/sfc-script-setup.html.

JulienVig:
@tharvik is this issue still relevant?

tharvik:
yep, it's currently in the making and should be fixed by #655

tharvik:
sadly, it isn't fixed by #655: the combo eslint flat + workspaces + vue + typescript is not documented. one should check https://eslint.vuejs.org/user-guide/ for updates.

hleepfl:
Outdated, eslint config works centrally now

PR #655, opened 2024-03-28
updates dependencies

tharvik:
currently, there is a bunch of security issues, mostly from old dependencies. this PR aims to fix that

* build web-client with `vite` instead of the deprecated `vue-cli`
  * way faster build time, ESM hot replacement
  * add `strict` compile options
  * add `lang="ts"` to all script tags
  * ~~commonize eslint #534~~
* bump all packages to latest versions
  * will reenable dependabot updates after that
  * `immutable@5` is still in beta, even if `npm outdated` marks it as updatable
  * `eslint@9` is too recent for plugins to support it yet
* drop doc generation (outdated and unused)
* drop dead web-client/src/components/data/preview (unused and depending on removed code)

also partially fix a race condition when receiving peers in decentralized. it would need a better synchronisation mechanism (an ACK when peer updates are done, like a real RPC) to ensure that slower nodes are indeed ready to communicate.

also rework the cypress tests to remove some uneeded files, and add a test of the training.


JulienVig (changes requested):
Good work!! You managed to modify almost every single file in the repo haha. Though I don't have anything to comment on the code.

* Can you update the Vus.js section of the [Contributing guide](https://github.com/epfml/disco/blob/develop/docs/CONTRIBUTING.md#contributing-to-web-client) with updated information now that we are using Vite?

* I see you've re-written some logic in the web-client, have you played around in the web client to replace our missing tests? I've quickly trained a titanic model and the line plot doesn't seem to be updated anymore during training for example.

tharvik:
soo, I driffted a bit from the original PR aim, oupsi

* added ~workaround for "unknown signal from peer"
* update cypress tests
* change training to expose logs of round & epochs and avoid mutating objects
  * remove a bunch of dead/killable callbacks
  * rm now obsolete TrainingInformant
  * GraphInformant is still around as it is used by Tester/Validator; but I don't see the need for keeping track of the evolution of the accuracy, do we remove it?

cli/src/cli.ts
```diff
@@ -1,22 +1,22 @@
...
+async function runUser (task: Task, url: URL, data: data.DataSplit): Promise<unknown> {
   const client = new clients.federated.FederatedClient(url, task, new aggregators.MeanAggregator())
 
   // force the federated scheme
   const disco = new Disco(task, { scheme: 'federated', client })
 
-  await disco.fit(data)
-  await disco.close()
-  return await disco.logs()
+  const logs = await disco.fitted(data)
```
JulienVig: I'm not a fan of the renaming to `fitted`, was there a particular reason? I guess it takes me off-guard because I'm used to seeing such methods in the imperative tense: `fitDataset` in tensorflow or `await fs.writeFile` rather than `await fs.wroteFile`

discojs/discojs-core/src/training/disco.ts
```diff
@@ -1,17 +1,18 @@
...
 
+import type { RoundLogs, Trainer } from './trainer/trainer.js'
+import { TrainerBuilder } from './trainer/trainer_builder.js'
+
 export interface DiscoOptions {
   client?: clients.Client
   aggregator?: Aggregator
   url?: string | URL
   scheme?: TrainingInformation['scheme']
-  informant?: TrainingInformant
```
JulienVig, line 14: I thought the training informant was used to keep track of training metrics, can it be removed because there didn't used to be training logs?

web-client/src/components/containers/ImageCard.vue
```diff
@@ -20,7 +20,7 @@
       <CustomButton
         class="text-xs"
       >
-        Click to show on Maps
+        click to show on maps
```
JulienVig, line 23: I feel we need a convention for capitalization, so far I think the web-app has been capitalizing everything and I gather you'd go for all lower case, what about only capitalizing the first word? Right now the web-app is not consistent, with some text capitalized and other in lower case

JulienVig (changes requested):
Lots of good stuff, thanks!!

I got an error specific to silicon chips when trying to build the web-client, I copy-pasted below the stack below. I followed the advice to delete package-lock.json and node-modules and run `npm i` and I can now build successfully. However the package-lock.json got modified, so I'm pushing it so you can check it out and decided whether to keep the now one or revert the previous one.
<details>
  <summary>Error stack</summary>

```bash
> build
> vue-tsc --build && vite build

/Users/.../disco/node_modules/rollup/dist/native.js:59
		throw new Error(
		      ^

Error: Cannot find module @rollup/rollup-darwin-arm64. npm has a bug related to optional dependencies (https://github.com/npm/cli/issues/4828). Please try `npm i` again after removing both package-lock.json and node_modules directory.
    at requireWithFriendlyError (/Users/.../disco/node_modules/rollup/dist/native.js:59:9)
    at Object.<anonymous> (/Users/.../disco/node_modules/rollup/dist/native.js:68:76)
... 3 lines matching cause stack trace ...
    at Module._load (node:internal/modules/cjs/loader:1023:12)
    at cjsLoader (node:internal/modules/esm/translators:356:17)
    at ModuleWrap.<anonymous> (node:internal/modules/esm/translators:305:7)
    at ModuleJob.run (node:internal/modules/esm/module_job:218:25)
    at async ModuleLoader.import (node:internal/modules/esm/loader:329:24) {
  [cause]: Error: Cannot find module '@rollup/rollup-darwin-arm64'
  Require stack:
  - /Users/.../disco/node_modules/rollup/dist/native.js
      at Module._resolveFilename (node:internal/modules/cjs/loader:1144:15)
      at Module._load (node:internal/modules/cjs/loader:985:27)
      at Module.require (node:internal/modules/cjs/loader:1235:19)
      at require (node:internal/modules/helpers:176:18)
      at requireWithFriendlyError (/Users/.../disco/node_modules/rollup/dist/native.js:41:10)
      at Object.<anonymous> (/Users/.../disco/node_modules/rollup/dist/native.js:68:76)
      at Module._compile (node:internal/modules/cjs/loader:1376:14)
      at Module._extensions..js (node:internal/modules/cjs/loader:1435:10)
      at Module.load (node:internal/modules/cjs/loader:1207:32)
      at Module._load (node:internal/modules/cjs/loader:1023:12) {
    code: 'MODULE_NOT_FOUND',
    requireStack: [
      '/Users/.../disco/node_modules/rollup/dist/native.js'
    ]
  }
}
```
</details>

It looks like updating dependencies created a few breaking changes in the web-client...
* During training, logs are not displayed in the "Training logs" div, but raw logs get printed in the lower left corner of the screen
Screenshot 2024-04-10 at 11 55 03

* After training, clicking on Test the model doesn't work and instead of a toaster, the error "Model not trained" gets also printed in the lower left corner of the page
Screenshot 2024-04-10 at 11 51 19

If you have the time and motivation, writing test cases for these errors would be great! 😁

web-client/src/components/containers/ImageCard.vue
JulienVig: I realized that maybe it's because some text is stylized in uppercase, is it to removing casing in the code of text being stylized in upper case? I pushed a commit to revert the lowercasing of the last 2 training steps which are not uppercased in the styling (the first 3 were still in capital casing)

discojs/discojs-core/src/training/disco.ts
JulienVig: This is answered in your last comment, then can't we also use the logs for the graph informant? I'm not sure what's the difference between training informant and graph informant

cli/src/cli.ts
tharvik: I wanted to potentially simplify the use of `Disco.fit`, thinking that awaiting a value rather than using a generator function would be easier. but yeah, I don't like it either, one can either `for await (const _ of disco.fit);` or get the logs themself. getting rid of it

discojs/discojs-core/src/training/disco.ts
tharvik: > I thought the training informant was used to keep track of training metrics

indeed, it was, but was partially done and IMO a bad design: there is no clear information on when the metrics are valids, when to reset theses, what was exposed (mutating state is hard). having a generator of logs is way safer and also clearer as when values are generated.

> I'm not sure what's the difference between training informant and graph informant

it's not really clear for me either. from what I gathered, it is used for keeping a list of recently generated logs of training in order to build an ApexChart. it's is only used in `Validator` now but I didn't rework the latter so I'm kind of scared of digging another rabbithole with wanting to remove it whole. anyway, let's keep it for now.

web-client/src/components/containers/ImageCard.vue
tharvik: > I realized that maybe it's because some text is stylized in uppercase, is it to removing casing in the code of text being stylized in upper case?

correct! having a specific case (lowercase being my default) was looking weird when it was actually uppercased on display. and matched with `ignoreCase` throughout the tests. so having everyone of theses fields as the same casing made it clearer.

> I pushed a commit to revert the lowercasing of the last 2 training steps which are not uppercased in the styling (the first 3 were still in capital casing)

good catch, I was a bit fast on it. it's quite probable that I forgot some other mixed-case-value-that-are-shown-as-uppercase field, but that's not that important.

tharvik:
> I got an error specific to silicon chips when trying to build the web-client
> However the package-lock.json got modified, so I'm pushing it so you can check it out and decided whether to keep the now one or revert the previous one.

yeah, it's a faster moving space now that dependencies are actual again. `vue` and friends are moving fast and we're only enforcing major versions.
btw, in between your remarks and me updating the PR, it changed again :rocket:

> * During training, logs are not displayed in the "Training logs" div, but raw logs get printed in the lower left corner of the screen

ho, I forgot to actually load a theme in the Toaster! I don't think we really need messages at all (the toaster itself is alreay quite fancy, there are better way to show that's models are training such as via the graphs we have).
but for now, I forwarded the logger output in there, that makes it a bit more fun to watch.
and a test that messages are shown.

> * After training, clicking on Test the model doesn't work and instead of a toaster, the error "Model not trained" gets also printed in the lower left corner of the page

ho, it seems that the model is not stored in memory automatically anymore. last commit should fix it.

> If you have the time and motivation, writing test cases for these errors would be great! 😁

héhé, and regressions tests added. it's actually super fast to write some with `vitest`'s components testing. we have no more excuses!

another point: after discussion with Klavdiia, it seems that I forgot to reimplement the calling of `onBatch{Begin,End}`, so no communication (or saving of the models) was done anymore! it wasn't catched because the server tests weren't shuffeling the datasets and the same initial model was used.

JulienVig (approved):
Looks good to me!

I noticed an error when predicting after training a model and I pushed a fix myself, there's a high chance this bug is my fault...

Issue #654, opened 2024-03-27
Improve and rework GPT-tfjs


JulienVig:
Here is a list of potential improvements for gpt-tfjs in Disco:
- [x] Create a compile method to initialize the optimizer (rather than initializing it when fitDataset is called). This ensures the optimizer state is persisted across multiple calls to fitDataset
- [x] Implement save and load methods to save and re-use a trained model
- [x] Rename classes for better clarity and consistency, e.g. multiple classes and functions are called `GPT`
- [x] Assess whether we can use tf.CustomCallbackArgs rather than redefining an interface for TrainingCallbacks
- [x] Assess whenever we can use TFJS' native fitDataset method rather than overriding it with a custom training loop
-> tfjs only implements Adam while GPT2 uses AdamW. Additionally, the custom optimizer allows having weight decay which is used in the original GPT2.
- [x] Reading a text file with TF.js only supports reading line by line which is not ideal for LLM inputs, try implementing a file reader chunk by chunk rather than by lines 
- [x] Training with gpt2 has NaN loss after the first epoch step 

#656 and #657 should be addressed first

Issue #647, opened 2024-03-05
split default tasks from core


tharvik:
currently, most tasks are hardcoded inside discojs-core, which looks to me as out of place: it kinda encourage people to expect that it will always contains theses tasks thus directly depending on it & it also mixes where the tests of theses tasks are with the core code (it is kinda done in server eventhough it's not really the correct place to do it).

splitting theses in a new package, `@epfml/discojs-tasks` would allow for easier adding/modifing/removing of tasks, and have a dedicated place for testing theses. we would also be able to store raw data (such as image or models) directly there.

JulienVig:
Before refactoring the tasks I think it is important for us to define what's the purpose of tasks in Disco and what they are, especially what are the use cases for pre-defined tasks and custom tasks. And how LLM use would fit in those.

tharvik:
we would also be able to store raw data (such as image or models) in there

tharvik:
> Before refactoring the tasks I think it is important for us to define what's the purpose of tasks in Disco and what they are,

clearly agree, it's currently used as an global context object everywhere, with way too many fields. you can see some very early draft of a split of `TrainingInformation` @ https://github.com/epfml/disco/blob/647-split-tasks-tharvik/discojs/discojs-core/src/task/training_information.ts
I'll try a definition of what I think it should be, discussion very welcome: a `Task` is a `Model` with a description. it's what users of disco will participate in, by adding data (training) or predicting with.
for now, it's also containing model specific config (mv to `Model` init), dataset config (mv to `Model` inputs type and dataset types) and network config (maybe showing which network are currently running).

> especially what are the use cases for pre-defined tasks

pre-defined task is for to showoff the possibilities of disco itself, but there is clearly no need to uncumber discojs-core with it. it should be only available for trying out/demo purposes. when put outside, it will effectively be the same as adding custom tasks.

> and custom tasks

custom task would be for specific uses of disco (such as the various bilateral projects that's being develop at MLO).

PR #658, opened 2024-04-08
Fix gpt-tfjs bugs, add tests and refactor code

### LINKED ISSUE #656: gpt-tfjs only repeats the last prompt token
Here are some issues with gpt-tfjs I noted while implementing tokenization:
- [x] There is a memory leak in the training loop. The memory doesn't grow much (~0.01MB per iteration) but the number of tensors keep growing (+14 new tensors allocated per dataset batch)
- [x] A trained model (e.g. on wikitext with iteration>1000, validation perplexity<4) almost always repeats the last prompt token
- [x] Create a test case for the wikitext task

JulienVig:
Addresses items listed in #654 #656

Fixes #656:
- [x] Fixes text preprocessing, now generation doesn't repeat the last input token and memory leak is fixed
- [x] Add test cases for gpt-tfjs
- [x] Create test cases for text preprocessing

Addresses parts of #654 
- [x] Add explanation comments to gpt-tfjs code
- [x] Refactor classes and code organization
- [x] Add a compile method

Additionally 
- Add support for choosing model parameters at initialization (rather than hard coded in the task or the class definition)
- Add support for model config serialization


JulienVig:
@tharvik do you have an opinion on putting the gpt-tfjs test spec file, which relies on tfjs-node, in the `discojs-core/.../gpt`? Or better to put it in discojs-node?

tharvik:
> @tharvik do you have an opinion on putting the gpt-tfjs test spec file, which relies on tfjs-node, in the `discojs-core/.../gpt`? Or better to put it in discojs-node?

for now, there isn't a good place to put theses tests, it's de facto in `server/tests`.
with default tasks splitted from discojs-core (tracked in #647), it'll fit naturally in this "tasks" package

JulienVig:
> for now, there isn't a good place to put theses tests, it's de facto in server/tests.
with default tasks splitted from discojs-core (tracked in https://github.com/epfml/disco/issues/647), it'll fit naturally in this "tasks" package

The thing is that this test is independent of the task, I tried to keep it as unitary as possible. It only tests gpt-tfjs implementation and capabilities

tharvik:
> > with default tasks splitted from discojs-core (tracked in #647), it'll fit naturally in this "tasks" package
> 
> The thing is that this test is independent of the task, I tried to keep it as unitary as possible. It only tests gpt-tfjs implementation and capabilities

ho, I misread, I though it was requiring **discojs**-node, not tfjs-node; well done then! so yeah, having it in discojs-core makes clearly sense. you can add tfjs-node to dev deps, so that's available for building/testing but not exposed from discojs-core (sadly not enforced).

discojs/discojs-core/src/dataset/data/preprocessing/text_preprocessing.spec.ts
```diff
@@ -0,0 +1,99 @@
+import { TEXT_PREPROCESSING } from './index.js'
+import { expect } from 'chai'    
+
+import type { Task } from '../../../index.js'
+import * as tf from '@tensorflow/tfjs'
+
+describe('text preprocessing', function () {
+  const [tokenize, leftPadding] = TEXT_PREPROCESSING
+  // Use a function to create different task object for each test (otherwise the tokenizer gets cached)
+  const initMockTask: () => Task = () => {
```
tharvik: I find the function notation to be much more readable.

```suggestion
  function initMockTask(): Task {
```

discojs/discojs-core/src/dataset/data/preprocessing/text_preprocessing.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+  const expectedTokens = [15496, 995, 11, 257, 47125, 352, 2242, 2231, 11, 705, 30, 860, 4304, 13, 15312, 318, 257, 1479, 2695, 2691, 45352, 3194, 290, 9456, 416, 257, 2055, 220, 198, 286, 11661, 11, 1900, 355, 11145, 46647, 1547, 13, 4062, 276, 416, 12963, 11769, 290, 13633, 311, 2564, 319, 3269, 1315, 11, 5878, 11, 15312, 318, 12007, 416, 262, 44877, 5693, 11, 281, 1605, 15346, 4009, 326, 24803, 257, 3085, 286, 625, 13037, 661, 3693, 22, 60]
+
+  it('can tokenize text', async () => {
+    const { tokens } = await tokenize.apply(Promise.resolve(text), initMockTask()) as { tokens: number[]}
+    expect(tokens).to.be.deep.equal(expectedTokens)
+  })
+
+  it('can truncate inputs when tokenizing', async () => {
+    const truncationTask = initMockTask()
+    truncationTask['trainingInformation']['maxSequenceLength'] = 10
```
tharvik: it ensure that the field exists in `TrainingInformation` at compile time

```suggestion
    truncationTask.trainingInformation.maxSequenceLength = 10
```

discojs/discojs-core/src/dataset/data/preprocessing/text_preprocessing.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+    const { tokens } = await tokenize.apply(Promise.resolve(text), truncationTask) as { tokens: number[] }
+    const expectedLength = truncationTask.trainingInformation.maxSequenceLength + 1 // + 1 because tokenization includes an extra token label for next label prediction
+    expect(tokens.length).to.be.equal(expectedLength)
+    expect(tokens).to.be.deep.equal(expectedTokens.slice(0, expectedLength))
+  })
+
+  it('can left pad tokens', async () => {
+    // Create a task where output token sequence should all have length 20
+    const paddingTask = initMockTask()
+    paddingTask['trainingInformation']['maxSequenceLength'] = 20
```
tharvik:
```suggestion
    paddingTask.trainingInformation.maxSequenceLength = 20
```

discojs/discojs-core/src/dataset/data/preprocessing/text_preprocessing.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+    expect(await ys.sum(-1).array()).to.be.deep.equal(expectedOneHot)
+    
+    // In each row, the index of the 1 should be the token id
+    const expectedYs = Array.from({length:10}).map(_ => paddingToken).concat(tokens.tokens)
+    expect(await ys.argMax(-1).array()).to.be.deep.equal(expectedYs)
+  })
+
+  it('throws an error if no tokenizer is specified', async () => {
+    const invalidTask = initMockTask()
+    invalidTask['trainingInformation']['tokenizer'] = undefined;
```
tharvik:
```suggestion
    invalidTask.trainingInformation.tokenizer = undefined;
```

discojs/discojs-core/src/dataset/data/preprocessing/text_preprocessing.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+    try {
+      await tokenize.apply(Promise.resolve("input text doesn't matter"), invalidTask)
+    } catch (e) {
+      console.error(e)
```
tharvik: do not show error on correct path

```suggestion
```

discojs/discojs-core/src/dataset/data/preprocessing/text_preprocessing.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+    }
+    throw new Error("undefined tokenizer should have thrown an error")
+  })
+  it('throws an error if the tokenizer name is invalid', async () => {
+    const invalidTask = initMockTask()
+    invalidTask['trainingInformation']['tokenizer'] = 'invalid-tokenizer-name'
+    try {
+      await tokenize.apply(Promise.resolve("input text doesn't matter"), invalidTask)
+    } catch (e) {
+      console.error(e)
```
tharvik:
```suggestion
```

discojs/discojs-core/src/dataset/data/preprocessing/text_preprocessing.ts
```diff
@@ -14,43 +14,58 @@ export enum TextPreprocessing {
...
+ * https://github.com/xenova/transformers.js/blob/8804c36591d11d8456788d1bb4b16489121b3be2/src/tokenizers.js#L2517
  */
 const leftPadding: PreprocessingFunction = {
   type: TextPreprocessing.LeftPadding,
   apply: async (x: Promise<tf.TensorContainer>, task: Task): Promise<tf.TensorContainer> => {
-    let { xs } = await x as TokenizedEntry
-    if (xs === undefined ||  !(xs instanceof tf.tensor) ||xs.rankType !== tf.Rank.R1) {
-      new Error("The leftPadding preprocessing expects a 1D tensor named 'xs' as input")
+    if (x === undefined || !Array.isArray(x) || x.length == 0 || typeof(x[0] != 'number')) {
+      new Error("The leftPadding preprocessing expects a non empty 1D array of number")
```
tharvik, line 36: does nothing as `x` is a Promise and do not throw created Error.

discojs/discojs-core/src/models/gpt/config.ts
```diff
@@ -50,7 +49,16 @@ export const DEFAULT_CONFIG: Required<GPTConfig> = {
...
   embdDrop: 0.2,
   tokEmb: true,
-  lmHead: true
+  lmHead: true,
+  nLayer: 3,
+  nHead: 3,
+  nEmbd: 48,
+}
+
+export type ModelSize = {
```
tharvik, line 58: it's equivalent but nicer to use `interface` when creating a `type` for an object.

```suggestion
export interface ModelSize {
```

discojs/discojs-core/src/models/gpt/gpt.spec.ts
```diff
@@ -0,0 +1,44 @@
...
+      const ys = tf.oneHot(tokens.slice(1), tokenizer.model.vocab.length + 1)
+      const xs = tf.tensor(tokens.slice(0, config.blockSize), undefined, 'int32')
+      return {xs, ys}
+    }).repeat().batch(64)
+
+    const model = new GPT(config)
+    const logGenerator = model.train(tokenDataset, undefined, 5) // 5 epochs
+    for await (const _ of logGenerator); // Await the end of training
+    const generation = await model.generate("Lorem ipsum dolor", tokenizer, 1)
+    console.log(generation)
```
tharvik, line 41:
```suggestion
```

discojs/discojs-core/src/models/gpt/index.ts
```diff
@@ -73,6 +38,7 @@ export class GPT extends Model {
     epochs = 1,
     tracker = new Sink()
   ): AsyncGenerator<EpochLogs, void> {
+    this.model.compile() // init the optimizer
```
tharvik: move it to the constructor, otherwise, it will re-init the optimizer every time someone wants to train (seems weird to me).

discojs/discojs-core/src/models/gpt/index.ts
```diff
@@ -91,6 +57,10 @@ export class GPT extends Model {
     for (let i = 0; i < epochs; i++) {
       await this.model.fitDataset(trainingData, trainingArgs)
       yield logs
+      if (logs !== undefined && logs.training_loss < 0.01) {
+        console.log("Early stopping")
+        break
+      }
```
tharvik: let's users control when to stop, maybe they want even less training_loss, or stop it for wathever reasons.
when using a generator, one [can stop it using `generator.return()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Generator/return), or simply never read the next value (will resolve it's promise still but not advance anymore).

```suggestion
```

discojs/discojs-core/src/serialization/model.ts
```diff
@@ -22,45 +23,57 @@ export async function encode (model: Model): Promise<Encoded> {
...
+      }
       // TODO totally unsafe casting
       return await models.TFJS.deserialize(rawModel as tf.io.ModelArtifacts)
-    case Type.GPT: {
+    case Type.GPT: {  
+      let config
+      if (raw.length == 2) {
+        config = undefined
+      } else if (raw.length == 3) {
+        config = raw[2] as GPTConfig
```
tharvik, line 60: you don't check that it is indeed a valid `GPTConfig`. you can put a function `isGPTConfig` in `models/gpt/config` to check for that, inspired of the various checkers in `tasks/*`

discojs/discojs-core/src/models/gpt/index.ts
```diff
@@ -117,13 +87,32 @@ export class GPT extends Model {
...
+      weights: this.weights,
+      config: this.config
+    }
   }
 }
+
+export type GPTSerialization = {
+  weights: WeightsContainer
+  config?: GPTConfig
+}
```
tharvik, line 126: no need to export it, it don't see it being used elsewhere. or maybe you wanted to use it `serialization.model`?

discojs/discojs-core/src/task/training_information.ts
```diff
@@ -16,8 +16,8 @@ export interface TrainingInformation {
   batchSize: number
   // preprocessingFunctions: preprocessing functions such as resize and normalize
   preprocessingFunctions?: Preprocessing[]
-  // dataType, e.g. image or tabular
-  dataType: string
+  // dataType, 'image', 'tabular' or 'text'
+  dataType: 'image' | 'tabular' | 'text'
```
tharvik, line 20: :raised_hands:

discojs/discojs-node/src/models/model_loader.ts
```diff
@@ -0,0 +1,23 @@
+import fs from 'node:fs'
+import fsPromises from 'node:fs/promises'
```
tharvik: async all the way! there is a bunch of [TOCTOU](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use) to otherwise.

```suggestion
import fs from 'node:fs/promises'
```

discojs/discojs-node/src/models/model_loader.ts
```diff
@@ -0,0 +1,23 @@
...
+import { serialization, models } from '@epfml/discojs-core'
+
+export async function saveModelToDisk(model: models.Model, modelFolder: string, modelFileName: string): Promise<void> {
+  try {
+    if (!fs.existsSync(modelFolder)) {
+      fs.mkdirSync(modelFolder)
+    }
+  } catch (err) {
+    console.error(err);
+  }
```
tharvik: let it fail (it would anyway when writing the file but it's more precise here)

```suggestion
  if (!fs.existsSync(modelFolder)) {
    fs.mkdirSync(modelFolder)
  }
```

docs/examples/wikitext.ts
```diff
@@ -17,56 +15,54 @@ async function main(): Promise<void> {
...
-  /**
-   * The commented code below shows how to load an existing model
-   */
-
-  // const content = await fsPromises.readFile(`${modelFolder}/model.json`)
-  // model = await serialization.model.decode(content) as models.GPT
-  // console.log(await model.generate(sample, tokenizerName, 10))
+  // Toggle TRAIN_MODEL to either train and save a new model from scratch or load an existing model
+  const TRAIN_MODEL = true
+  if (TRAIN_MODEL) {
```
tharvik, line 24: consider having a common code after this block, as single way to use the model, that makes it clearer IMO

docs/examples/wikitext.ts
```diff
@@ -17,56 +15,54 @@ async function main(): Promise<void> {
...
+  }
 }
 
 async function loadWikitextData (task: Task): Promise<data.DataSplit> {
   const loader = new NodeTextLoader(task)
   const dataSplit: data.DataSplit = {
-    train: await data.TextData.init(await loader.load('../../datasets/wikitext/wiki.train.tokens', {shuffle: false}), task),
-    validation: await data.TextData.init(await loader.load('../../datasets/wikitext/wiki.valid.tokens', {shuffle: false}), task)
+    train: await data.TextData.init(await loader.load('../../datasets/wikitext/wiki.train.tokens', {shuffle: true}), task),
+    validation: await data.TextData.init(await loader.load('../../datasets/wikitext/wiki.valid.tokens', {shuffle: true}), task)
```
tharvik, line 58: it's shuffling by default
```suggestion
    train: await data.TextData.init(await loader.load('../../datasets/wikitext/wiki.train.tokens'), task),
    validation: await data.TextData.init(await loader.load('../../datasets/wikitext/wiki.valid.tokens'), task)
```

tharvik (approved):
wouhou, we've a talking model, thanks!

mostly cosmetics changes, except for `models/gpt` changes

discojs/discojs-core/src/models/gpt/config.ts
JulienVig: I was actually searching for pros and cons and didn't find anything significant, why do you prefer interface?

docs/examples/wikitext.ts
JulienVig: I want to be make it explicit in the example, I found out myself way too late that shuffling was happening behind the scene

docs/examples/wikitext.ts
JulienVig: What do you mean? Like reverting to when loading a model was commented code rather than having a boolean constant?

discojs/discojs-core/src/models/gpt/config.ts
tharvik: hum, it's a habit mostly, I got it from [one of eslint rule](https://typescript-eslint.io/rules/consistent-type-definitions). now that I reread the documentation, there [aren't much differences](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#differences-between-type-aliases-and-interfaces). the only plus for interface that I found myself agreeing with it that [it shows better error messages]( https://www.typescriptlang.org/play#example/types-vs-interfaces)

docs/examples/wikitext.ts
tharvik: ho right, I wrote this comment quite in a hurry, let me reformulate:
it would be nice to show that training the model is optional. having a boolean is already a good way to do it, and to push it a bit further, having the same body of code at the end on both paths would be even nicer. smth like the following
```ts
// optionnaly train the model if wanted
if (TRAIN_MODEL) { ... }

// use saved model 
const model = ...
const prompt = ...
```

JulienVig: Ah yes yes absolutely

PR #659, opened 2024-04-16
Benchmark GPT-tfjs

### LINKED ISSUE #657: Benchmark gpt-tfjs integration
Benchmark and report some information on what to expect and the row for improvement:
- [ ] number of training iterations needed on an example dataset, expected loss/perplexity, example of text generation after training
- [x] number of tokens/sec
- [x] model sizes
- [x] memory requirements
- [ ] compare numbers with ONNX

#656 should be addressed first

JulienVig:
* CLI benchmark of gpt-tfjs. Closes #657
* Fixes a bunch of issues as I uncover them, memory leaks, training loop end condition, clean the GPT code.
* Bumps vue-tsc version

## Training
Benchmark on a 2022 MacBook Air M2 with 16GB of RAM. 
To reproduce, check out [58f018f](https://github.com/epfml/disco/pull/659/commits/58f018f07fb0c61a28fd9bd9a7ff403e06fbfd3a) and run `npm -w cli run benchmark_gpt -- --contextLength 128 --batchSize 8` for example.

Time per token is obtained by measuring the time of 10 training update iterations and diving by (batch size * context length)
Memory values are the max memory allocated between the attention mechanism and the memory after computing the gradients. So far, the attention mechanism always had higher memory requirements. The actual peak memory allocated during training may be different but tfjs doesn't let us get this information easily.

I leave empty `-` cells where I deemed the benchmark too slow to perform. If needed, missing values can be extrapolated.

`gpt-nano`:
- 2.50M parameters
- number of layers: 3
- number of heads: 3
- embedding dimension: 48

| `gpt-nano`  | `batch_size=8` | `batch_size=16` | `batch_size=32` | `batch_size=64` |
| -| - | -| -| -|
| `context_length=128`  | 0.53 ms/token <br> 0.33 GB  | 0.60 ms/token <br> 0.56 GB  | 0.82 ms/token <br> 1.12 GB  | 1 ms/token <br> 2.18 GB  |
| `context_length=256`  | 0.53 ms/token <br> 0.64 GB  | 0.68 ms/token <br> 1.22 GB  | 1.06 ms/token <br> 2.36 GB  | 1.81 ms/token <br> 4.66 GB  |
| `context_length=512`  | 0.72 ms/token <br> 1.42 GB | 1.08ms/token <br> 2.75 GB | 2 ms/token <br> 5.42 GB  | -  |
| `context_length=1024`  | 1.24 ms/token <br> 3.56 GB | 2.10 ms/token <br> 6.98 GB | -  | -  |
| `context_length=2048`  | 2.47 ms/token <br> 10.2 GB | - | -  | -  |

<br>

`gpt-micro`:
- 7.23M parameters
- number of layers: 4
- number of heads: 4
- embedding dimension: 128

| `gpt-micro`  | `batch_size=8` | `batch_size=16` | `batch_size=32` |
| -| - | -| - |
| `context_length=128`  | 0.84 ms/token <br> 0.6 GB  | 0.77 ms/token <br> 1 GB  | 1.08 ms/token <br> 1.86 GB  |
| `context_length=256`  | 0.77 ms/token <br> 1.1 GB  | 1.07 ms/token <br> 2 GB  | 1.57 ms/token <br> 3.8 GB  |
| `context_length=512`  | 1.09 ms/token <br> 2.3 GB  | 2.12 ms/token <br> 4.4 GB  | -  |
| `context_length=1024`  | 2.11 ms/token <br> 5.8 GB  | -  | -  |

<br>

`gpt-mini`:
- 12.32M
- number of layers: 6
- number of heads: 6
- embedding dimension: 192

| `gpt-mini`  | `batch_size=8` | `batch_size=16` | 
| -| - | -| 
| `context_length=128`  |  1.3 ms/token <br> 1 GB  | 1.12 ms/token <br> 1.75 GB  |
| `context_length=256`  | 1.17 ms/token <br> 1.9 GB  | 1.43 ms/token <br> 3.5 GB  |

<br/>

`gpt2`:
- number of layers: 12
- number of heads: 12
- embedding dimension: 768

| `gpt2`  | `batch_size=8` | 
| -| - | 
| `context_length=128`  |  7.56 ms/token <br> 7.7 GB  | 
| `context_length=256`  | 6.83 ms/token <br> 12.7 GB  | 

### Comparisons
Using the Python [nanoGPT benchmark script](https://github.com/karpathy/nanoGPT?tab=readme-ov-file#efficiency-notes) on the same machine, I get the following comparisons between Python and JS:

| `gpt-nano` | `gpt-tfjs` | `python` (nanoGPT repo) |
| -| - | -|
|`batch size=8` and `context_length=128` | 0.53 ms/tokens | 0.17 ms/tokens |
|`batch size=32` and `context_length=512` | 2 ms/tokens | 0.29 ms/tokens |

## Inference

Run `npm -w cli run benchmark_gpt -- --inference --modelPath <path to trained model>`
For `gpt-nano` trained with context length 128, **inference time averages between 6 and 8 ms/token**.

WebGPT reports 3 ms/token at 5M parameters, which is between gpt-nano (2.5M) and gpt-micro (7.2M). They also managed to scale up to 1.5B parameters on a M1 Mac with WebGPU.

### DIFF
--- a/discojs/discojs-core/src/models/gpt/model.ts
+++ b/discojs/discojs-core/src/models/gpt/model.ts
@@ -72,7 +72,6 @@ class GPTModel extends tf.LayersModel {
     
     for (let epoch = 1; epoch <= trainingArgs.epochs; epoch++) {
       let averageLoss = 0
-      let averageWeightUpdateTime = 0
       let iteration = 1
       const iterator = await dataset.iterator()
       let preprocessingTime = performance.now()
@@ -105,7 +104,6 @@ class GPTModel extends tf.LayersModel {
         const loss = await lossTensor.array()
         averageLoss += loss
         weightUpdateTime = performance.now() - weightUpdateTime
-        averageWeightUpdateTime += weightUpdateTime
         // Probably never the case. Empirically the attention mechanism always allocates
         // more memory than the backward pass
         if (backwardPassMemory > this.peakMemory.value) {
@@ -203,7 +201,6 @@ export class GPTForCausalLM extends GPTModel {
       const idxNew = idx.concat(idxNext, 1)
       tf.dispose(idx)
       idx = idxNew
-      const idxNextArr = await idxNext.array()
       tf.dispose(idxNext)
     }
     const idxArr = await idx.array()
--- a/discojs/discojs-core/src/models/gpt/model.ts
+++ b/discojs/discojs-core/src/models/gpt/model.ts
@@ -24,18 +24,27 @@ export declare abstract class Dataset<T> {
  */
 class GPTModel extends tf.LayersModel {
   protected readonly config: Required<GPTConfig>
+  private readonly disposalRefs: Array<() => void>
 
   constructor(partialConfig?: GPTConfig) {
-    // Complete missing config parameters with default values
+    // Fill missing config parameters with default values
     let completeConfig: Required<GPTConfig> = { ...DEFAULT_CONFIG, ...partialConfig }
     // Add layer sizes depending on which model has been specified
     completeConfig = { ...completeConfig, ...getModelSizes(completeConfig.modelType) }
 
     // Init the tf.LayersModel and assign it to this
-    const gpt = GPTArchitecture(completeConfig)
+    const disposalRefs:Array<() => void> = []
+    const gpt = GPTArchitecture(completeConfig, disposalRefs)
     const { inputs, outputs, name } = gpt
     super({ inputs, outputs, name })
     this.config = completeConfig
+    this.disposalRefs = disposalRefs
+  }
+
+  disposeRefs() {
+    for (let disposeFn of this.disposalRefs) {
+      disposeFn()
+    }
   }
 
   get getGPTConfig() {
@@ -96,16 +105,16 @@ class GPTModel extends tf.LayersModel {
         averageMemory += peakMemory
         tf.dispose([xs, ys, lossTensor, next.value])
 
-        console.log(
-          `Epoch: ${epoch}`,
-          `\tStep: ${iteration} / ${this.config.maxIter}`,
-          `\tLoss: ${loss.toFixed(3)}`,
-          `\tPeak memory: ${peakMemory.toFixed(2)} MB`,
-          `\tMemory: ${(tf.memory().numBytes / 1024 / 1024).toFixed(2)} MB`,
-          `\tNumber of tensors allocated: ${tf.memory().numTensors}`,
-          `\tPreprocessing time: ${preprocessingTime.toFixed(0)} ms`,
-          `\tWeight update time: ${weightUpdateTime.toFixed(0)} ms`
-        )
+        // console.log(
+        //   `Epoch: ${epoch}`,
+        //   `\tStep: ${iteration} / ${this.config.maxIter}`,
+        //   `\tLoss: ${loss.toFixed(3)}`,
+        //   `\tPeak memory: ${peakMemory.toFixed(2)} MB`,
+        //   `\tMemory: ${(tf.memory().numBytes / 1024 / 1024).toFixed(2)} MB`,
+        //   `\tNumber of tensors allocated: ${tf.memory().numTensors}`,
+        //   `\tPreprocessing time: ${preprocessingTime.toFixed(0)} ms`,
+        //   `\tWeight update time: ${weightUpdateTime.toFixed(0)} ms`
+        // )
 
         if (evalDataset !== undefined && this.config.evaluateEvery !== undefined
           && iteration % this.config.evaluateEvery == 0) {
--- a/discojs/discojs-core/src/models/gpt/model.ts
+++ b/discojs/discojs-core/src/models/gpt/model.ts
@@ -68,8 +68,8 @@ class GPTModel extends tf.LayersModel {
   async fitDataset<T>(dataset: Dataset<T>, trainingArgs: tf.ModelFitDatasetArgs<T>): Promise<tf.History> {
     const callbacks = trainingArgs.callbacks as tf.CustomCallbackArgs
     const evalDataset = trainingArgs.validationData as tf.data.Dataset<{ xs: tf.Tensor2D, ys: tf.Tensor3D }>
-    console.log(`Begin train - Memory: ${(tf.memory().numBytes / 1024 / 1024).toFixed(2)} MB`, `Num tensors: ${tf.memory().numTensors}`)
     await callbacks.onTrainBegin?.()
+    
     for (let epoch = 1; epoch <= trainingArgs.epochs; epoch++) {
       let averageLoss = 0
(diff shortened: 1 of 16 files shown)


martinjaggi:
thanks, very interesting!
would you have a just very rough reference value also for the python equivalent (nanoGPT python)?

JulienVig:
@tharvik I'd be curious to hear your opinion on a few things:
1. Where do you think we should report the benchmark? I was thinking of all reporting them in this PR and linking it where relevant (e.g. in gpt/config.ts or the GPT class docstring)
2. Benchmarking performance needs modifying the gpt source code to keep track of memory, do you think it's worth keeping around or leave the benchmark on this branch and not merge it?

discojs/discojs-core/src/models/index.ts
```diff
@@ -1,6 +1,6 @@
 export { EpochLogs, Model } from './model.js'
 
 export { GPT } from './gpt/index.js'
-export { GPTConfig } from './gpt/config.js'
+export { GPTConfig, GPTModelType } from './gpt/config.js'
```
tharvik: it is exposed via `GPTConfig["modelType"]`, no need to expose it directly.

discojs/discojs-core/src/models/model.ts
```diff
@@ -13,6 +13,8 @@ export interface EpochLogs {
     loss: number,
     accuracy: number
   };
+  weightUpdateTime?: number;
+  peakMemory?: number;
```
tharvik: I do prefer to have the minimum amount of fields as `undefined`, that's cumbersome to work with
* `weightUpdateTime` is pretty much equals to the time of an epoch, no? so one can measure it themself instead of make us compute it
* `peakMemory` can be computed on every model, and at every epoch, let's do that rather than having it undefined

docs/examples/benchmark_gpt.ts
tharvik, line 1: consider moving the whole benchmark to the CLI: examples are only entrypoints into disco, not a full benchmark

discojs/discojs-core/src/models/model.ts
```diff
@@ -50,4 +52,7 @@ export abstract class Model {
   /** Predict likely values */
   // TODO extract in separated TrainedModel?
   abstract predict(input: Sample): Promise<Prediction>;
+
+  // Cleanup the memory occupied by the model
+  abstract dispose(): void;
```
tharvik: weird to have to do our own memory management in a GC language. sadly, per upstream, it [is still needed](https://github.com/tensorflow/tfjs/issues/4864). but I just found out about an [ECMAScript proposal for this pattern](https://github.com/tc39/proposal-explicit-resource-management) which [is already implemented in typescript](https://www.typescriptlang.org/docs/handbook/variable-declarations.html#using-declarations)! consider using that instead.

docs/examples/benchmark_gpt.ts
```diff
@@ -0,0 +1,90 @@
...
+      modelType: 'gpt-nano',
+      lr: 0.0001,
+      maxIter: 10,
+      evaluateEvery:10000,
+      maxEvalBatches: 10,
+      blockSize: 8,
+      vocabSize: 50258
+    }
+    
+    const modelType = 'gpt-nano'//['gpt-nano', 'gpt-micro', 'gpt-mini', 'gpt2']
```
tharvik: I think you exposed `GPTModelType` to type that correctly. you can use [`as const`](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#const-assertions) to tell the compiler to not type the variable as `string` but as `"gpt-nano"`
```suggestion
    const modelType = 'gpt-nano' as const //['gpt-nano', 'gpt-micro', 'gpt-mini', 'gpt2']
```

docs/examples/benchmark_gpt.ts
```diff
@@ -0,0 +1,90 @@
...
+      const msPerToken = updateTime / batchSize / contextLength  
+      console.log(`\t\t\t${msPerToken.toFixed(2)} ms/token <br> ${logs.peakMemory?.toFixed(2)} GB`)
+    }
+    model.dispose()
+
+    // Check for memory leak. Currently, there are a few tensors that are still not disposed (one per attention layer in the model)
+    console.log(`End loop - Memory: ${(tf.memory().numBytes / 1024 / 1024).toFixed(2)} MB`, `Num tensors: ${tf.memory().numTensors}`)
+  } else {
+    
+    const model = await loadModelFromDisk(`models/model_random.json`) as models.GPT
```
tharvik: better to throw an error than casting

discojs/discojs-core/src/models/gpt/model.ts
```diff
@@ -184,17 +217,17 @@ export class GPTForCausalLM extends GPTModel {
 
   private generateOnce (model: tf.LayersModel, idx: tf.Tensor2D, config: GenerateConfig): { idxNext: tf.Tensor2D, timePerToken: number } {
```
tharvik: would measuring `timePerToken` from the outside make sense?

tharvik (commented):
superbe! that's very nice to have metrics on what's we're doing, thanks!

> 1. Where do you think we should report the benchmark? I was thinking of all reporting them in this PR and linking it where relevant (e.g. in gpt/config.ts or the GPT class docstring)

the nicest thing would be to be able to generate such metrics via the CLI, with the example output of the cmd being what you've here (not the tables I mean, but the same content).

> 2. Benchmarking performance needs modifying the gpt source code to keep track of memory, do you think it's worth keeping around or leave the benchmark on this branch and not merge it?

not merging it means that it'll slowly drift off. I think adding the memory usage to the `EpochLogs` as you did is the way to go. see my comments related to it.

docs/examples/benchmark_gpt.ts
JulienVig: Great! thank you for the documentation, I've been searching how to circumvent this issue for a while

docs/examples/benchmark_gpt.ts
JulienVig: Thanks for all the typescript advices I really appreciate them!

discojs/discojs-core/src/models/model.ts
JulienVig: Measuring the whole epoch adds an overhead but it's not that much

discojs/discojs-core/src/models/gpt/model.ts
JulienVig: Yes I'm moving all the benchmark-related logic at the highest level

tharvik (approved):

JulienVig:
The memory values needs to be slightly updated when #807 is merged

martinjaggi:
up or down? ;)

JulienVig:
A very superficial benchmark showed 10-20% decrease in memory usage!

PR #660, opened 2024-04-23
LUS COVID demo

JulienVig:
Bug fixes to set up the LUS COVID demo for next week

- [x] Add test cases lus_covid (image_loader, validator, end to end training)
- [x] Fix node image_loader default channel error (default to 1 while it should be 3)
- [x] Add lus_covid task to CLI
- [x] Fine-tune hyper parameters
- [ ] Image dataset shuffling is done by hand while tfjs already implements it
- [x] Fix label encoding mistake (using tf.oneHot on 'positive' and 'negative' yields the same label for both class)
- [x] "Train Collaboratively" raises an error
- [x] Training loss and other metrics stay constant after the first few epochs
- [x] Model always predicts positive
- [x] ~~Training collaboratively with two web-client results in timeouts (server drop client contributions)~~
    - Addressed by #661 
- [x] ~~UI number of participants doesn't get updated as they join a task~~
    - Addressed by #661 
- [x] web-client training performance stays constant while training via node.js works well

Model used to only predict the majority class (~0.5 accuracy), by normalizing image values from 0-255 to 0-1 the training now yields >0.9 on both training and validation sets

### DIFF
--- a/server/src/tasks.ts
+++ b/server/src/tasks.ts
@@ -38,36 +38,33 @@ export class TasksAndModels {
     const modelPath = `./models/${discoTask.id}/`
     try {
       const content = await fs.readFile(`${modelPath}/model.json`)
-      console.log("Loading local model for task", discoTask.id)
       return await serialization.model.decode(content)
     } catch {
-      throw new Error("No local model found")
       // unable to read file, continuing
     }
-    // console.log("Instantiating a new model", discoTask.id)
     
-    // if (isTask(task)) {
-    //   throw new Error('saved model not found and no way to get it')
-    // } else {
-    //   model = await task.getModel()
-    // }
-    // console.log('Saving model to disk')
-    // await fs.mkdir(modelPath, { recursive: true })
-    // const encoded = await serialization.model.encode(model)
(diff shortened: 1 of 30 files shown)


server/src/tasks.ts
```diff
@@ -31,40 +31,43 @@ export class TasksAndModels {
...
+    //   try {
+    //     await this.checkDigest(discoTask.digest, modelPath)
+    //   } catch (e) {
+    //     console.warn('removing model files at', modelPath)
+    //     await fs.rm(modelPath, { recursive: true, force: true })
+    //     throw e
+    //   }
+    // }
+    
+    // return model
```
tharvik: avoid commented code

```suggestion
```

tharvik (approved):
génial, thanks for the hard work!

server/src/tasks.ts
JulienVig: Thanks! I think you reviewed an older commit? The code should be uncommented in the latest

tharvik: ho right, that was an old commit; nvm then

martinjaggi:
super nice, congrats!

PR #661, opened 2024-04-25
small lus covid fixes

tharvik:
some fixes for making the COVID lungs task work

### DIFF
--- a/datasets/.gitignore
+++ b/datasets/.gitignore
@@ -12,6 +12,9 @@
 /titanic_wrong_number_columns.csv
 /titanic_wrong_passengerID.csv
 
+# lungs ultrasound
+/lus_covid/
+
 # wikitext
 /wikitext/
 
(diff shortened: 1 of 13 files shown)


datasets/.gitignore
```diff
@@ -12,6 +12,9 @@
 /titanic_wrong_number_columns.csv
 /titanic_wrong_passengerID.csv
 
+# lungs ultrasound
+/lungs ultrasound/
```
JulienVig:
```suggestion
/lus_covid/
```
To keep it similar to the task name and not use whitespace in the file system :)

datasets/populate
```diff
@@ -8,8 +8,16 @@ cd "$(dirname "$0")"
 curl 'http://deai-313515.appspot.com.storage.googleapis.com/example_training_data.tar.gz' |
 	tar --extract --strip-components=1
 
+# lungs ultrasound
+curl 'https://drive.switch.ch/index.php/s/zM5ZrUWK3taaIly/download' > archive.zip
+unzip -u archive.zip
+rm -rf 'lungs ultrasound'
+mv DeAI-testimages 'lungs ultrasound'
```
JulienVig:
```suggestion
mv DeAI-testimages 'lus_covid'
```

Issue #665, opened 2024-05-03
inform users on where the dataset is kept


tharvik:
in the webapp, when adding files, users might think that they are uploading data to the server, that it is not kept locally. labelling the button stating that would help ("connecting data" or "direct the model to your data"), with a tip/long description explaing it more ("Your data always stays on your computer. Only the AI model is downloaded and uploaded. Don't trust us? Try linking a large file and check your network usage with Task Manager on Windows or Activity Monitor on Mac"). better even would be to be explicit on the training process, such as having an animation or progress bar showing "downloading model from server", "training model in browser", "sharing model to server".

animation idea from @EduardDurech:
> we can show the server node and maybe even where it's located for transparency (e.g. All of them would say "EPFML GitHub" at the moment), then the amount of users would be connected with a pipe to a cloud or something saying the amount of users instead of the textbox there is now, then there would be the user node which would show an animation of files going into it, staying there, an animation when the model is fetched from server to user node, animation for the model being trained and updated within the user node, then the upload of the model to server, all while the file icon is in the user node

JulienVig:
I very much agree with that, the current data form really looks like users are uploading their data. We could also add something like a lock icon and a mention of "Secure Form" (and make the form secure :) ) and something showing that their data has not been uploaded anywhere

EduardDurech:
Maybe also:

> Most won't be technical enough to check the code, but we could also hyperlink the parts on the repo that do the data linking when tfjs accesses data

Also maybe a link to a "How does it work" where we have a small explanation on federated learning and decentralized learning, or link to the information tab and have more explanation on the first information page

hleepfl:
Continued in #1137

Issue #662, opened 2024-04-29
Pre-defined task improvements


JulienVig:
- [x] Training when a saved model already exists reuses it by default, while it should create a new model overwrite it.
- [x] The training graphs should show the training and validation loss too, sometimes the accuracies stay constant while the loss still changes.
- [x] The test accuracy graph doesn't update and stays at 0 -> #678 removed the test line plot (because it was only plotting a single scalar value)
- [x] The testing an image model (e.g. lus_covid) shows what are the predictions for each image and where the model made classification mistakes (groundtruth vs prediction). However errors are sometimes not displayed (only prediction).
- [ ]  Validation set sometimes contains only one class
    Image tasks support choosing a train-validation split in which a percentage of the training set is set apart as a validation set (`discojs-core/src/dataset/data_loader/image_loader.ts`). In some cases however, e.g., when using imbalanced or small datasets, this can result in some classes not being included in the validation set at all.
Unfortunately, it seems that tensorflow.js doesn't offer any function for stratified splits.
- [x] Cached model is used by default while it is said in the task that it would be overwritten. Moreover the toggle button to disable this isn't displayed.

Issue #628, opened 2024-02-13
Fix and improve UI


JulienVig:
After reviewing the DISCO website as a new user, here is a list of things that can be fixed and improved. This issue goes in pair with [#627](https://github.com/epfml/disco/issues/627) which addresses higher-level concerns with the website. Some changes made after #627 may render some items in this issue obsolete and both issues should be considered at the same time.

### Fixes
- [ ] On Safari, hovering over (any) buttons seems makes the button border bug (I noticed it does bug if I approach the cursor quickly). On the image below I'm hovering on the Next button. Works well on Firefox.
![Image](https://github.com/epfml/disco/assets/33122365/8f026408-9540-459a-a01f-6dd6099799db)    
- [ ] On Safari, hovering over (any) paragraphs creates a square border while the paragraphs have rounded corners (pay attention to the corners on the image. Works well on Firefox.
![Image](https://github.com/epfml/disco/assets/33122365/2ac01f48-e02e-449e-9c35-b6e635be3ffa)
- [x] Hovering over the top-left disco logo creates an empty text bubble.
When hovering:
![Image](https://github.com/epfml/disco/assets/33122365/3c17ef46-abab-4ab8-9201-cce69fcd08f7)
Not hovering:
![Image](https://github.com/epfml/disco/assets/33122365/57f24a00-22d4-49f8-ba10-d3f3a4ef5235)
### Features
- [x] Make sure all links open in new tab, especially links to download datasets
- [x] The website remembers at which step you were but sometimes that should be reset. An example would be clearer: Going to Tasks, Joining the Titanic task and clicking next until the Train your Model step gets you to a particular step. Now if you use the sidebar to go back to the Tasks page, and click again on Joining the Titanic task brings you back to the Train your Model step and not the first step (Task Description). This is great but it should not always behave this way. For example, download two pre-trained models from the Evaluate page, let's say Titanic and MNIST. From the model library click on Titanic to Test it, that leads you to the Connect Your Data step. Now from the homepage click on Evaluate. That directly leads you to Connect Your Data step, implying that you are currently testing Titanic while the expected behavior is to let the user choose which model they want to evaluate (since we clicked on the noon-specific Evaluate button).
- [x] I think the Previous and Next buttons should be at the bottom of pages rather than top, to follow the user flow who fills information or reads info from top to bottom. Right now users need to scroll up after every page to get to the next step.

### Fixes
- [ ] update GIF and logo quality, it is pixelated

### Features

[...]

Issue #627, opened 2024-02-13
Refactor the UI to be more user-friendly


JulienVig:
This issue addresses the overall web app structure, taking a step back from implementation details to look at the website's objectives and user experience. The following contains reflections and to-dos stemming from my personal experience navigating the DISCO website without any prior knowledge. Some items are questions rather than to-dos to emphasize what may be unclear or confusing to new users.

TLDR:
1. Restructure the website navigation
2. Make showcase training (preexisting tasks) much faster (minimize the number of clicks a new user has to make between landing on the homepage and having evaluating a model)
3. Make the custom training (creating and joining custom tasks) the center of the website since this is what most users will use DISCO for.

## Current problems

* The navigation is confusing and sometimes unintuitive
* What are preexisting tasks is unclear at first. They are examples, and that should be immediately obvious.
* Preexisting tasks take too much "effort" to train a model (i.e. clicks, navigate pages, upload datasets) 
* The website is oriented towards preexisting tasks while it should be oriented towards custom training.
* What's the difference between a DISCOllaborative, a task and a custom task?
* How do you connect to an existing DISCOllaborative?

### Refactor the navigation

This is the main item of this issue. The website navigation isn't clear for newcomers, it is hard to realize there are different ways with different names to get to the same page (for example, clicking on Train from the homepage gets you to the Preexisting task list, which you can also reach from the sidebar menu). Restructure the website to make navigation coherent, intuitive and centered around custom training:
- [x] Homepage button should be more intuitive, e.g. "Join a DISCOllaborative to train a model" is misleading. It actually points to preexisting tasks as examples while a user may understand it as "train your own model"
- [x]  "Evaluate your DISCO-trained model" should also specify that users can download pre-trained model
- [x] Sidebar menu "Task" tab should be renamed Examples
- [x] Custom training should be in the sidebar menu too

### Rework preexisting tasks

The purpose of preexisting tasks is unclear. Despite what the home page says, it is not for training your model but to showcase how DISCO works. When I first got on the website, the intuitive navigation flow of the website led me to preexisting tasks. Given that I want to train my own model with my own data, I was confused at why I couldn't upload a model and couldn't change the hyperparameters. Why do I have to upload data if it's a preexisting task?

- [x] Rename preexisting tasks as Examples
- [ ] Make preexisting tasks much easier: minimize the number of clicks a new user has to make between landing on the homepage and having evaluating a model. 
[...]

PR #668, opened 2024-05-03
Improves and fixes the web-client UI

JulienVig:
Addresses parts of #627 #628, #662, #665 and various things noted while working on the LUS COVID demo

- [x] Rework website navigation to make it more intuitive. Home page now links to tutorial information, existing tasks and custom task creation. The sidebar links to the home page, existing tasks, task creation, model evaluation, model library and the tutorial information. Settings has been merged with model library (since it only had the model library toggle button)
- [x] Add error handling when selected data (e.g. wrong CSV header, image file not found in CSV labels etc) 
- [x] Removed pre-defined tasks geotags and skin mnist because no data is provided and training has not been not tested
- [x] Harmonize the outdated tutorial information UI with the rest of the website.
- [x] Remove l18n internationalization plugin, only english is currently used and there are no plans to add new languages.

* Rework ModelCaching, current behavior now explicit and card moved to training page

* Improve the training charts with loss plots, tooltip, and dynamic y axis range

* Add an explanation on what pre-defined "DISCOllaboratives" are
 

* #665 Specifies that data is not uploaded anywhere

Addresses #627 
Addresses #628 
Addresses #662 
Addresses #665


discojs/discojs-core/src/default_tasks/cifar10.ts
```diff
@@ -17,7 +17,8 @@ export const cifar10: TaskProvider = {
         tradeoffs: 'Training success strongly depends on label distribution',
         dataFormatInformation: 'Images should be of .png format and of size 32x32. <br> The label file should be .csv, where each row contains a file_name, class.  <br> <br> e.g. if you have images: 0.png (of a frog) and 1.png (of a car) <br> labels.csv contains: (Note that no header is needed)<br> 0.png, frog <br> 1.png, car',
         dataExampleText: 'Below you can find 10 random examples from each of the 10 classes in the dataset.',
-        dataExampleImage: 'https://storage.googleapis.com/deai-313515.appspot.com/example_training_data/cifar10-example.png'
+        dataExampleImage: 'https://storage.googleapis.com/deai-313515.appspot.com/example_training_data/cifar10-example.png',
+        sampleDatasetLink: 'https://www.kaggle.com/competitions/cifar-10/data'
```
tharvik, line 19: consider adding that to `datasets/populate` instead of pulling from our own archive

discojs/discojs-core/src/task/display_information.ts
```diff
@@ -14,7 +14,10 @@ export interface DisplayInformation {
   dataExample?: DataExample[]
   // TODO no need for undefined
   headers?: string[]
+  // Displays the image at this URL in the UI as an example when connecting data
   dataExampleImage?: string
+  // URL to download a dataset for the task, is displayed in the UI when asking to connect data
+  sampleDatasetLink?: string
```
tharvik, line 18: better to use an URL as to avoid issues with parsing it

```suggestion
  sampleDatasetLink?: URL
```

discojs/discojs-node/src/data/image_loader.spec.ts
```diff
@@ -84,8 +84,8 @@ describe('image loader', () => {
   })
 
   it('loads multiple samples with labels', async () => {
-    const labels = Range(0, 24).map((label) => (label % 10))
-    const stringLabels = labels.map((label) => label.toString())
+    const labels = Range(0, 24).map(_ => 3) //internally, disco maps string labels to their index in the task LABEL_LIST
```
tharvik:
```suggestion
    const labels = Repeat(3, 24) //internally, disco maps string labels to their index in the task LABEL_LIST
```

cli/src/data.ts
```diff
@@ -23,7 +23,7 @@ async function simplefaceData (task: Task): Promise<data.DataSplit> {
 async function cifar10Data (cifar10: Task): Promise<data.DataSplit> {
   const dir = '../datasets/CIFAR10/'
   const files = (await fs.readdir(dir)).map((file) => path.join(dir, file))
-  const labels = Range(0, 24).map((label) => (label % 10).toString()).toArray()
+  const labels = Range(0, 24).map(_ => 'airplane').toArray() // TODO read labels in csv
```
tharvik: immutablejs has a specific way to have a repeated value, which is quite readable

```suggestion
  const labels = Repeat('airplane', 24).toArray() // TODO read labels in csv
```

server/tests/e2e/federated.spec.ts
```diff
@@ -29,7 +29,7 @@ describe("end-to-end federated", function () {
   async function cifar10user (): Promise<WeightsContainer> {
     const dir = DATASET_DIR + 'CIFAR10/'
     const files = (await fs.readdir(dir)).map((file) => path.join(dir, file))
-    const labels = Range(0, 24).map((label) => (label % 10).toString()).toArray()
+    const labels = Range(0, 24).map(_ => 'cat').toArray() // TODO read labels in csv
```
tharvik:
```suggestion
    const labels = Repeat('cat', 24).toArray() // TODO read labels in csv
```

web-client/src/components/data/dataset_input/DatasetInput.vue
```diff
@@ -150,17 +160,19 @@ import { data } from '@epfml/discojs-core'
...
-  isOnlyPrediction: Boolean
+  isOnlyPrediction: Boolean //TODO: seems to always be false
 }
+const toaster = useToaster()
 
 const props = defineProps<Props>()
 
 const csvRows = ref<{ filename: string, label: string }[]>()
-const isBoxView = ref<boolean>(false)
+const connectImagesByGroup = ref<boolean>(true)
```
tharvik: nitpicking 3000: no need to precise the type when it's derived from the argument

```suggestion
const connectImagesByGroup = ref(true)
```

web-client/src/components/data/dataset_input/DatasetInput.vue
```diff
@@ -184,13 +203,20 @@ const addFiles = (files: FileList, label?: string) => {
...
       }
-
-      csvRows.value.forEach(row => {
-        const imageFile = filesArray.find(file => row.filename === file.name.split('.').slice(0, -1).join('.'))
-        if (imageFile) {
-          props.datasetBuilder.addFiles([imageFile], row.label)
-        }
-      })
+      try {
+        csvRows.value.forEach(row => {
```
tharvik, line 207: better to `map` it and use the resulting array to `datasetBuilder.addFiles`. it's clearer and avoid adding a few files before finding that the CSV is reference an unknown image.

web-client/src/components/data/dataset_input/FileSelection.vue
```diff
@@ -92,7 +92,9 @@
 
         <!-- If no preview of the selected file, display the nbr. of uploaded files -->
         <div class="pt-8 flex flex-col items-center pb-2">
-          <div class="mb-4 flex justify-center items-center text-center md:text-left sm:text-lg text-disco-blue">
+          <div 
+            v-if="selectedFiles?.length"
```
tharvik, line 96: [explicit is better than implicit.](https://peps.python.org/pep-0020/)

```suggestion
            v-if="selectedFiles?.length ?? 0 > 0"
```

web-client/src/components/sidebar/ModelLibrary.vue
```diff
@@ -133,8 +202,25 @@ export default defineComponent({
...
+    ) {
+      return (
+        defaultClass +
+        (state === undefined
+          ? ' '
+          : state
+            ? ' border-slate-900 text-slate-900'
+            : ' text-slate-500')
+      )
+    },
```
tharvik, line 217: consider using a `computed` evaluation (function calls gets called way more often). there is even dedicated infos on [how to do with classes in vuejs](https://vuejs.org/guide/essentials/class-and-style.html) (YMMV I didn't look much into it)
morever, the state is never `undefined` from what I saw so you can simplify it

web-client/src/components/training/Finished.vue
```diff
@@ -88,7 +88,7 @@ async function saveModel () {
     }
   } else {
     toast.error(
-      'The model library is currently turned off. See settings for more information'
+      'The model library is currently turned off. Go to the model library settings to turn it one.'
```
tharvik, line 91:
```suggestion
      'The model library is currently turned off. Go to the model library settings to turn it on.'
```

web-client/src/components/information/Further.vue
```diff
@@ -1,100 +1,73 @@
...
-          </div>
+  <div class="flex justify-center">
+    <IconCard class="justify-self-center w-full lg:w-4/5">
+      <template #title>
+        Further Information
+      </template>
+      <template #content>
+        <div class="mb-4">
+          <b>Goal</b><br>
+          Disco enables collaborative and privacy-preserving training of machine learning models. Disco is an easy-to-use mobile app & web software, running directly in your browser.
```
tharvik, line 11: html needs some of its chars to be encoded :roll_eyes: 

```suggestion
          Disco enables collaborative and privacy-preserving training of machine learning models. Disco is an easy-to-use mobile app &amp; web software, running directly in your browser.
```

web-client/src/components/training/Trainer.vue
```diff
@@ -88,6 +101,8 @@ async function startTraining(distributed: boolean): Promise<void> {
       // missing field is specified between two "quotes"
       const missingFields: String = e.message.split('"')[1].split('"')[0];
       toaster.error(`The input data is missing the field "${missingFields}"`);
+    } else if (e instanceof Error && e.message.includes("provide dataset input files")) {
+      toaster.error("First connect your data at the previous step.")
```
tharvik, line 103: yeah, it's slowly getting out of hand of parsing error messages. consider adding a specific `Error` type for theses cases.

web-client/src/components/training/Trainer.vue
```diff
@@ -61,7 +65,9 @@ const props = defineProps<{
   datasetBuilder: data.DatasetBuilder<File>;
 }>();
 
-const training =
+const displayModelCaching = ref(true)
+
+const trainingLogs =
```
tharvik: hum, it's not really logs (list of events), but a generator (cancellable function outputting data) IMO
that's why I prefered a present participled verb. I'm not pushing for `training` but against `logs`

web-client/src/components/training/Trainer.vue
```diff
@@ -70,7 +76,14 @@ const hasValidationData = computed(
   () => props.task.trainingInformation.validationSplit > 0,
 );
 
+function resetTrainingInformation() {
```
tharvik: it's only called once, consider inlining it (or at least, putting the function inside the body of `startTraining`)

web-client/src/components/training/TrainingInformation.vue
```diff
@@ -37,6 +37,61 @@
...
+        <!-- Card header -->
+        <template #title>
+          Training Loss of the Model
+        </template>
+        <template #content>
+          <span class="text-2xl font-medium text-slate-500">
+            {{ (latestEpoch?.training.loss ?? 0).toFixed(2) }}
+          </span>
+          <span class="text-sm font-medium text-slate-500">
+            ' training loss'
```
tharvik: I'm not sure, is it wanted?

```suggestion
            'training loss'
```

web-client/src/components/training/TrainingInformation.vue
```diff
@@ -180,6 +226,44 @@ const accuracySeries = computed(() => props.logs
...
+    training: training.concat([cur.training]),
+    validation: validation.concat([cur.validation]),
+  }}, {
+    training: [] as number[],
+    validation: [] as number[],
+  })
+)
+
+const yAxisMax = computed(() => {
+  let maxVal = Math.max(...[...lossSeries.value.training, ...lossSeries.value.validation])
```
tharvik:
```suggestion
  let maxVal = Math.max(...lossSeries.value.training, ...lossSeries.value.validation)
```

web-client/src/components/training/TrainingInformation.vue
```diff
@@ -180,6 +226,44 @@ const accuracySeries = computed(() => props.logs
...
+        max: () => yAxisMax.value,
+        min: 0,
+        labels: {
+          show: true,
+          formatter: function (value: number) {
+            return value.toFixed(2);
+          }
+        }
+      }
+    }
```
tharvik: no need for double spread

```suggestion
    yaxis: {
      max: () => yAxisMax.value,
      min: 0,
      labels: {
        show: true,
        formatter: function (value: number) {
          return value.toFixed(2);
        }
      }
    }
```

tharvik (approved):
wow, a fresher and clearer interface, great work! :sparkles: 

as usual, nothing blocking, only a few comments

* good idea to remove i18n, it's was barely used anyway
* and rm useles tests on Landing is a nice too
* for the geotags & skin_mnist task, nothing to salvage?

discojs/discojs-core/src/default_tasks/cifar10.ts
JulienVig: Would it work when the webapp is deployed? I think datasets/populate can only be used in local isn't it?

JulienVig:
> for the geotags & skin_mnist task, nothing to salvage?

There is a map modal implemented in Tester.vue for the geotag task which can be interesting but it relies on https://disco-polygon.web.app/ which doesn't exist anymore.  I couldn't find what was the training data and with the base map missing I'm going to remove the task. It would be interesting to keep it in mind and go back to it if we ever want to implement another location based task
The skin mnist is very similar to LUS Covid but multi-class classification. I don't think it's worth uploading a new sample dataset and some work is required to create a correct CSV file to use the kaggle dataset in Disco

JulienVig:
@tharvik It looks like the npm deployment broke the build-server-docker action?

discojs/discojs-core/src/default_tasks/cifar10.ts
tharvik: correct, do keep it in cifar10's `Task` and *add* it to `datasets/populate` so that we ensure that we also train our tests with the sample dataset.

JulienVig: Ah so some cifar10 data is already included in the `example_training_data` (which requires manipulation to used in the webapp...) but I wanted to provide a public facing link too.

tharvik:
> @tharvik It looks like the npm deployment broke the build-server-docker action?

looks like upstream released a broken nodejs version, followup [there](https://github.com/nodejs/docker-node/issues/2081). openned a PR fixing that #672

PR #675, opened 2024-05-15
Integrate the wikitext task in the webapp

### LINKED ISSUE #669: Can't load tokenizer from the web-client
Uploading the training file `wiki.train.tokens` in the web-client and hitting Train alone triggers the error:
```
SyntaxError: JSON.parse: unexpected character at line 1 column 1 of the JSON data
    getModelJSON hub.js:584
    loadTokenizer tokenizers.js:62
    from_pretrained tokenizers.js:4406
    getTaskTokenizer tokenizer.js:18
    apply text_preprocessing.js:65
    async*get preprocessing/preprocessingChain</< data.js:62
    get preprocessing/preprocessingChain</< data.js:62
    get preprocessing/< data.js:63
    next lazy_iterator.ts:803
    serialNext lazy_iterator.ts:650
    lastRead lazy_iterator.ts:643
    promise callback*next lazy_iterator.ts:643
    next lazy_iterator.ts:711
    fitDataset model.js:56
    train index.js:31
    fitModel trainer.js:36
```
It seems the web-client doesn't manage to load the Transformers.js tokenizer.

### LINKED ISSUE #673: WebGL prevents training wikitext on Firefox
Training wikitext on Firefox raises the following error when first calling `model.fitDataset`:
```
Error: Requested texture size [10145x10146] greater than WebGL maximum on this browser / GPU [8192x8192].
```
According to https://github.com/tensorflow/tfjs/issues/689 this is due to a "fundamental limitation of Firefox". The issue doesn't happen on Chrome.

Potential solution: display a message recommending switching web browsers when a user tries to train wikitext in Firefox

### LINKED ISSUE #674: Saving and loading GPT-tfjs models with IndexedDB fails with error
tfjs fails to load a GPT-tfjs model previously saved to the browser's IndexedDB memory:
```
TypeError: Cannot read properties of undefined (reading 'push')
    at new CausalSelfAttention (layers.js:66:22)
    at fromConfig (@tensorflow_tfjs.js?v=27f417ed:12618:12)
    at deserializeKerasObject (@tensorflow_tfjs.js?v=27f417ed:19329:25)
    at deserialize (@tensorflow_tfjs.js?v=27f417ed:22696:10)
    at processLayer (@tensorflow_tfjs.js?v=27f417ed:24102:21)
    at fromConfig (@tensorflow_tfjs.js?v=27f417ed:24116:7)
    at deserializeKerasObject (@tensorflow_tfjs.js?v=27f417ed:19329:25)
    at deserialize (@tensorflow_tfjs.js?v=27f417ed:22696:10)
    at processLayer (@tensorflow_tfjs.js?v=27f417ed:24102:21)
    at fromConfig (@tensorflow_tfjs.js?v=27f417ed:24116:7)
```

The code triggering the error is the updateWorkingModel method of discojs-web memory:
```
await model.extract().save(indexedDBURL, { includeOptimizer: false })
const loadedModel = await tf.loadLayersModel(indexedDBURL); // raises an error
```
which is called at the end of each training epoch.

Ideally, we would also like to save the optimizer but saving fails when including it. (which is not surprising given that the optimizer is re-implemented manually)

Overall, the GPT-tfjs models have not be integrated to the browser memory framework, which currently assumes that all models are TFJS models.

JulienVig:
- [x] Fixes #669. 
Fixed by preventing local caching

- [x] Fixes #673.  
Added an error message suggesting changing browser.
Screenshot 2024-05-27 at 11 38 19

- [x] Fixes #674 
Extended memory caching and saving for gpt-tfjs

Screenshot 2024-06-11 at 17 23 36

- [x] Rework the validator's test and inference method to enable users to stop and also fix tfjs memory leaks
Screenshot 2024-06-13 at 16 35 53


discojs-web/src/memory/memory.ts
```diff
@@ -10,62 +10,72 @@ import { Map } from 'immutable'
...
 export class IndexedDB extends Memory {
-  override pathFor (source: ModelSource): Path {
+  override getModelMemoryPath (source: ModelSource): Path {
     if (typeof source === 'string') {
       return source
     }
-
-    if (source.type === undefined || source.taskID === undefined || source.name === undefined) {
+    if (source.type === undefined || source.taskID === undefined ||
+      source.name === undefined || source.tensorBackend === undefined) {
```
tharvik: `ModelSource` has only `type` & `version` as potentially undef, no need to check for everyone

```suggestion
    if (source.type === undefined) {
```

discojs-web/src/memory/memory.ts
```diff
@@ -10,62 +10,72 @@ import { Map } from 'immutable'
...
-    const [stringType, taskID, fullName] = source.split('/').splice(2)
+    const [stringType, tensorBackend, taskID, fullName] = source.split('/').splice(2)
 
-    const type = stringType === 'working' ? ModelType.WORKING : ModelType.SAVED
+    const type = stringType === 'working' ? StoredModelType.WORKING : StoredModelType.SAVED
 
     const [name, versionSuffix] = fullName.split('@')
     const version = versionSuffix === undefined ? 0 : Number(versionSuffix)
-    return { type, taskID, name, version }
+    if (tensorBackend != 'tfjs' && tensorBackend != 'gpt') {
```
tharvik: js is quite bad at comparing stuff, it loves to convert things if types differ. that's why `===` & `!==` where introduce, to not do that, as it creates unexcepted bugs in some cases. for example `{ toString: () => 'asd' } == 'asd'` is true which might come at a surprise.
anyway, I think that explicit is way more readable than implicit, or if the latter is used, a comment is welcome.

```suggestion
    if (tensorBackend !== 'tfjs' && tensorBackend !== 'gpt') {
```

discojs-web/src/memory/memory.ts
```diff
@@ -10,62 +10,72 @@ import { Map } from 'immutable'
...
+    const layersModel = await tf.loadLayersModel(this.getModelMemoryPath(source))
+    
+    const tensorBackend = this.getModelInfo(source).tensorBackend
+    if (tensorBackend == 'tfjs') {
+      return new models.TFJS(layersModel)
+    } else if (tensorBackend == 'gpt') {
+      return new models.GPT(undefined, layersModel)
+    } else {
+      throw Error("tensor backend unrecognized")
+    }
```
tharvik: there is support for [exaustiveness checking](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#exhaustiveness-checking), which allow TS to know that every possible value was checked. a neat feature when changing the type of `tensorBackend` (as it will make compilation fail instead of throwing at runtime), and reads more nicely IMO

```suggestion
    switch (tensorBackend) {
      case 'tfjs':
        return new models.TFJS(layersModel)
      case 'gpt':
        return new models.GPT(undefined, layersModel)
    }
```

discojs/src/memory/model_type.ts
```diff
@@ -3,7 +3,7 @@
  * being trained ("working model") or a regular model saved in memory ("saved model").
  * There can only be a single working model for a given task.
  */
-export enum ModelType {
+export enum StoredModelType {
```
tharvik: while at it, we can simply have `export type StoredModelType = 'saved' | 'working'` instead of an enum (or even inlining it in `memory/base.ts` as it's in the only place where it's explicitly needed)

discojs/src/models/gpt/index.ts
```diff
@@ -116,16 +117,22 @@ export class GPT extends Model {
       config: this.config
     }
   }
+  extract (): tf.LayersModel {
+    return this.model
+  }
 
-  [Symbol.dispose](): void{
-    console.log("Disposing model")
+  dispose() {
```
tharvik: consider not having a public function for disposing, only the `[Symbol.dispose]()`. it pushes users to have a clear way to get rid of value, not two

discojs/src/models/tokenizer.ts
```diff
@@ -16,6 +16,7 @@ export async function getTaskTokenizer(task: Task): Promise<PreTrainedTokenizer>
   let tokenizer = task.trainingInformation.tokenizer
   if (tokenizer === undefined) throw Error('No tokenizer specified in the task training information')
   if (typeof tokenizer == 'string') {
+    env.allowLocalModels = false
```
tharvik, line 23: hum, so that allow for loading of file-backed tokenizer, right? I don't see it used and might break in browsers (there is no filesystem, [kinda](https://developer.mozilla.org/en-US/docs/Web/API/File_System_API))

discojs/src/task/training_information.ts
```diff
@@ -57,6 +57,8 @@ export interface TrainingInformation {
   // maxSequenceLength: the maximum length of a input string used as input to a GPT model. It is used during preprocessing to
   // truncate strings to a maximum length. The default value is tokenizer.model_max_length
   maxSequenceLength?: number
+  // Tensor framework used by the model
+  tensorBackend: 'tfjs' | 'gpt'
```
tharvik, line 61: arf, I don't like adding stuff to `Task` and friends, but after looking more into it, it seems required until `Memory` is fixed to be able to store any type of model. too bad.

webapp/src/store/memory.ts
```diff
@@ -33,11 +33,19 @@ export const useMemoryStore = defineStore('memory', () => {
     const models = await tf.io.listModels()
     for (const path in models) {
       // eslint-disable-next-line no-unused-vars
-      const [location, _, directory, task, fullName] = path.split('/')
+      const [location, _, directory, tensorBackend, task, fullName] = path.split('/')
       if (location !== 'indexeddb:') {
         continue
       }
-      const [name, version] = fullName.split('@')
+      console.log(path, fullName)
```
tharvik: debug?

webapp/src/store/memory.ts
```diff
@@ -33,11 +33,19 @@ export const useMemoryStore = defineStore('memory', () => {
...
+      let name
+      let version
+      if (fullName == undefined) {
+        name = 'akjsd'
+        version = 0
```
tharvik: that looks wrong, isn't `fullName` always defined?

webapp/src/store/memory.ts
```diff
@@ -58,10 +66,14 @@ export const useMemoryStore = defineStore('memory', () => {
         metadata.weightDataBytes,
       ].reduce((acc: number, v) => acc + (v === undefined ? 0 : v), 0)
 
+      if (tensorBackend != 'tfjs' && tensorBackend != 'gpt') {
```
tharvik:
```suggestion
      if (tensorBackend !== 'tfjs' && tensorBackend !== 'gpt') {
```

tharvik (approved):
wouhou, gpt in browser, well done!

testing the model is a bit weird, as there is not specific handler for that (got 5000% accuracy, such impressive model). but that's alright for now IMO, we can add it in a later iteration

discojs/src/models/tokenizer.ts
JulienVig: It's actually needed in the browser, it disallows trying to load from the file system and enforces fetching on the huggingface remote hub.

webapp/src/store/memory.ts
JulienVig: Ah good catch that's a remnant from transitioning some old models in memory to the new mechanism

discojs/src/task/training_information.ts
JulienVig: Yes sadly I did this reluctantly...

discojs/src/models/tokenizer.ts
tharvik: ho, right, that's a default true for `allowLocalModels`, nvm then

discojs-web/src/memory/memory.ts
JulienVig: Good to know thanks!

discojs-web/src/memory/memory.ts
JulienVig: Ah yes indeed

discojs-web/src/memory/memory.ts
JulienVig: I'll actually make `ModelInfo.type` required, I don't see any uses of it being left undefined right now

discojs/src/models/tokenizer.ts
JulienVig: I'll add some doc actually it's quite an obscure line of code though very important

discojs/src/models/gpt/index.ts
JulienVig: I find having a `dispose` method expected as it follows [tfjs API](https://github.com/tensorflow/tfjs/blob/fcad805d35a16efc4b9e42c9ef5950cf6c0693f2/tfjs-layers/src/engine/training.ts#L1916) but I guess we're trying to become backend agnostic... I'll remove it

JulienVig:
Did you really get 5000% accuracy? If that's the case this is alarming, I don't see how it could happen

discojs/src/memory/model_type.ts
JulienVig: After turning `StoredModelType` into a `type` I had to replace every call to `StoredModelType.{WORKING|SAVED}` by `'working'` and `'saved'`, was that the right way to do it?

tharvik: yes, that's the typescript way™©®

tharvik:
> Did you really get 5000% accuracy? If that's the case this is alarming, I don't see how it could happen

yep, it's fluctuating around 5000%±1000%, and use a huge amount of memory (before crashing my tab). the reproductability that I've is training wikitext once with "wiki.train.tokens" then testing it with "wiki.test.tokens". I'm using Firefox fwiw. before crashing it shows the next line in console:
`High memory usage in GPU: 1199.58 MB, most likely due to a memory leak`
and indeed, my system memory gets filled up quite rapidly (which triggers OOM, which kills the tab)

JulienVig:
Indeed I'll try to fix that before merging the PR, testing the model takes up more than 50GB

JulienVig:
@tharvik all fixed, I reworked the validator:
* removed the graph_informant (it was only used to query the validator accuracy)
* fixed the memory leak (tfjs sucks)
* fixed the logic mistake that yielded 5000 of accuracy
* renamed assess and predict into test and inference respectively
* made these methods generators to let users stop them. I had to turn the mapAsync into an iterator while loop to allow stopping, do you know if we could combine the functional way and generators somehow?




Until an LLM UI is implemented running inference is useless as it doesn't display anything

tharvik:
> @tharvik all fixed, I reworked the validator:

superbe, it tests nicely now :partying_face: (tfjs is shitty indeed)
and good idea to update the validator to be a real generator, that's more in line w/ training and more usuable.

> * made these methods generators to let users stop them. I had to turn the mapAsync into an iterator while loop to allow stopping, do you know if we could combine the functional way and generators somehow?

hum, no, you sadly have to have a while loop as `tf.data.Dataset` is not using generators. however, as datasets are lazy, you can have a dataset shaped with `mapAsync` and have a small while loop taking care of transforming it in a generator (maybe a bit more readable).
generators a quite powerful, but if libraries aren't exposing theses, it doesn't know how to stop processing a function. it can only `return` between lines, and in this case, there is a huge `await dataset.mapAsync.[...].toArray` so it can only stop before or after that.

> Until an LLM UI is implemented running inference is useless as it doesn't display anything

yep, I'm trying to draft up something basic in my PR, let's see what comes out of it :)

PR #676, opened 2024-05-24
discojs/aggregator: rework tests

tharvik:
to check that the secure aggregator is working, it's better to have some tests for the aggregator

* rework tests of the aggregators
* readd the aggregator to the federated server
* bump some deps

not too much code change, the implementations are pretty much the same.

### DIFF
--- a/discojs/src/aggregator.spec.ts
+++ b/discojs/src/aggregator.spec.ts
@@ -1,0 +1,132 @@
+import { expect } from "chai";
+import { Map, Range, Set } from "immutable";
+
+import { Model, WeightsContainer } from "./index.js";
+import {
+  Aggregator,
+  MeanAggregator,
+  SecureAggregator,
+} from "./aggregator/index.js";
+import { NodeID } from "./client/types.js";
+
+const AGGREGATORS: Set<[name: string, new () => Aggregator]> = Set.of<
+  new (model?: Model) => Aggregator
+>(MeanAggregator, SecureAggregator).map((Aggregator) => [
+  Aggregator.name,
+  Aggregator,
+]);
+
+AGGREGATORS.forEach(([name, Aggregator]) =>
+  describe(`${name} implements Aggregator contract`, () => {
+    it("starts at round zero", () => {
+      const aggregator = new Aggregator();
+
+      expect(aggregator.round).to.equal(0);
+    });
+
+    it("moves forward with enough contributions", async () => {
+      const aggregator = new Aggregator();
+      aggregator.setNodes(Set.of("client 0", "client 1", "client 2"));
+
+      const results = aggregator.receiveResult();
+
+      for (let i = 0; i < 3; i++)
+        for (let r = 0; r < aggregator.communicationRounds; r++)
+          aggregator.add(`client ${i}`, WeightsContainer.of([i]), 0, r);
+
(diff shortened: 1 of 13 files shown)


discojs/src/aggregator.spec.ts
```diff
@@ -0,0 +1,132 @@
...
+
+AGGREGATORS.forEach(([name, Aggregator]) =>
+  describe(`${name} implements Aggregator contract`, () => {
+    it("starts at round zero", () => {
+      const aggregator = new Aggregator();
+
+      expect(aggregator.round).to.equal(0);
+    });
+
+    it("move forward with enough contribution", async () => {
```
JulienVig:
```suggestion
    it("moves forward with enough contributions", async () => {
```

discojs/src/aggregator/secure.ts
```diff
@@ -13,89 +13,140 @@ import { aggregation } from '../index.js'
...
-    const contribs = this.contributions.get(this.communicationRound)
-    if (contribs === undefined) {
-      return false
-    }
-    return contribs.size === this.nodes.size
+  override isFull(): boolean {
+    return (
+      (this.contributions.get(this.communicationRound)?.size ?? 0) ===
+      this.nodes.size
+    );
```
JulienVig, line 96: What's the expected behavior in case `contributions.get(communicationRound)` is not defined?
The previous implementation returned false while this one can return true if `this.nodes` is empty (not sure if this can happen)

server/src/router/federated/server.ts
```diff
@@ -205,32 +193,12 @@ export class Federated extends Server {
         const { payload, round } = msg
         const weights = serialization.weights.decode(payload)
 
-        console.log("federated: recv payload for round", round)
+	this.createPromiseForWeights(task.id, aggregator, ws)
```
JulienVig, line 196: The server answers with a promise for weights both when receiving SendPayload and RequestServerStatistics messages (but only adds the weights to the aggregator for SendPayload messages), is it expected?

JulienVig (approved):
Looks good thanks for the tests!
No blocking comments, only questions for my own understanding

discojs/src/aggregator/secure.ts
tharvik: there is no clear answer. as you pointed out, the aggregators are not well defined with an empty list of nodes. and I didn't wanted to change the aggregator code as I think I needs a real rewrite which isn't mangle with the rest of code (Model & Task extraction, enforcing correct communication rounds via some well typed state transition function (not sure about it), overall simplification of definition and use, …)

in here, I took what I think is the [least suprising path](https://en.wikipedia.org/wiki/Principle_of_least_astonishment): a buffer is full when it has all the contributions it needs. as a zero-sized buffer needs zero contributions, thus it is always full.

server/src/router/federated/server.ts
tharvik: hum no, it's unexcepted. I took [the commit that removed the aggregator support](https://github.com/epfml/disco/commit/8311f2bb6dd691f4d01b0182313063494e50707c) and partially revert it. so it seems that the double use is from before that.
from another look at it, it seems that `{Request,Receive}ServerStatistics` are unused by any client. removing theses.

PR #678, opened 2024-05-27
Create a Skin Condition classification task

JulienVig:
Add a skin condition multi-class classification task as a predefined task in Disco for the upcoming demo. The SCIN dataset is described [here](https://github.com/google-research-datasets/scin/tree/main?tab=readme-ov-file).

- [x] original model previously used for this task currently only predicts the benign class
- [x] create a sample image dataset to provide to users (original dataset is 3GB and requires creating an account on Kaggle and creating the csv label mapping file) -> Available on the [google cloud bucket](https://storage.googleapis.com/deai-313515.appspot.com/skin_lesion_samples.tar.gz)
- [x] Testing a model fails with error
- [x] fine tune architecture and hp. I tried using a pre-trained mobilenet (a334219720b433fedd58b6f733ff8b6c6a365020, commented code in `skin_mnist.ts`) but encountered #679 and decided to settle on a smaller handmade CNN for now.
For 800 images
Screenshot 2024-05-30 at 12 31 58
- [x] Allow connecting a folder rather than individual files -> moved to #680 
- [x] Add an example of how to create csv file to connect data -> `docs/dataset_csv_creation.ipynb`
- [x] At test time, show label names rather than integer numbers
Before:
Screenshot 2024-05-30 at 11 29 53
Now:
Screenshot 2024-05-30 at 12 33 24


discojs/src/default_tasks/skin_mnist.ts
tharvik, line 1: a bunch of commented code in here, please remove it.

discojs/src/default_tasks/skin_mnist_mobilenet/group1-shard1of2.bin
tharvik, line 1: I don't like binaries, do we really need it?

discojs/src/default_tasks/skin_mnist_mobilenet/model.json
tharvik, line 1: you can have it as a real TS file, and `default export` it, so that I can get typechecked in call sites

discojs/src/validation/validator.ts
```diff
@@ -21,23 +21,22 @@ export class Validator {
...
-        throw new Error(`unable to reduce tensor of shape: ${ys.shape.toString()}`)
+  private async getLabel(ys: tf.Tensor): Promise<Float32Array | Int32Array | Uint8Array> {
+    // Binary classification
+    if (ys.shape[1] == 1) {
+      return await ys.greaterEqual(tf.scalar(0.5)).data()
+    // Multi-class classification
+    } else {
+      return await ys.argMax(-1).data()
     }
+    // Multi-label classification is not supported
```
tharvik, line 32: consider throwing here

tharvik (approved):
well done (and sorry that you had to rush it)! a bit of cleanup but nothing vital

discojs/src/default_tasks/skin_mnist.ts
JulienVig: I think you reviewed an old commit, there shouldn't be any commented code right now

discojs/src/default_tasks/skin_mnist_mobilenet/group1-shard1of2.bin
JulienVig: No I forgot to take this folder out it's unused

discojs/src/validation/validator.ts
JulienVig: Hmm I don't think there is a way to detect if the user is doing multi-label classification with the output logits only. I think it will be quite obvious when creating a new task that there is no setting allowing multiple labels for one datapoint.

JulienVig:
@tharvik do you want to review the latest changes or shall I merge the PR?

tharvik (approved):

PR #682, opened 2024-06-05
*: framework agnostic loaders

tharvik:
on the road of supporting ONNX, we first need to extract TFJS from the code base, making discojs framework agnostic.
based on #735.

* adding a Dataset type, wrapping standard [AsyncGenerator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/AsyncGenerator) with some helpers (map, zip, split, batch, ...)
  * for now, `disco.fit`, `validator.{assess,predict}`, requires more type informations (via `TypedDataset`) as theses are supporting any dataset for any task. it should be enforced at some point via `Task` splitting so that we can drop it.
  * now, any users of disco have to use it, keeping the remaining `Datasplit`, `Data`, `tf.*` internal to disco (and removing theses when preprocessing and task typing are reworked)
* also exposes types: `Image`, `Tabular` & `Text` 
* redo loaders to be small functions rather than cumbersome-to-use classes
* added some `convertors` as a potential new way to do preprocessing
  * flat functions, that encodes type changes instead of wrapping in `tf.TensorContainer` (pretty much equivalent to `any`)
* webapp are now typing its dataset, having a clear separation between Image, Tabular & Text, with `Data` component emitting changes


cli/src/benchmark_gpt.ts
```diff
@@ -72,8 +88,28 @@ async function main(args: Required<CLIArguments>): Promise<void> {
...
+                (tokens) =>
+                  tf.oneHot(
+                    tokens.slice(1),
+                    tokenizer.model.vocab.length + 1,
+                  ) as tf.Tensor2D,
+              )
+              .toArray(),
+          ) as tf.Tensor3D,
+        })),
+    );
```
JulienVig: Is the Task's `preprocessingFunctions` attribute obsolete now? I guess they are still used by the webapp on default tasks
Why reimplement the preprocessing rather than use wikitext's preprocessingFunctions ?
I feel a bit uneasy with this way of preprocessing text dataset because it requires users a deep understand of the implementation of tokenize_and_left_pad to know that they should then slice `xs` and `ys` accordingly.

It also seems that setting `task.trainingInformation.batchSize` and `maxSequenceLength` is useless now

discojs-node/src/loaders/index.ts
```diff
@@ -0,0 +1,6 @@
+export { load as loadCSV } from "./csv.js";
+export {
+  load as loadImage,
+  loadAllInDir as loadImagesInDir,
```
JulienVig, line 4: Why not directly rename `loadAllInDir` to `loadImagesInDir`?

discojs-node/src/loaders.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+
+// Array.fromAsync not yet widely used (2024)
+async function arrayFromAsync<T>(iter: AsyncIterable<T>): Promise<T[]> {
+  const ret: T[] = [];
+  for await (const e of iter) ret.push(e);
+  return ret;
+}
+
+describe("csv parser", () => {
+  it("parses basic file", async () => {
```
JulienVig, line 21: Could you add a test for a csv with missing values? (e.g. titanic has missing values)

discojs-node/src/loaders/image.ts
```diff
@@ -0,0 +1,26 @@
+import sharp from "sharp";
+import * as path from "node:path";
+import * as fs from "node:fs/promises";
+
+import { Dataset, Image } from "@epfml/discojs";
+
+export async function load(path: string): Promise<Image> {
+  const { data, info } = await sharp(path).removeAlpha().raw().toBuffer({
+    resolveWithObject: true,
+  });
```
JulienVig: I used to have an issue with the default number of channel being 1 for B&W images, is it always 3 with sharp now?

cli/src/data.ts
```diff
@@ -1,76 +1,53 @@
...
+      return ["image", await loadSimpleFaceData()];
+    case "titanic":
+      return ["tabular", loadCSV("../datasets/titanic_train.csv")];
+    case "cifar10":
+      return [
+        "image",
+        (await loadImagesInDir(path.join("..", "datasets", "CIFAR10"))).zip(
+          Repeat("cat"),
+        ),
+      ];
```
JulienVig, line 48: There seems to be something wrong with training CIFAR10 with the CLI now.
Previously, running `npm -w cli start -- -u 2 -e 3 -t cifar10` would show an initial loss around 2 (which is slightly higher than expected for a 10-class classification: -log(1/n_classes) = 1) but now it is at 20 which is way too high. I tried fiddling with things but I didn't find the source of the issue, maybe the image loader?

discojs-web/src/loaders.spec.ts
```diff
@@ -0,0 +1,22 @@
...
+
+import { loadCSV } from "./loaders/index.js";
+
+async function arrayFromAsync<T>(iter: AsyncIterable<T>): Promise<T[]> {
+  const ret: T[] = [];
+  for await (const e of iter) ret.push(e);
+  return ret;
+}
+
+describe("csv parser", () => {
```
JulienVig, line 11: Is it hard to add tests for images and text? I know there wasn't any test at all before

discojs/src/convertors.ts
```diff
@@ -0,0 +1,69 @@
...
+function isArrayOfNumber(raw: unknown): raw is number[] {
+  return Array.isArray(raw) && raw.every((e) => typeof e === "number");
+}
+
+/** Tokenize and truncates input strings
+ *
+ * @param length number of tokens
+ * @returns encoded string in an array of token, size of max_length
+ **/
+export function tokenize_and_left_pad(
```
JulienVig: Is there a plan to remove the code duplication between this function and the text preprocessing in the future?

discojs/src/convertors.ts
JulienVig, line 1: What's the idea of this file? Is it a repository of preprocessing functions?

discojs/src/dataset/types.ts
```diff
@@ -0,0 +1,12 @@
+import { Dataset } from "./dataset.js";
+
+// TODO use full type to check shape of array
```
JulienVig: What do you mean by "use full type"?

discojs/src/models/model.ts
```diff
@@ -19,6 +18,7 @@ export interface EpochLogs {
 // TODO still bound to tfjs
 export type Prediction = tf.Tensor;
 export type Sample = tf.Tensor;
+type Dataset = tf.data.Dataset<tf.TensorContainer>
```
JulienVig: I find the name overlap between the discojs Dataset and tf.data.Dataset confusing, what about renaming this type tfDataset or directly using tf.data.Dataset rather than creating a new type?
Or is it a temporary solution?

discojs/src/task/training_information.ts
```diff
@@ -138,7 +138,7 @@ export function isTrainingInformation (raw: unknown): raw is TrainingInformation
     if (!(Array.isArray(inputColumns) && inputColumns.every((e) => typeof e === 'string'))) {
       return false
     }
-    if (!(Array.isArray(outputColumns) && outputColumns.every((e) => typeof e === 'string'))) {
+    if (!(Array.isArray(outputColumn) && outputColumn.every((e) => typeof e === 'string'))) {
```
JulienVig: Doesn't `outputColumn` have to be a string now? Can it be an array?

webapp/src/components/data/TabularDatasetInput.vue
JulienVig, line 1: There is now ImageDatasetInput and TabularDatasetInput, what about TextDatasetInput?

docs/examples/wikitext.ts
JulienVig, line 1: Running `npm run language_model` in the `examples` folder now hangs forever (or a long time) between the training start and the first epoch. There's currently no issue on the develop branch.

webapp/src/components/containers/ImageCard.vue
```diff
@@ -1,43 +1,52 @@
...
       v-if="showButton"
       class="opacity-0 hover:opacity-100 duration-500 absolute inset-0 z-10 flex justify-center items-end text-white font-semibold p-5"
     >
-      <CustomButton
-        class="text-xs"
-      >
-        click to show on maps
-      </CustomButton>
+      <CustomButton class="text-xs"> click to show on maps </CustomButton>
     </div>
```
JulienVig: I think the showButton prop, the CustomButton and the draw function can be removed, they were probably used for the Polyfill task

webapp/src/components/data/ImageDatasetInput.vue
```diff
@@ -0,0 +1,254 @@
...
+      return [[file, label] as const];
+    })
+    .reduce(
+      (acc, [file, label]) =>
+        acc.set(label, acc.get(label, Set<File>()).add(file)),
+      Map(),
+    );
+
+  if (!fullyFilled.value)
+    throw new Error("given folder is missing some files referenced in the CSV");
```
JulienVig: This error message is confusing because fullyFilled can be false if not all labels are represented in the dataset, even though all files referenced in the CSV are connected. This happened when I tried connecting datasets/cifar10-labels.csv

JulienVig (changes requested):
Huge amount of work, well done!! Things feel much cleaner now!

I think there are a few issues with the webapp right now:
* Clicking next on a task without connecting data doesn't display anthing anymore at the Training step (while it used to display the training board)
Screenshot 2024-06-06 at 13 54 23
* Connecting data, clicking next to the Training board now displays the training things correctly but if I go back I can't clear the connected files anymore.
* Connecting the LUS COVID dataset and training fails with error `Error: Based on the provided shape, [224,224,3], the tensor should have 150528 values but has 200704`, maybe linked to the new image loading function? A similar errors occurs on simple face: `Error: Based on the provided shape, [200,200,3], the tensor should have 120000 values but has 160000`
* Training on titanic works well, however the webapp fails to display the testing page correctly afterward and the Test button doesn't appear. (Sidenote titanic training currently fails on develop with error: `Error: A Dataset iterator for fitDataset() is expected to generate objects of the form {xs: xVal, ys: yVal}, where the two values may be tf.Tensor, an array of Tensors, or a map of string to Tensor.  The provided Dataset instead generates Tensor` so you seemed to have fixed this error)
Screenshot 2024-06-06 at 14 01 45
* Training on the Skin condition and CIFAR10 tasks fails with error `TypeError: valueScalar is undefined`, you can find the sample skin condition dataset [here](https://storage.googleapis.com/deai-313515.appspot.com/scin_sample.zip).
* There is no TextDatasetInput.vue so there is no way to connect data to the wikitext task in the webapp Screenshot 2024-06-06 at 14 10 57

I've been terribly confused with the naming of `Dataset`, `Data`, `DataSplit` and `tf.data.Dataset`. A list of specific things I don't like about the current abstractions:
* Dataset is not a data structure composed of Data elements
* Data is actually a dataset (and it is `Data` that wraps `tf.data.Dataset` rather than `Dataset`)
* `DataSplit` represents a train-test split of `Data` rather than a dataset
* dataset ends up being an attribute of data like in `data.train.preprocess().batch().dataset`

Can we either add more class documentation to clarify the relationships or rework the oriented-object architecture in a way that makes things self-explanatory? This is a bit beyond the scope of this PR but it feels like the right time to address it? Feel free to ignore otherwise I'm also happy to create an additional PR myself.

tharvik:
> * Clicking next on a task without connecting data doesn't display anthing anymore at the Training step (while it used to display the training board)

as no dataset was entered (it's undef), one can't train anything. I added a small message to tell users to enter some. one more functional way would be to rework the steps to disable the next step until data are entered.

> * Connecting data, clicking next to the Training board now displays the training things correctly but if I go back I can't clear the connected files anymore.

good catch! I indeed forgot to propagate the "clear file" event

> * Connecting the LUS COVID dataset and training fails with error `Error: Based on the provided shape, [224,224,3], the tensor should have 150528 values but has 200704`, maybe linked to the new image loading function? A similar errors occurs on simple face: `Error: Based on the provided shape, [200,200,3], the tensor should have 120000 values but has 160000`

indeed, I mixed RGBA & RGB images, added a real `Image` type and related `remove_alpha` convertor.

> * Training on titanic works well, however the webapp fails to display the testing page correctly afterward and the Test button doesn't appear.

right, thanks for catching that, I missplaced a `toRaw` in there. NB: this is needed as VueJS proxies every value, but it doesn't work with [JS private fields](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_properties) (`#` prefixed values).

> * Training on the Skin condition and CIFAR10 tasks fails with error `TypeError: valueScalar is undefined`, you can find the sample skin condition dataset [here](https://storage.googleapis.com/deai-313515.appspot.com/scin_sample.zip).

ho oupsi, I'm generating an empty dataset when uploading a folder, should have added a throwing TODO, thanks for catching that.

> * There is no TextDatasetInput.vue so there is no way to connect data to the wikitext task in the webapp

correct, there isn't any. we can add one in a next iteration but that out of scope of the PR. from what I remember, one student should have worked on that but it wasn't clear if it happened in the end.

> I've been terribly confused with the naming of `Dataset`, `Data`, `DataSplit` and `tf.data.Dataset`. A list of specific things I don't like about the current abstractions:

I totally agree with you, it's not understandable. that's also why this PR is introducing a new one that should replace all of theses in its next iterations (when I reworked the convertors).

> * Dataset is not a data structure composed of Data elements

that would be the central (and only) way to have a serie of data (image/tabular/text/...) with transformations applied on it.

> * Data is actually a dataset (and it is `Data` that wraps `tf.data.Dataset` rather than `Dataset`)

it will be dropped, replaced by `Dataset<Image>`, `Dataset<Tabular>` & `Dataset<Text>` and some convertors applied on it. 

> * `DataSplit` represents a train-test split of `Data` rather than a dataset

it will be replaced by the return type of `Dataset<T>.split()` ie `[Dataset<T>, Dataset<T>]`.

> * dataset ends up being an attribute of data like in `data.train.preprocess().batch().dataset`

it will be real values, not fields, by shaping `Dataset<T>` via convertors functions. for example, `Dataset<Tabular> -> extract_columns -> convert_to_numbers -> Dataset<[features: number[], label: number]>`

> Can we either add more class documentation to clarify the relationships or rework the oriented-object architecture in a way that makes things self-explanatory? This is a bit beyond the scope of this PR but it feels like the right time to address it? Feel free to ignore otherwise I'm also happy to create an additional PR myself.

as I'm hoping to remove most of theses soon, it seems a bit premature to update the doc now. I'll update the doc accordingly in the next one, but feel free to setup a PR if you feel that's it's taking too long.

JulienVig:
Sounds good! Is there a quick fix to support text dataset? Merging this PR means that wikitext will not be available on discolab.ai until a new PR implements text datasets.

cli/src/benchmark_gpt.ts
tharvik: the issue with `preprocessingFunctions` is that there aren't typesafe at all. there is no way that I can think of in TS to transform values by applying an array of function that doesn't use a common type (in this case, `tf.TensorContainer` which is akin to `any`, that gets cast to the correct shape in every `apply`).
what I've in mind is that `Disco` can take "user-friendly Dataset" (`Dataset<Image/Tabular/Text>`) and takes care of shaping theses into what `models.*` can handle (tokenized inputs, resized images, ...). as such, users that direclty want to use models should apply theses functions themself.
but yeah, as this is a ongoing update, this is quite ugly to use models direclty for now. in a next iteration, models will directly take our `Dataset` and takes care of mapping to its own tensors (which would be framework specific).
does that makes sense and/or do you think that's a good idea?

discojs-node/src/loaders/index.ts
tharvik: I'm overly scope sensitive: names inside a file called `image.ts` doesn't need to contain `Image` as it is already in the filename. this has some logic when moving files around, as it avoid changing the file itself, only its parents.

discojs-node/src/loaders/image.ts
tharvik: I rewrote part of it, now `Image` is a real class (for checking the size of the array with width, height and depth) and type-encode most arguments. I'm failling with 1 & 2 channels image now, is it common to have B&W images?

cli/src/data.ts
tharvik: ha ho, oupsi, it seems that I inverted the generator oneHot tensor, returning an array full of ones and a single zero for the label, rather than the other way around. hard to find, thanks for the check!

discojs-web/src/loaders.spec.ts
tharvik: yeah, for images, it's hard: the testing environnement (jsdom) isn't a full browser and doesn't have support for either OffScreenCanvas nor `URL.createObjectURL`, which is used by the current implementation. I don't see another (more testable and not adding deps) to do so sadly.

for text, there isn't even a loader for now, but that would be easy IMO. I didn't wanted to write one before it was needed as to have a better understanding of the needed interface.

discojs/src/convertors.ts
tharvik: yeah, for sure, removing `{Text,Image,Tabular}Processing` and simply use the convertors

discojs/src/convertors.ts
tharvik: pretty much yeah, it's more than preprocessing as some will be used for postprocessing (such as untokenization, `number -> label` translation, ...)
it has the advantage of not being weird classes but simple function, encoding real type transformation, and fittin pretty nicely into `Dataset.map` (even more nicely if curring was supported in JS but yeah, it's not Haskell)
I'll probably change it some more, probably splitting it into various files, but the interface should be quite similar

that's a proposition of replacement anyway, if you think of another nicer way to do it, do tell me :)

discojs/src/dataset/types.ts
tharvik: using a class in this case, to actively check that the array size is consistent with width, height and depth. it is now implemented in the latest commits.

discojs/src/models/model.ts
tharvik: yeah, inline makes more sense, pushed a commit. I've been working on this PR for some time now, there is a bunch of leftovers from various temporary ideas

discojs/src/task/training_information.ts
tharvik: ho right, that doesn't make sense, good catch!

cli/src/benchmark_gpt.ts
JulienVig: Yes this makes sense! I was mostly wondering about the code duplciation, I agree that the current preprocessing should be reworked and how you describe it sounds good. I would try to follow pytorch and tensorflow's api as much as we can

discojs-node/src/loaders/image.ts
JulienVig: Medical images are frequently black and white yes

JulienVig:
> as no dataset was entered (it's undef), one can't train anything. I added a small message to tell users to enter some. one more functional way would be to rework the steps to disable the next step until data are entered.

I quite liked simply displaying an error message when training without data ("input a dataset first"). That allowed new users to walk around and see things when they don't want to train an actual model. What do you think of it?

docs/examples/wikitext.ts
tharvik: yeah, that's bothering. it is working but very slowly as it is not using tfjs's node accelerator. adding it back in the example.

webapp/src/components/containers/ImageCard.vue
tharvik: indeed, `showButton` is never used, removing that.
I'm keeping the `draw` function as it is used when showing results of evaluation for images

webapp/src/components/data/ImageDatasetInput.vue
tharvik: ha yes, I tried to only `emit` Dataset when theses are ~full, but you're right, one can input partial dataset also. removed the `fullyFilled` logic and inversed the folder parsing thus respecting the error message but doing it right :musical_note:

discojs-node/src/loaders/image.ts
tharvik: alright, added support for depth of one, and related convertor for expanding an image to multicolor, thanks for point that out!

discojs-node/src/loaders.spec.ts
tharvik: what do you mean by missing value? the loader is quite basic, only checking that is has the same number of fields as the headers. for example
```
a,b,c
1,2,3
4,,6
```
will get parsed as
```js
[
  {a: "1", b: "2", c: "3"},
  {a: "4", b: "", c: "6"},
]
```

I think it should stay dumb and that convertors should ensure that theses fields are valid (ie, in a specific range or element of a list, mapped to number, ...).

tharvik:
> Is there a quick fix to support text dataset? Merging this PR means that wikitext will not be available on discolab.ai until a new PR implements text datasets.

hum, I think that's doable, not for the webapp Tester (it's not really supported now anyway) but for the others parts, yes.

> > as no dataset was entered (it's undef), one can't train anything. I added a small message to tell users to enter some. one more functional way would be to rework the steps to disable the next step until data are entered.
> 
> I quite liked simply displaying an error message when training without data ("input a dataset first"). That allowed new users to walk around and see things when they don't want to train an actual model. What do you think of it?

hum, users are weird, but yeah, make sense, it's very demo-y; adding it back.

tharvik:
so, it's alive again! way smaller to review thanks to extracted scraps. description updated.

JulienVig:
There seems to be something wrong with the image data loaders: connecting images (by group for lus_covid and by csv for mnist) results in an error (`Something went wrong on our side`) when clicking train (both alone and collaboratively). Titanic and the llm task work fine.

cli/src/data.ts
```diff
@@ -1,76 +1,51 @@
...
-    case 'cifar10':
-      return await cifar10Data(task)
-    case 'lus_covid':
-      return await lusCovidData(task)
-    case 'YOUR CUSTOM TASK HERE':
-      throw new Error('YOUR CUSTOM FUNCTION HERE')
+    case "simple_face":
+      return ["image", await loadSimpleFaceData()];
+    case "titanic":
+      return ["tabular", loadCSV("../datasets/titanic_train.csv")];
```
JulienVig:
```suggestion
      return ["tabular", loadCSV(path.join("..", "datasets", "titanic_train.csv"))];
```
To make it consistent with line 42

discojs-node/src/loaders.spec.ts
JulienVig: Oops forgot to answer. Yes the b in your example was what I was talking about so all good!

discojs/src/dataset/dataset.ts
```diff
@@ -1,6 +1,177 @@
...
- */
-export type Dataset = tf.data.Dataset<tf.TensorContainer>
+type DatasetLike<T> =
+  | AsyncIterable<T>
+  | Iterable<T>
+  // generators
+  | (() => AsyncIterator<T, void>)
+  | (() => Iterator<T, void>);
+
+/** Immutable serie of data */
```
JulienVig:
```suggestion
/** Immutable series of data */
```
Series always has an s even when there's a single series...

discojs/src/dataset/image.ts
```diff
@@ -0,0 +1,15 @@
+export class Image<
+  D extends 1 | 3 | 4 = 1 | 3 | 4,
```
JulienVig, line 9: Could you add a comment clarifying what the superclass is?

cli/src/benchmark_gpt.ts
```diff
@@ -74,12 +87,29 @@ async function main(args: Required<CLIArguments>): Promise<void> {
...
```
JulienVig: I noticed a memory leak when running `DEBUG=* npm -w cli run benchmark_gpt`. I expect that the preprocessing needs some tf.tidy and tf.dispose

JulienVig: Similarly, printing the peakMemory from the EpochLogs shows an increasing memory when using the webapp

webapp/src/components/training/Training.vue
```diff
@@ -13,106 +10,121 @@
...
+      :dataset="dataset"
     />
 
-    <Finished
-      v-show="trainingStore.step === 4"
-      :task="task"
-    />
+    <Finished v-show="trainingStore.step === 4" :task="task" />
+
+    <Finished v-show="trainingStore.step === 4" :task="task" />
```
JulienVig:
```suggestion
```
Duplicate line

webapp/src/components/testing/Testing.vue
```diff
@@ -165,152 +185,186 @@
...
+  try {
+    for await (const [columns, i] of dataset
+      .map((row) => Set(Object.keys(row)))
+      .zip(Range(1)))
+      if (!columns.isSuperset(wantedColumns))
+        throw new Error(
+          `row ${i} is missing columns ${wantedColumns.subtract(columns).join(", ")}`,
+        );
+  } catch (e) {
+    let msg = "Error when loading CSV";
```
JulienVig: When connecting the same titanic csv I used during training at the testing step, I get the error: `Error when loading CSV: can't access private field or method: object is not the right class`

JulienVig: Same error if I connect a misformatted csv

webapp/src/components/testing/Testing.vue
```diff
@@ -144,19 +165,18 @@
       <div v-if="stepRef === 1">
```
JulienVig:
```suggestion
      <div 
        v-if="stepRef === 1"
        class="flex flex-col space-y-4 md:space-y-8"
      >
```
Follow same style as training. Right now there is no gap between the divs:
Screenshot 2024-08-13 at 14 06 34

webapp/src/components/dataset_input/LabeledImageDatasetInput/__tests__/ByCSV.spec.ts
JulienVig, line 1: Why are we removing the tests?

discojs/src/convertors.ts
JulienVig: Edit: this is a reply to a comment in the previous review that has been detached from its thread

I like the idea of having simple stateless functions for preprocessing. I would prefer if it was more organized, maybe by dataset type but that can be handled later on as we add more functions. 
I think there may be some overlap between this file and `data/helpers.ts`, could you add some doc about the scope of each file to clarify where functions should go? 
On that matter maybe `convertors.ts` is a confusing name as not all functions there convert things (and some functions in helpers.ts also convert things). What about something like `processing.ts` (for both pre- and post-processing)?

discojs/src/convertors.ts
```diff
@@ -0,0 +1,126 @@
...
+function isArrayOfNumber(raw: unknown): raw is number[] {
+  return Array.isArray(raw) && raw.every((e) => typeof e === "number");
+}
+
+/** Tokenize and truncates input strings
+ *
+ * @param length number of tokens
+ * @returns encoded string in an array of token, size of max_length
+ **/
+export function tokenizeAndLeftPad(
```
JulienVig: Right now this function is only used by the cli and not by the trainingInformation preprocessing, is this intened?

tharvik:
> There seems to be something wrong with the image data loaders

right, I missplaced a `toRaw` call. should be fixed. and added a test for lus_covid's training to catch this earlier.

docs/examples/training.ts
```diff
@@ -66,35 +68,15 @@ async function main (): Promise<void> {
...
-  const titanicTask = defaultTasks.titanic.getTask()
-  return await new NodeTabularLoader(task, ',').loadAll(files, {
-    features: titanicTask.trainingInformation.inputColumns,
-    labels: titanicTask.trainingInformation.outputColumns,
-    shuffle: false
-  })
+  return adults.chain(childs);
 }
 
 // You can run this example with "npm start" from this folder
```
JulienVig:
```suggestion
// You can run this example with "npm run train" from this folder
```

docs/examples/custom_task.ts
```diff
@@ -1,4 +1,4 @@
-import tf from '@tensorflow/tfjs'
+import tf from '@tensorflow/tfjs-node'
```
JulienVig, line 1: This is probably due to something else than this import but running `npm run custom_task` hangs forever (or at least a long time)

webapp/src/components/dataset_input/LabeledImageDatasetInput/__tests__/ByCSV.spec.ts
JulienVig: Ah I guess it's been moved to e2e/datasetInput.cy.ts

JulienVig (changes requested):
Impressive work! You've virtually rewritten most of the codebase haha

The evaluation seems to have an issue with data loading:
* Go to the Evaluation tab, test a GPT language model, connect a text file, click next: Toaster error "Something went wrong" and nothing displayed. Now If I click Previous the text file I previously connected has been cleared
* Same error with lus_covid (test a model, connect images, next -> nothing displayed).

discojs/src/dataset/dataset.ts
tharvik: human languages will never stop astonishing me :eye_speech_bubble:

webapp/src/components/testing/Testing.vue
tharvik: and same fix as before, was a bit fast on this one :sweat_smile:

webapp/src/components/dataset_input/LabeledImageDatasetInput/__tests__/ByCSV.spec.ts
tharvik: indeed; I moved to cypress as vue's test-utils is not moving fast [on v-model issues](https://github.com/vuejs/test-utils/issues/2468).

cli/src/benchmark_gpt.ts
tharvik: yeah, I forgot to clean it nicely, it should be fixed now. (tfjs is hard to memory manage..)

discojs/src/convertors.ts
tharvik: for now yes, I didn't wanted to modify too much the internals of discojs. a rewrite of preprocessing is baking in the next PR.

discojs/src/convertors.ts
tharvik: > I like the idea of having simple stateless functions for preprocessing. I would prefer if it was more organized, maybe by dataset type but that can be handled later on as we add more functions.

yep, it's a draft for now, it'll get nicer in the next PR. thanks for the remarks on small  functions, I think it indeed does help with understanding what is actually going on in preprocessing. it is also a requirement to have type safe preprocessing (as it modify the type of values which is sadly currently hidden via `tf.TensorContainer`)

> I think there may be some overlap between this file and `data/helpers.ts`, could you add some doc about the scope of each file to clarify where functions should go? On that matter maybe `convertors.ts` is a confusing name as not all functions there convert things (and some functions in helpers.ts also convert things).

yeah, the difference is not really clear in code. `data/helpers` is temporary and only for converting `Dataset` to `Data` & `DataSplit`, it'll get removed soon. `convertors` (or rather `processing` now) is a collections of functions to shape `Dataset` into what's models can ingest, on the way of replacing what the soon-to-be-old `data/preprocessing`.
I added some comments to make it clearer.

> What about something like `processing.ts` (for both pre- and post-processing)?

`processing` it is!

docs/examples/custom_task.ts
tharvik: hum, it's already the case in develop, is it not wanted? from a quick look, it seems to be only starting a server and never closing it (node doesn't exit as long a their are some stuff/promises running)

tharvik:
> Impressive work! You've virtually rewritten most of the codebase haha

haha, slowly getting there! when changing something as fundamental as tfjs, there is some good rewrite indeed.

> The evaluation seems to have an issue with data loading:

that should be fixed now, and some cypress tests to ensure that also. we still have a memory leak of one tensor every round. I didn't wanted to sink _too_ much time into it.

JulienVig (changes requested):
Thanks for the improvements!! I found a few things remaining:
* Training on lus_covid and then testing the model displays 0% accuracy and 0 samples visited
* Trying to test/predict a model on titanic data by connecting the titanic_test.csv or titanic_train.csv doesn't display enough and there is a toaster error
* Idem when trying to test a language model
* The Stop training button doesn't work when the training fails (for example if you train collaboratively with mnist with a single user, the user timeouts waiting for other contributions. After the timeout, hitting stop training doesn't work) 
* (At some point I also got the error "multiple dataset entered info=render function thrown undefined" but I can't reproduce it...)

tharvik:
> * Training on lus_covid and then testing the model displays 0% accuracy and 0 samples visited
> * Trying to test/predict a model on titanic data by connecting the titanic_test.csv or titanic_train.csv doesn't display enough and there is a toaster error
> * Idem when trying to test a language model

haa, one of the dark corner of disco. so I finally took hold of it and rework the whole discojs' validation stack
* predict yield the prediction for every line of the dataset
* test yield a boolean if the prediction is correct
* no more duplicated logic between the two, no more materialisation of the dataset's values in the results, no more mutability :sun_with_face:
* specialize Tester into PredictSteps & TestSteps
* simplify validation's store
  * we can even remove it down the road, if we rework the URL to actually encode the model ID in it

> * The Stop training button doesn't work when the training fails (for example if you train collaboratively with mnist with a single user, the user timeouts waiting for other contributions. After the timeout, hitting stop training doesn't work)

arf, too bad, it comes from the `AbortController` only reacting when between await points. and as Disco is never returning when an error occurs (that should be the case but I was already kinda yack shaving), it hangs on the `for await`. throwing allows for propagation of the failure. I'm reintroducing it and we can change it again when Disco behaves a bit more nicely.

> * (At some point I also got the error "multiple dataset entered info=render function thrown undefined" but I can't reproduce it...)

hum, that shouldn't happen. it comes when entering multiple type of dataset in test/predict, which shouldn't be possible. I tried a bit to reproduce it but wasn't able to.

webapp/src/store/validation.ts
```diff
@@ -1,15 +1,9 @@
...
+import { defineStore } from "pinia";
+import { ref } from "vue";
 
-export const useValidationStore = defineStore('validation', () => {
-  const state = ref(false)
-  const step = ref(0)
-  const model = ref<string>()
-  const isOnlyPrediction = ref(false)
+export const useValidationStore = defineStore("validation", () => {
+  const step = ref<0 | 1 | 2>(0);
```
JulienVig, line 5: Very neat!

webapp/src/components/testing/PredictSteps.vue
```diff
@@ -0,0 +1,342 @@
+<template>
+  <!-- Language model prompting is currently unavailable   -->
+  <div
+    v-if="task.trainingInformation.dataType === 'text'"
+    v-show="validationStore.step !== 0"
+  >
+    <div class="flex justify-center items-center mb-4">
+      <span class="shrink-0 py-4 px-4 bg-orange-100 rounded-md">
+        <p class="text-slate-600 text-xs">
```
JulienVig:
```suggestion
      <span class="shrink-0 py-4 px-4 bg-purple-100 rounded-md
        flex flex-row gap-x-4 items-center"
      >
        <InfoIcon custom-class="min-w-6 min-h-6 w-6 h-6 text-slate-600" />
```
To follow the new purple banner style
Make sure to import InfoIcon :)
```
import InfoIcon from "@/assets/svg/InfoIcon.vue";
```

JulienVig (approved):
Awesome!! Better validator and validator store and the test suite is getting more and more significant!

All good for me, could you maybe merge your PR after #710 (or rebase on it) so that @tomasoignons doesn't have to do it?

webapp/src/components/testing/PredictSteps.vue
JulienVig: hmm it seems the code suggestion isn't pointing at the right line... it should point to l.8 not l.9

martinjaggi:
huge thanks, massive change here!

JulienVig:
Yay we've done it! 🎉

martinjaggi:
congrats!

PR #686, opened 2024-06-17
UI polishing

JulienVig:
UI fixes and improvements before the public release

- [x] Make the wikitext task a generic LLM training task
- [x] Reword tasks textual descriptions. Add explanations on how to download and connect sample datasets
  - [x] LLM
  - [x] CIFAR10 -> hiding it because performance are too bad when using just a few images
  - [x] Titanic
  - [x] Simple Face -> hiding it for the PR, couldn't find the original dataset and face recognition isn't the most ethical use case
  - [x] Covid
  - [x] MNIST
  - [x] Skin condition -> removed
- [x] ~~Fix the custom task form~~ wasn't tested thus drifted behind, would need a rewrite
- [x] In the task list, display "federated" and "decentralized" as UI tags rather than in the title
- [x] Limit the number of pre-defined tasks (Remove Skin Condition?)
- [x] Add more user feedback between training started and the first epoch
  - exposing batch processing in #688 

Addresses parts of #628 

Among other things, new home screen text:
Screenshot 2024-06-26 at 12 10 16

New task list UI and textual descriptions:
Screenshot 2024-06-26 at 12 09 08

Improved file selection logic and UI
Screenshot 2024-06-26 at 12 10 01

**Moving the following to a new PR:**
- [ ] Add UI messages that this is a demo and training states and custom tasks are not persisted
- [ ] Add feedback form link
- [ ] Add LLM training ETA
- [ ] Display LLM inference message that currently only the CLI inference is available


JulienVig:
@martinjaggi Regarding the wikitext dataset I'm planning to add a simple progress status like 10/1000 epochs but as we are using a batch size of 1 I expect the number of epochs and ETA to be very very large. Should I leave it as it is or provide a smaller subset of wikitext?

martinjaggi:
i think your current setting took only a few minutes to run in the browser so the training config seems already ok? (does it do an epoch or only some number of steps?). either i would say that, or put the the info n/N steps or n/N epochs progress status as you mentioned.

anyway it's only a UI issue about the info i think.

webapp/src/store/tasks.ts
```diff
@@ -9,11 +9,16 @@ import { useToaster } from '@/composables/toaster'
 import { CONFIG } from '@/config'
 import { useTrainingStore } from './training'
 
+const TASKS_TO_FILTER_OUT = ['simple_face', 'cifar10']
```
JulienVig, line 12: Dirty and ugly fix to hide some tasks in the webapp, is it worth implementing something better? (many tests rely on those tasks that's why I didn't simply remove them)

tharvik: it'll be nice to have theses at all, so we need to update tests at some point. but no rush there, let's keep that for the release.

webapp/src/assets/svg/PlugIcon.vue
```diff
@@ -0,0 +1,25 @@
...
+
+</template>
+<script lang="ts">
+export default {
+  name: 'PlugIcon',
+  props: {
+    customClass: {
+      default: 'w-8 h-8 text-disco-cyan',
+      type: String
+    },
```
tharvik, line 21: I saw this way of having an extra class in most svg. why do we need it? can't it be set by the parent component?

webapp/src/components/data/DataDescription.vue
```diff
@@ -0,0 +1,59 @@
+<script setup lang="ts">
+import type { Task } from '@epfml/discojs'
+
+import DropdownCard from '@/components/containers/DropdownCard.vue'
+
+interface Props {
+  task: Task
+}
+const props = defineProps<Props>()
```
tharvik: consider not depending on `Task` but rather on `DisplayInformation` as too specify which part of this global is needed. also, I link inlining, and no need to bind it to `props`, you can direclty use it (`props.taks.displayinformation.[...] -> info.[...]`)

```suggestion
defineProps<{
  info: DisplayInformation
}>()
```

webapp/src/components/data/dataset_input/FileSelection.vue
```diff
@@ -90,76 +94,161 @@
...
-
   e.dataTransfer.dropEffect = 'copy'
-  emit('input', e.dataTransfer.files)
-  selectedFiles.value = e.dataTransfer.files
+  const files = e.dataTransfer.files
+  addFiles(files)
 }
+
+const addFiles = (files: FileList) => {
+  if (props.expectCsvMapping === true) {
```
tharvik: it's always def and boolean

```suggestion
  if (props.expectCsvMapping) {
```

webapp/src/components/data/dataset_input/FileSelection.vue
```diff
@@ -90,76 +94,161 @@
...
+    if (props.task.trainingInformation.dataType === 'image' && requireLabels.value) {
+      if (props.label !== undefined) {
+        props.datasetBuilder.addFiles(filesArray, props.label)
+      } else {
+        if (props.csvRows === undefined) {
+          throw new Error('adding files but no CSV rows defined')
+        }
+        props.csvRows.forEach(row => {
+          // Match the selected files with the csv file names and label
+          const imageFile = filesArray.find(file => row.filename === file.name.split('.').slice(0, -1).join('.'))
```
tharvik: this is O(n²), which with many images might cause quite a slowdown. consider creating a `Map` from filename to file instead (to have O(n·log(n))). I had the same issue when adding framework agnostic loaders, maybe [that](https://github.com/epfml/disco/pull/682/files#diff-2bde47ca2e621ffb510dd750c86bd0718fc25a6680a81b38e8c4f997af913376R226) can help.

webapp/src/components/data/dataset_input/FileSelection.vue
```diff
@@ -90,76 +94,161 @@
...
+          if (imageFile === undefined) {
+            toaster.error("Some images listed in the CSV file are missing, make sure the CSV filenames don't include file extensions.")
+            throw new Error("Image not found in the CSV file")
+          } else if (imageFile !== undefined) {
+            props.datasetBuilder.addFiles([imageFile], row.label)
+          }
```
tharvik: unecessary if nesting, as the other branch throws. the flatter, the more readable it is as it avoid keeping more context while reading it.
```suggestion
          }
          props.datasetBuilder.addFiles([imageFile], row.label)
```

webapp/src/components/data/dataset_input/FileSelection.vue
```diff
@@ -90,76 +94,161 @@
...
+    }
+    selectedFiles.value = files
+    hideConnectField.value = true
+  } catch {
+    clearFiles()
+  }
+}
+
+const readCsvImageMapping = (files: FileList) => {
+  try {
```
tharvik: as `files[0].text()` is not awaited, it can't throw, so this try-catch will never be triggered. consider moving it to a full async function.

webapp/src/components/pages/TaskList.vue
```diff
@@ -89,6 +99,27 @@ const sortedTasks = computed(() => [...tasks.value.values()].sort(
   (task1, task2) => task1.displayInformation.taskTitle.localeCompare(task2.displayInformation.taskTitle)
 ))
 
+function getSchemeColor(task: Task) {
```
tharvik: hoo, nice idea to vary color based on task type!

consider returning `string` has to ensure that's it won't be undef when updating `taks.trainingInformation.scheme` (same remark for `getDataTypeColor`)

```suggestion
function getSchemeColor(task: Task): string {
```

webapp/src/store/tasks.ts
```diff
@@ -22,12 +27,18 @@ export const useTasksStore = defineStore('tasks', () => {
...
+      const tasks = (await fetchTasks(CONFIG.serverUrl)).filter((t: Task) => !TASKS_TO_FILTER_OUT.includes(t.id))
       tasks.forEach(addTask)
+      loadingAlreadyFailed.value = false
     } catch (e) {
       console.error('Fetching of tasks failed with error', e instanceof Error ? e.message : e)
-      const toaster = useToaster()
-      toaster.error('The server is unreachable. \n Please try again later or reach out on slack.')
+
+      //Only display UI message once
+      if (loadingAlreadyFailed.value === false) {
```
tharvik: always def and boolean

```suggestion
      if (!loadingAlreadyFailed.value) {
```

tharvik (approved):
impressive, well done!

webapp/src/store/tasks.ts
JulienVig: Nice to remove them or to keep them?

webapp/src/assets/svg/PlugIcon.vue
JulienVig: I'll try, I was already happy to find out how to add new icons I didn't push it further haha

JulienVig: So this is just a default style and can be overriden like `<PlugIcon customClass="w-6 w-6"/>` which I think is nice

webapp/src/components/data/DataDescription.vue
JulienVig: Unfortunately there is some logic that depends on `trainingInformation` (to get the task data type)

webapp/src/components/data/dataset_input/FileSelection.vue
JulienVig: Good catch! (pun intended)
I've re-written the functions into async functions, could you have a look to check if that's what you meant?

PR #688, opened 2024-06-19
*: prerelease fixes

tharvik:
on the way to v3, we have some stuff to iron out, tracked in #686.
taking over some from there.

* update `{Trainer,Model}.fit` to return async iterators to follow progression of every part of processing. allows for more reactive UI as it refreshes on every batch rather on every round.
  * I don't really like how async iterators are handled in JS (no easy way to get return values), added an "utils" module to ease its use
  * there is not ETA but it helps users not get borded by a slow interface.
* commonize branding to have the same "DISCO" everywhere
* add demo notice
* rework some webapp internals


docs/examples/README.md
```diff
@@ -22,7 +22,9 @@ As you can see in `training.ts` a client is represented by a `Disco` object:
 
 ```js
 const disco = new Disco(task, { url, scheme: "federated" });
-await disco.fit(dataset); // Start training on the dataset
+for await (const round of disco.fit(dataset))
+  for await (const epoch of round)
+    for await (const batch of epoch);
```
martinjaggi, line 27: quick question here, why do we have epochs actually? as rounds (of aggr) can often also be much smaller than one epoch. wouldn't it be easiest to only have rounds and batches(=steps)?

slightly related to the LLM also, where it's also fine to train for less than one epoch if time is of concern?

tharvik: > quick question here, why do we have epochs actually? as rounds (of aggr) can often also be much smaller than one epoch. wouldn't it be easiest to only have rounds and batches(=steps)?

historically, it was taken from tfjs as it is working by epochs, so the same relation of "round -> epoch -> batch" was kept.
we can flatten it to have "round -> batch", that does seems easier. won't there be an issue if the dataset is not fully used? we can have batches randomly sampled from the dataset to avoid that, which would make datasets random access rather than linear, require much rework of the loaders. but that's doable.

> slightly related to the LLM also, where it's also fine to train for less than one epoch if time is of concern?

for now, it's hardcoded to only train 5 batches per epoch with gpt-tfjs, as to have a faster result (good for demo).

JulienVig: Fyi in tfjs every dataset (image text tabular) is currently already accessed in random batch order rather than linearly by default

martinjaggi: epoch has a clear meaning, that every element of the dataset is used exactly once.  so better not redefine it to be 5 batches. much easier to make your round 5 batches (or whatever value it already had now, per round)

either remove epoch entirely, or make clear it still has a precise meaning (that a round is 0.2 epochs for example, but don't say an epoch is 5 batches if that's not respecting the meaning). i'm pretty sure at some point we did allow fractional values for epoch so we could keep doing that also. as you prefer

if a training only accesses a small fraction of the data overall, that's fine too. 

if it's random or not is another story, wouldn't matter here either way

tharvik: > epoch has a clear meaning, that every element of the dataset is used exactly once. so better not redefine it to be 5 batches. much easier to make your round 5 batches (or whatever value it already had now, per round)

this redefining to at most 5 batches is only done in gpt-tfjs, to avoid taking too much time training it, the classic tfjs models are using the full dataset. it was done to avoid training for too long.

> either remove epoch entirely, or make clear it still has a precise meaning (that a round is 0.2 epochs for example, but don't say an epoch is 5 batches if that's not respecting the meaning).

I do agree that we should respect the meaning of "epoch". it's quite weird to arbitrary stop. we can remove this limitation of gpt-tfjs. showing the increasing batch count does help user with choosing to reduce the size of the training dataset (wikitext is quite large). it's reduce a bit the usability of the demo tough.

> i'm pretty sure at some point we did allow fractional values for epoch so we could keep doing that also. as you prefer

I was unable to find such fractional epoch, maybe it got removed at some point. 

> if a training only accesses a small fraction of the data overall, that's fine too.
> 
> if it's random or not is another story, wouldn't matter here either way

for sure, that would be for a next iteration.

martinjaggi: ok as you prefer then, could remove it everywhere, or keep the gpt exception (maybe add another comment there that epochs are not used.) is epochs actually in the task configs, or do we have only rounds and steps per round there?

JulienVig:
Great work!! I'll review the implementation details next week.

I noticed a few UI things that I would change, like off centered buttons, small margins, I would also want to make the Demo warning a banner rather than a Card on the home page. If you want, I'm happy to tweak that myself next week (along with other things like the feedback form)
Screenshot 2024-06-27 at 15 53 41

martinjaggi:
for the async iter every batch vs every round: does it affect training speed or not much? (thinking of the cases where local iterations might be super fast but communication might be slow, which might often be the case)

docs/examples/README.md
tharvik: added [an issue](https://github.com/epfml/disco/issues/689) as to not loss the discussion.

> is epochs actually in the task configs, or do we have only rounds and steps per round there?

in `Task`, we have `epochs` (total number of run over the whole dataset) & `roundDuration` (number of epochs before performing a round). steps are internal to gpt-tfjs (but should be flatten to reflect the `roundDuration` IMO).

> ok as you prefer then, could remove it everywhere, or keep the gpt exception (maybe add another comment there that epochs are not used.)

I was thinking of doing it after the v3 release but if you think it needs to be done before, I can get right to it.

tharvik:
> for the async iter every batch vs every round: does it affect training speed or not much? (thinking of the cases where local iterations might be super fast but communication might be slow, which might often be the case)

hum, not from what I remember. we didn't actively test with slow communications but it didn't seem like a high time cost in comparaison to the training (sub-seconds vs whole seconds). of course, if the network is very flaky, it might come as an issue.

tharvik:
> like off centered buttons

ho good catch, I mixed two tailwindcss classes, fixed in the new commits.

> small margins, I would also want to make the Demo warning a banner rather than a Card on the home page. If you want, I'm happy to tweak that myself next week (along with other things like the feedback form)

gladly, web UI is not really my strongest suit :)

JulienVig:
> gladly, web UI is not really my strongest suit :)

It's not mine either but I unfortunately have strong opinions on the matter... If that's fine with you I'll make changes in a follow-up PR rather than this one

discojs/src/models/gpt/model.ts
```diff
@@ -57,6 +57,7 @@ class GPTModel extends tf.LayersModel {
     await callbacks.onTrainBegin?.()
     
     for (let epoch = 1; epoch <= trainingArgs.epochs; epoch++) {
```
JulienVig, line 59: Something is not working well with the new iterator. Before, the preprocessing methods (tokenize and leftPadding) were called on `iterator.next()` but now they are called before (on `dataset.iterator()`?)
Potentially the only thing to change is how we measure the preprocessing time (which is always 0ms now because it's not measuring anything).

JulienVig: I'm using the cli benchmark_gpt script to check the preprocessing time.
For me ```npm -w cli run benchmark_gpt -- --batchSize 16``` used to show ~130ms of preprocessing time.

cli/src/benchmark_gpt.ts
```diff
@@ -80,10 +80,10 @@ async function main(args: Required<CLIArguments>): Promise<void> {
     console.log(`\tmodel type ${modelType} \n\tbatch size ${batchSize} \n\tcontext length ${contextLength}`)
 
     let epochTime = performance.now()
-    const logGenerator = model.train(preprocessedDataset, undefined, epoch)
-    for await (const logs of logGenerator) {
+    for (let epochsCounter = 0; epochsCounter < epochsCount; epochsCounter++) {
```
JulienVig:
```suggestion
    for (let epochsCounter = 1; epochsCounter <= epochsCount; epochsCounter++) {
```

There is a division by 0 for the first `msPerToken` if `epochCounter` starts at 0

discojs/src/models/gpt/index.ts
```diff
@@ -38,51 +45,90 @@ export class GPT extends Model {
...
+      const batch = iteration.value;
+
+      const batchLogs = await this.#runBatch(batch);
+      tf.dispose(batch);
+
+      yield batchLogs;
+      batchesLogs = batchesLogs.push(batchLogs);
+    }
+
+    const validation = validationData && (await this.#evaluate(validationData));
```
JulienVig, line 69: Fancy

discojs/src/models/gpt/index.ts
```diff
@@ -38,51 +45,90 @@ export class GPT extends Model {
...
+    if (logs === undefined) throw new Error("batch didn't gave any logs");
+
+    const { loss, acc: accuracy } = logs;
+    if (loss === undefined || isNaN(loss))
+      throw new Error("training loss is undefined or NaN");
+
+    return {
+      accuracy,
+      loss,
+      memoryUsage: tf.memory().numBytes / 1024 / 1024 / 1024,
```
JulienVig, line 95: Why not use `peakMemory` from the logs? If intentional then we can remove the `peakMemory` measurements altogether

webapp/src/components/training/TrainingInformation.vue
```diff
@@ -178,92 +173,141 @@
...
-    ? props.logs.reduce((acc, logs) => acc + logs.participants, 0) / props.logs.size
-    : 0
-  return {
-    current: props.logs.last()?.participants ?? 0,
-    average
-  }
-})
+const batchesCount = computed(() => props.batchesOfEpoch.size);
+const epochsCount = computed(() => props.epochsOfRound.size);
+const roundsCount = computed(() => props.rounds.size);
```
JulienVig, line 207: The epoch count is reset whenever the round is incremented, is it expected? When I train wikitext I get the following sequence:
round 0 - epoch 0 - batch 0
round 0 - epoch 0 - batch 1
[...]
round 0 - epoch 0 - batch 5
round 0 - epoch 1 - batch 0
[...]
round 0 - epoch 1 - batch 5
round 0 - epoch 2 - batch 0
round 1 - epoch **0** - batch 0
round 1 - epoch 0 - batch 1

Shouldn't the epoch continue incrementing rather than being reset? (Clearly there needs to be more explanations in the UI regarding rounds epochs and batches #691).

discojs/src/models/gpt/model.ts
```diff
@@ -69,18 +70,34 @@ class GPTModel extends tf.LayersModel {
...
+        tf.dispose(accSumTensor)
+        if (typeof accSum !== 'number')
+          throw new Error('got multiple accuracy sum')
+        accuracyFraction = [accuracyFraction[0] + accSum, accuracyFraction[1] + accSize];
+	tf.dispose([accTensor])
+
         const lossTensor = tf.tidy(() => {
-          const { grads, value: lossTensor } = this.optimizer.computeGradients(lossFn)
+          const { grads, value: lossTensor } = this.optimizer.computeGradients(() => {
+            const logits = this.apply(xs)
```
JulienVig, line 95: This is already computed l76, can we run inference a single time?

webapp/src/components/training/TrainingInformation.vue
```diff
@@ -178,92 +173,141 @@
...
+        validation: validation.concat([cur.validation]),
+      }),
+      {
+        training: [] as number[],
+        validation: [] as number[],
+      },
+    ),
+);
+
+const commonChartsOptions = {
```
JulienVig, line 249: I think the series are a bit shaky now and aren't as smooth as before.
Before:

https://github.com/epfml/disco/assets/33122365/3168fd4f-4dbf-4afa-a5c3-1eab29c674a4

Now (see how the animations seems to go one way and then the opposite way):

https://github.com/epfml/disco/assets/33122365/e28c0ee7-811d-4912-9444-694c003f2d7b

server/tests/e2e/federated.spec.ts
```diff
@@ -68,7 +69,10 @@ describe("end-to-end federated", function () {
 
     let logs = List<RoundLogs>()
     for await (const round of disco.fit(data)) {
-      logs = logs.push(round)
+      const [roundGen, roundLogs] = async_iterator.split(round)
+      for await (const epoch of roundGen)
+        for await (const _ of epoch);
```
JulienVig: Would there be a way to inline or shorten these 4 lines? I find this very bulky for something used so frequently.
What about having different methods (e.g. .fitByEpoch) exposing the different levels (batch/epoch/round) depending on what we need?

webapp/src/components/home/GetStarted.vue
```diff
@@ -1,52 +1,60 @@
 <template>
-  <div class="grid grid-cols-1 gap-8 md:mx-20 lg:gap-16 lg:grid-cols-3 items-stretch">
-     <!-- Model testing -->
-    <ButtonCard
-      title-placement="center"
-      @action="goToDiscoInformation()"
+  <div
+    class="grid grid-cols-1 gap-8 md:mx-20 lg:gap-16 lg:grid-cols-3 items-stretch"
+  >
+    <!-- Model testing -->
```
JulienVig:
```suggestion
    <!-- Disco information -->
```

webapp/src/components/simple/DISCOllaborative.vue
```diff
@@ -1,5 +1,5 @@
-<template>
-  <span>
-    <span class="font-disco text-disco-cyan">DIS</span>
-    <span class="font-disco text-disco-blue">CO</span>llaborative</span>
-</template>
+<template><DISCO />llaborative</template>
```
JulienVig, line 1: I think we should include a prop to make DISCOllaborative plural (right now there's just an s being added after  the component)

webapp/src/components/containers/ButtonsCard.vue
```diff
@@ -0,0 +1,54 @@
...
+      <slot name="title" />
+    </div>
+
+    <div class="text-slate-500">
+      <slot />
+    </div>
+
+    <div
+      class="flex flex-wrap mt-auto items-center"
+      :class="`justify-${buttonsJustify}`"
```
JulienVig:
```suggestion
      class="flex flex-wrap mt-auto items-center gap-x-4"
      :class="`justify-${buttonsJustify}`"
```
Suggested change:
Screenshot 2024-07-01 at 15 16 23

Currently:
Screenshot 2024-07-01 at 15 05 40

webapp/src/components/home/Landing.vue
```diff
@@ -1,34 +1,39 @@
...
       </template>
-      <template #text>
-          <p>✅ &nbsp;Data stays private and offline</p>
-          <p>✅ &nbsp;Code-free and installation-free</p>
-          <p>✅ &nbsp;Collaborative training via federated and decentralized learning</p>
-      </template>
-      <template #button>
-        get started
+
+      <ul style="list-style: &quot;✅&quot;">
```
JulienVig:
```suggestion
      <ul style="list-style: &quot;✅&quot;" class="ml-5">
```
The list style bullets don't respect the parent padding

webapp/src/components/information/Tutorial.vue
```diff
@@ -27,7 +27,7 @@
         </div>
       <div class="flex flex-wrap items-center justify-center space-x-8">
         <CustomButton @click="router.push('/list')" class="mb-2">
-          explore DISCOllaboratives
+          explore <DISCOllaborative />
```
JulienVig:
```suggestion
          explore <DISCOllaborative />s
```

webapp/src/components/pages/TaskList.vue
```diff
@@ -32,10 +27,10 @@
             <Tasks/>
           </template>
           <template #content>
-          <DISCOllaborative/> are machine learning tasks, such as diagnosing COVID from ultrasounds or classifying hand written digits, that users can join to train and contribute to with their own data. To give you a sense of DISCO, we pre-defined some tasks
+            <DISCOllaborative/> are machine learning tasks, such as diagnosing COVID from ultrasounds or classifying hand written digits, that users can join to train and contribute to with their own data. To give you a sense of <DISCO />, we pre-defined some tasks
```
JulienVig:
```suggestion
            <DISCOllaborative/>s are machine learning tasks, such as diagnosing COVID from ultrasounds or classifying hand written digits, that users can join to train and contribute to with their own data. To give you a sense of <DISCO />, we pre-defined some tasks
```

JulienVig (approved):
Great work!! Mostly minor comments but a few important ones:
* the preprocessing time measurement are now wrong
* the chart animations are jerky

discojs/src/models/gpt/index.ts
tharvik: :tipping_hand_person:

discojs/src/models/gpt/index.ts
tharvik: ho right, that's what I had in mind when adding `memoryUsage` but forgot to remove it in the end :sweat_smile:

discojs/src/models/gpt/model.ts
tharvik: hum, that's shouldn't have happend. I'm not really sure of what I changed to have that. anyway, I'm now expliciting the preprocessing by waiting on tensor's `data`.

discojs/src/models/gpt/model.ts
tharvik: yeah, I don't like it either. when trying to pull out the inference, `computeGradients` complains about it being unable to correlate input and output. after reading tfjs' code, the way that accuracy is computed is during training, by attaching metrics as tensors to the model.
I can try some shenaniganz

---

soo, it tried some and was able to remove the whole `fitDataset` implementation and simply using the base one. it seems to be still achieving the same results and gives a speedup of ~10% which is nice.
I don't know if clipping the gradient weight by the global norm was specifically needed but it's unused now, as the custom evaluation logic.
please do tell me if I broke smth.

JulienVig: Hmm I think the whole point of using the custom implementation was to add the gradient clipping. I couldn't find a clear answer quickly, can we move that change to a later time so I have more time to research and test that?

server/tests/e2e/federated.spec.ts
tharvik: yeah, it's quite ugly, javascript doesn't yet have a good way to deal with async iterators. the async_iterator module was to help a bit with it but I agree, it's way too verbose.
following your suggestion, I added a few helpers: `trainBy{Round,Epoch,Batch}` yielding corresponding logs, `trainFully` that runs everything but as a simple `Promise`. also rename "fit" to "train" which I find more understandable from a non-ML POV.

webapp/src/components/simple/DISCOllaborative.vue
tharvik: good idea; I created a new component "DISCOllaboratives", which simply adds the S (a prop is kinda harder to read)

discojs/src/models/gpt/model.ts
tharvik: alright too bad, reverting that (I was quite proud of finding that :frowning:)

webapp/src/components/training/TrainingInformation.vue
tharvik: moving discussion to issue, but yeah, it's excepted in my understanding.

webapp/src/components/training/TrainingInformation.vue
tharvik: haa fun, this one was subtle: during a batch/epoch/round, the logs are added to the training information. when on ends, the resulting upper-layer logs are also added to the logs and the current logs are reset _but only after a new epoch/round is started_! so the components renders as been at the end of the next round then rapidly returns to the beginning of the next round thus the shakiness :tornado: 
fixed in latest commit

discojs/src/models/gpt/model.ts
JulienVig: Hopefully, we can use that in the near future

PR #692, opened 2024-07-01
UI tweaks

### LINKED ISSUE #691: Minor UI changes for the release
- [x] On the training page, add hover explanations for "round", "batch" and "epoch" and their relationships with each other.
- [x] Display a banner saying that Disco is currently in a demo state
- [x] Add a link to a feedback form
- [x] Add user feedback when downloading a model
- [x] Add LLM training ETA -> display the total number of epochs
- [x] Clarify the use of "epochs" for the language modeling task (which aren't really epochs)
- [x] Display LLM inference message that currently only the CLI inference is available

JulienVig:
Closes #691 

Screenshot 2024-07-09 at 16 56 05

Screenshot 2024-07-09 at 16 56 11

Screenshot 2024-07-09 at 17 00 30

Screenshot 2024-07-09 at 17 00 20

Changed the training board metrics and added tool tip informations
Screenshot 2024-07-09 at 17 14 47
Screenshot 2024-07-09 at 17 14 49
Screenshot 2024-07-09 at 17 14 51
Screenshot 2024-07-09 at 17 14 53


webapp/package.json
```diff
@@ -13,7 +13,7 @@
   "dependencies": {
     "@epfml/discojs": "*",
     "@epfml/discojs-web": "*",
-    "apexcharts": "3",
+    "apexcharts": "3.50",
```
tharvik: this means "this specific minor version", not "3.50 and up". did the API change so that it's not usable before 3.50? if yes, better to use `^3.50`, else "package-lock.json" is where we kinda state the version we are supporting so keeping it as `3` would be best.

webapp/src/assets/svg/PerformanceIcon.vue
```diff
@@ -17,7 +17,7 @@
 </template>
 <script lang="ts">
 export default {
-  name: 'Performances',
+  name: 'PerformanceIcon',
```
tharvik: NB: using Vue's composition API (setup based script) instead of the options does avoid having to rename in code (and I personally find it nicer)

webapp/src/components/progress_bars/ProgressIcon.vue
```diff
@@ -34,10 +35,14 @@
...
+import { toRef } from 'vue';
 interface Props {
   active: boolean
-  lined: boolean
+  currentStep: boolean
+  hasLeftLine: boolean
 }
 
 const props = defineProps<Props>()
+const isCurrentStep = toRef(props, 'currentStep')
```
tharvik, line 46: looks weird to me, why do you need it? consider changing the "currentStep" instead and use it directly

webapp/src/components/progress_bars/TrainingBar.vue
```diff
@@ -1,50 +1,42 @@
...
     <div
-      v-if="scheme !== undefined && displayTitle"
+      v-if="taskTitle !== undefined && displayTitle"
       class="flex flex-wrap font-disco text-3xl justify-center"
     >
-      <span class="text-disco-blue" style="text-transform: capitalize">{{ scheme }}</span>
-      <span class="text-disco-cyan">&nbsp;Learning</span>
+      <span class="text-disco-blue">{{ taskTitle }}</span>
+      <!-- <span class="text-disco-blue" style="text-transform: capitalize">{{ scheme }}</span>
+      <span class="text-disco-cyan">&nbsp;Learning</span> -->
```
tharvik: rm

webapp/src/components/progress_bars/TrainingBar.vue
```diff
@@ -1,50 +1,42 @@
...
-              viewBox="0 0 24 24"
-              width="24"
-              height="24"
-            >
-              <path
-                class="heroicon-ui"
-                d="M5 3h14a2 2 0 0 1 2 2v14a2 2 0 0 1-2 2H5a2 2 0 0 1-2-2V5c0-1.1.9-2 2-2zm14 8V5H5v6h14zm0 2H5v6h14v-6zM8 9a1 1 0 1 1 0-2 1 1 0 0 1 0 2zm0 8a1 1 0 1 1 0-2 1 1 0 0 1 0 2z"
-              />
-            </svg>
+            <TasksIcon  customClass="w-full w-7 h-7" view-box="-4 -4 24 24"/>
```
tharvik: (nitpicking: double space)
is it not working as wanted to simply set the class on the component?

```suggestion
            <TasksIcon class="w-full w-7 h-7" view-box="-4 -4 24 24"/>
```

webapp/src/components/simple/DISCOllaborative.vue
```diff
@@ -1,4 +1,6 @@
-<template><DISCO />llaborative</template>
+<template>
+  <DISCO /><span class="font-disco">llaborative</span>
```
tharvik, line 2: nitpicking: using `span` creates separated entities, so "disco" then "llaborative", instead of a single word. that does matter on screen readers and other accessibility tools.

```suggestion
  <span class="font-disco"><DISCO />llaborative</span>
```

webapp/src/components/simple/DISCOllaboratives.vue
```diff
@@ -1,4 +1,4 @@
-<template><DISCOllaborative />s</template>
+<template><DISCOllaborative /><span class="font-disco">s</span></template>
```
tharvik:
```suggestion
<template><span class="font-disco"><DISCOllaborative />s</span></template>
```

webapp/src/components/testing/Testing.vue
```diff
@@ -129,6 +129,17 @@
           As such, please ensure your dataset of choice was not used during the training phase of your model.
         </template>
       </IconCard>
+      <!-- Language model prompting is currently unavailable   -->
+      <div 
+        v-if="currentTask.id === 'llm_task' && validationStore.isOnlyPrediction"
```
tharvik: more resilient to changes

```suggestion
        v-if="currentTask.trainingInformation.dataType === 'text' && validationStore.isOnlyPrediction"
```

webapp/src/components/testing/Testing.vue
```diff
@@ -232,17 +245,32 @@ onActivated(async () => {
...
+    const scrollableDiv = document.getElementById('scrollable-div')
+    if (scrollableDiv !== null) {
+      scrollableDiv.scrollTo({
+        top: 0,
+        left: 0,
+        behavior: "smooth",
+      });
+    }
+  } catch (e) {
+    toaster.error("Something went wrong, please try again later.")
```
tharvik, line 269: WDYT of showing the error message in the toaster? I find myself wanting to know why a particular thing failed. you think that any remotly dev people would open the devtool (so the `console.error` below is enough)? I know we didn't do that anywhere else but maybe if people see ~"something went wrong: *host unreachable*" they'll check their own network before giving up/sending feedback?

webapp/src/components/training/Trainer.vue
```diff
@@ -27,11 +27,27 @@
...
+        <p class="text-slate-600 text-xs pt-0.5">In this live demo, the model you are training is a newly initialized one. 
+          In a real use case you would start training with the latest model resulting from all users' collaborative training. 
+          To persist collaborative models, you can launch your own DISCO instance following
+          <a
+          class='underline text-blue-400 font-bold'
+          target="_blank"
+          href="https://github.com/epfml/disco/blob/develop/DEV.md"
+          >these steps.</a>
+          <!-- Warning about the maximum nb of iteration per epoch for LLMs -->
+          <span v-if="props.task.id === 'llm_task'" class="text-slate-600 text-xs"> Additionally, when training language models we have limited the number of batches per epoch to 10.</span>
```
tharvik:
```suggestion
          <span v-if="props.task.trainingInformation.dataType === 'text'" class="text-slate-600 text-xs"> Additionally, when training language models we have limited the number of batches per epoch to 10.</span>
```

webapp/src/components/training/TrainingInformation.vue
```diff
@@ -174,37 +175,76 @@
 
 <script setup lang="ts">
 import { List } from "immutable";
-import { computed } from "vue";
+import { computed, onMounted } from "vue";
 import ApexChart from "vue3-apexcharts";
+import tippy from 'tippy.js'
```
tharvik: consider using [vuejs wrapper](https://vue-tippy.netlify.app/) as to more declarative and avoid using lifecycle hooks (`onMounted`)

webapp/src/components/training/TrainingInformation.vue
```diff
@@ -174,37 +175,76 @@
...
-      ? props.rounds.reduce((acc, round) => acc + round.participants, 0) /
+const participants = computed(() => {
+  const roundParticipants = props.rounds.last()?.participants
+  let nbOfParticipants;
+  if (roundParticipants !== undefined) {
+    nbOfParticipants = roundParticipants
+  } else {
+    // 1 if started training (potentially in local) otherwise 0
+    nbOfParticipants = props.messages.size > 0 ? 1 : 0
+  }
```
tharvik: it's weird to check for `props.messages` to know that it started. logging can (and hopefully will) change.
we can default to 1 participant when nothing started, as the one looking at the page is already kinda participating?

if you want to keep using `props.messages`, you simplify the expression anyway
```ts
// 1 if started training (potentially in local) otherwise 0
const nbOfParticipants =
  props.rounds.last()?.participants ??
    props.messages.size > 0 ? 1 : 0
```

discojs/src/client/event_connection.ts
tharvik, line 1: NB: this timeout-via-exception is getting out of hand. it's not a rare event but should be seen as an expected return value. in a next iteration :crossed_fingers:

webapp/src/components/data/dataset_input/DatasetInput.vue
```diff
@@ -192,7 +200,7 @@ import FileSelection from './FileSelection.vue'
 interface Props {
   task: Task
   datasetBuilder: data.DatasetBuilder<File>
-  isOnlyPrediction: boolean //TODO: seems to always be false
+  isOnlyPrediction: boolean //is true when predicting with a trained model
```
tharvik, line 203: haha, yeah, I marked as TODO a while back and it got out-of-sync with my current understanding, thanks!

tharvik (approved):
superb work as usual! it's getting a more professional look-and-feel with every PR, love the auto-scroll btw

webapp/package.json
JulienVig: Ah yes I forgot to remove the minor version from package.json

webapp/src/components/progress_bars/ProgressIcon.vue
JulienVig: Yea I'm not sure of that but I first tried using currentStep directly and it wasn't reactive so it was stuck to step 0.
https://stackoverflow.com/questions/70631273/vuejs3-reactivity-of-props-object

webapp/src/components/simple/DISCOllaborative.vue
JulienVig: Oh good to know

webapp/src/components/testing/Testing.vue
JulienVig: I think it's bad practice to dump error messages in a user feedback, in particular it can comprise security and privacy. If the error can be solved user-side we should handle the error and state potential solutions, if it can't be then I don't think we should display the error (like a TypeError is not useful to show if we already tell them it's on our side)

webapp/src/components/training/TrainingInformation.vue
JulienVig: I agree it's a really hacky way but I think that clicking on "train" and seeing the number of participant increase from 0 to 1 is useful user feedback. I found it confusing when it was defaulting to 1, but that's personal.

I de-inlined the conditional statement to make it more legible but I can inline it back

What do you think of propagating a boolean prop to state when we are currently training?

webapp/src/components/data/dataset_input/DatasetInput.vue
JulienVig: Oh I thought I was the one to write the previous TODO haha

webapp/src/components/progress_bars/TrainingBar.vue
JulienVig: Not really when I tried and it seems to be a common issue: https://dev.to/saifobeidat/override-tailwind-classes-of-a-reusable-vue-component-without-using-props-660

JulienVig:
@tharvik I'm merging this PR right now to have more time for internal feedback and I'll create a new PR to address your answers if needed (as well as feedback)

webapp/src/components/testing/Testing.vue
tharvik: make sense, thanks for the input!

webapp/src/components/training/TrainingInformation.vue
tharvik: > What do you think of propagating a boolean prop to state when we are currently training?

an [event](https://vuejs.org/guide/components/events.html) is better in this case. a prop can't be modified from inside a component, only from the outside.
btw, if you want two way bindings, look into [v-models](https://vuejs.org/guide/components/v-model.html), but that's doesn't seems to fit here.

PR #693, opened 2024-07-03
add auto-release

tharvik:
tried out releasing v2.2.0 & v2.2.1 and it seems to be working.
* container image published with tag 2.2.0 & 2.2.1, "develop" image is now called edge
* npm packages have two new version 2.2.0 & 2.2.1
* we can easily [write a new relase](https://github.com/epfml/disco/releases/new) based on theses tag

so when pushing a commit and tag with npm version bumps for main packages, it works. we can release v3.0.0 anytime.

### DIFF
--- a/.github/workflows/publish-deploy.yml
+++ b/.github/workflows/publish-deploy.yml
@@ -2,6 +2,7 @@ name: publish-deploy
 on:
   push:
     branches: [develop]
+    tags: [v*]
 
 jobs:
   publish-npm:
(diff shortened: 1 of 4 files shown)


JulienVig (approved):
Nice! 
Is it triggered by merging a commit message like "v3.0.0" into the develop branch? Is there an exact format to follow?
Could you add the details of the publishing process somewhere in the doc?

tharvik:
> Is it triggered by merging a commit message like "v3.0.0" into the develop branch? Is there an exact format to follow?

it is using [git tag](https://git-scm.com/book/en/v2/Git-Basics-Tagging) to know which commit is to be released. with our CI, it needs to have the `v$version` format to be picked up (it's quite used all around).

> Could you add the details of the publishing process somewhere in the doc?

good idea, I rewrote the docs' deployement file to explain the process in more depth.

Issue #631, opened 2024-02-13
use `prettier` to format *.{ts,json,md,vue}


tharvik:
currently, formatting is done using `eslint` & `standard`. the first one is a linter actually warning when a line doesn't fit the formatting specs, but doesn't enforce a single representation of a code (one can add empty lines everywhere, or disabling eslint on a line disallow formatting). it's also only for JS/TS, not json, not markdown.

[`prettier`](https://prettier.io/) is way better in my opinion, by ensuring that for a given code/AST, it always yield the same output, folding lines as needed, replacing quotes or adding/removing trailing commas automatically. it also works with many filetypes (looking at you vue), and is opiniated so no more bikesheding. it also provide a clear separation between formatting (nice code writing) and linting (bugs lurking), which is confortable.

tharvik:
standard seems to be dying: https://github.com/standard/standard/issues/1948

JulienVig:
I updated your link so that our mention doesn't appear in their discussion feed.
Edit: it seems it is still in their feed, you may have to delete your comment to remove it.
fyi https://github.com/orgs/community/discussions/23123

tharvik:
standard seems to be dying: [https://](https://www.github.com/standard/standard/issues/1948)[www.github.com/standard/standard/issues/1948](https://www.github.com/standard/standard/issues/1948)

tharvik:
> I updated your link so that our mention doesn't appear in their discussion feed.

haha, keeping github clean, that's nice

> Edit: it seems it is still in their feed, you may have to delete your comment to remove it. fyi https://github.com/orgs/community/discussions/23123

arf, it didn't work..

tharvik:
I've been using [biome](https://biomejs.dev/) instead now, which is way faster than prettier.

JulienVig:
Is there a way to use biome that uses tabulation compatible with github displays?

tharvik:
yeah, it seems that [using an editorconfig file does the trick](https://github.com/isaacs/github/issues/170#issuecomment-150489692)

JulienVig:
Nice! 🎉 I just saw on Biome's [doc](https://biomejs.dev/guides/big-projects/) that "Biome doesn’t support monorepos very well due to some limitations in resolving nested configuration files [#2228](https://github.com/biomejs/biome/issues/2228)", did you have any issue?

tharvik:
[...]

PR #697, opened 2024-07-11
Basic Testing for Discojs

josh-freeman:


discojs/src/aggregator.spec.ts
```diff
@@ -20,10 +20,53 @@ AGGREGATORS.forEach(([name, Aggregator]) =>
...
+    })
+
+    it("model correctly initialized", async () => {
+      const aggregator = new Aggregator();
+
+      const model = mock(Model);
+
+      aggregator.setModel(instance(model));
+      expect(aggregator.model).to.equal(instance(model));
+    });
```
tharvik: yeah, no need to test for that, aggregator would need to be redone at some point and avoiding to have any reference to the task or the model will be the first thing

discojs/src/aggregator.spec.ts
```diff
@@ -1,6 +1,6 @@
 import { expect } from "chai";
 import { Map, Range, Set } from "immutable";
-
+import { mock, instance, when, anything } from 'ts-mockito';
```
tharvik: hum, discojs is still using mocha/chai but if we want new features not present in theses, we will rather move to [vitest](https://vitest.dev/) (already used in the webapp & discojs-node) which provides mocking, parallel testing, good typescript support, …
that said, after the other comments, I don't think you'll even need "ts-mockito" anymore

discojs/src/aggregator.spec.ts
```diff
@@ -20,10 +20,53 @@ AGGREGATORS.forEach(([name, Aggregator]) =>
   describe(`${name} implements Aggregator contract`, () => {
     it("starts at round zero", () => {
       const aggregator = new Aggregator();
-
+
```
tharvik, line 23: that's not needed (and is being shown as very red in my `git diff`).

we don't enforce a formatter [yet](https://github.com/epfml/disco/issues/631) but I'm all for slowly moving to prettier (that is, when you add changes, prettify it around, but not the whole file if you only changed a line). I'm hopeful that the transition will be smoother this way.

discojs/src/aggregator.spec.ts
```diff
@@ -20,10 +20,53 @@ AGGREGATORS.forEach(([name, Aggregator]) =>
...
+      expect(aggregator.model).to.equal(instance(model));
+    });
+
+    it("is full when created with no nodes", () => {
+      const aggregator = new Aggregator();
+
+      expect(aggregator.isFull());
+    });
+
+    it("is not full when created with more than no nodes and empty", () => {
```
tharvik, line 48:
```suggestion
    it("is empty when without contributions", () => {
```

discojs/src/aggregator.spec.ts
```diff
@@ -20,10 +20,53 @@ AGGREGATORS.forEach(([name, Aggregator]) =>
...
+      const aggregator = new Aggregator();
+      aggregator.setNodes(Set.of("client 0"));
+
+      expect(aggregator.isFull()).to.be.false;
+    });
+
+
+
+    it("is full when enough contributions", () => {
```
tharvik, line 57:
```suggestion
    it("is full with enough contributions", () => {
```

discojs/src/aggregator.spec.ts
```diff
@@ -35,10 +78,36 @@ AGGREGATORS.forEach(([name, Aggregator]) =>
           aggregator.add(`client ${i}`, WeightsContainer.of([i]), 0, r);
 
       await results; // nothing to test
-
+
```
tharvik, line 81: ws

discojs/src/aggregator.spec.ts
```diff
@@ -20,10 +20,53 @@ AGGREGATORS.forEach(([name, Aggregator]) =>
...
+      aggregator.setNodes(Set.of("client 0", "client 1", "client 2"));
+
+      for (let i = 0; i < 3; i++)
+        for (let r = 0; r < aggregator.communicationRounds; r++)
+          aggregator.add(`client ${i}`, WeightsContainer.of([i]), 0, r);
+
+      expect(aggregator.isFull()).to.be.true;
+    });
+
+
```
tharvik, line 68: whitespaces

```suggestion
```

discojs/src/aggregator.spec.ts
```diff
@@ -20,10 +20,53 @@ AGGREGATORS.forEach(([name, Aggregator]) =>
...
+    });
+
+    it("is not full when created with more than no nodes and empty", () => {
+      const aggregator = new Aggregator();
+      aggregator.setNodes(Set.of("client 0"));
+
+      expect(aggregator.isFull()).to.be.false;
+    });
+
+
```
tharvik, line 55: too much whitespaces
```suggestion
```

discojs/src/aggregator.spec.ts
```diff
@@ -35,10 +78,36 @@ AGGREGATORS.forEach(([name, Aggregator]) =>
           aggregator.add(`client ${i}`, WeightsContainer.of([i]), 0, r);
 
       await results; // nothing to test
-
+    
       expect(aggregator.round).to.equal(1);
     });
 
+    it("does not move forward with not enough contributions", async () => {
```
tharvik, line 85: double negation are hard on my brain. it also feels as a duplicate of the previous one. you can merge the two and check that it didn't advance with only two clients contributions but does with three.

discojs/src/aggregator.spec.ts
```diff
@@ -35,10 +78,36 @@ AGGREGATORS.forEach(([name, Aggregator]) =>
...
+      for (let i = 0; i < 2; i++)
+        for (let r = 0; r < aggregator.communicationRounds; r++)
+          aggregator.add(`client ${i}`, WeightsContainer.of([i]), 0, r);
+
+      await results; 
+
+      expect(aggregator.round).to.equal(0);
+    });
+
+    it("Adding at the wrong round does not count", () => {
```
tharvik, line 100: reads more nicely. also, you can keep the test itself smaller by not having three clients but only one. less code == faster to understand => less bug.

```suggestion
    it("drops contributions for futur rounds", () => {
```

package.json
```diff
@@ -11,7 +11,8 @@
     "webapp"
   ],
   "dependencies": {
-    "immutable": "4"
+    "immutable": "4",
+    "ts-mockito": "^2.6.1"
```
tharvik, line 15: the root package.json is hardly ever changed, especially for dev deps used in a single package. I use it for syncing version of dependencies that are needed across many packages.
please modify discojs/package.json if you want to add a package for discojs. (which you probably don't need as stated previously)

discojs/src/weights/aggregation.spec.ts
```diff
@@ -4,33 +4,68 @@ import { WeightsContainer, aggregation } from './index.js'
...
+      [[2, 13, 4], [0, 1]]
+    ]))
+  })
+
+  it('diff of weights with different dimensions throws an error', () => {
+    assert.throws(() => aggregation.diff([
+      [[3, -4], [9]],
+      [[2, 13, 4], [0, 1]]
+    ]))
+  })
```
tharvik, line 68: this is a bench test, you can simplify it by using a `Set(…).forEach((op) => …)`

while at it, should it fail? an addition without value can be zero, a multiplication can be one; why is this test needed?

```suggestion
  Set.of(aggregation.avg, aggregation.sum, aggregation.diff).forEach((op) => {
    it(`${op.name} of weights with no operands throws`, () =>
      expect(() => op([])).to.throw());

    it(`${op.name} of weights with different dimensions throws`, () =>
      expect(() => op([[[1, 2]], [[3, 4, 5]]])).to.throw());
  });
```

tharvik (changes requested):
thanks for the contribution! some stuff to iron out, I hope that the remarks will help.

you didn't explicitly ask for a review but your PR wasn't in draft so I though I'll take a look
also, the CI failed with some lint error, you might want to look into it.

btw, the branch name is not following [our guidelines](https://github.com/epfml/disco/blob/develop/docs/CONTRIBUTING.md#1-creating-a-new-branch), its format is `$issue-$description-in-very-few-workd-$username`

discojs/src/aggregator.spec.ts
josh-freeman: thank you. It's the first time I work on GitHub with PRs that are not made by cloning for a PR. Should I make a new branch from this branch for the PR?

tharvik: I'm not sure that I follow you: this is already a PR, you can added changes by pushing new commits to the same branch ("test_components") and it'll show up here.

tharvik:
aggregators where redo recently so outdated

PR #698, opened 2024-07-16
dependabot: fix for monorepo

tharvik:
default dependabot configs [wrongly ask to update non-existing package-lock in workspaces](https://github.com/dependabot/dependabot-core/issues/4993). implementing workaround.


JulienVig (approved):
All good, thanks for the quick fix!

PR #706, opened 2024-07-16
docker: bump base image

tharvik:
#702 tried to auto-upgrade base image but fails to do so.
working around unfixed upstream bug and upgrading.


JulienVig (approved):
LGTM!

PR #708, opened 2024-07-16
Fix decentralized learning fail

### LINKED ISSUE #667: Decentralized learning fails with error
Training collaboratively for decentralized learning tasks fails with error:

```
Uncaught Error: Connection failed.
    _onConnectionStateChange simplepeer.min.js:6
    onconnectionstatechange simplepeer.min.js:6
    _onConnectionStateChange simplepeer.min.js:6
    onconnectionstatechange simplepeer.min.js:6
    dispatchEvent PeerConnection.sys.mjs:836
    dispatchEvent PeerConnection.sys.mjs:1822
    onStateChange PeerConnection.sys.mjs:1952
```
The error arises right after clicking Train collaboratively, before any training update happens.

**To Reproduce**
1. Start a server and two web-clients
2. Connect some data for a decentralized task (I used the lus_covid tasks and modified it to be decentralized)
3. Clicking train collaboratively for both web-clients.
4. Check the browser console logs where the error should appear

### LINKED ISSUE #694: Decentralized training on MNIST fails with error
Starting a collaborative training on the decentralized MNIST task fails during the first round. There are a few things not working well:
* It takes a long time before anything happens after hitting 'Train collaboratively' (while training starts immediately when training in local)
* When training with two local clients, the number of participants shown is 16
* When we get to the first communication round, the training fails with a timeout error thrown by the `waitForPeers` function.

Potentially linked to #667

### LINKED ISSUE #707: The number of participants is wrong for both federated and decentralized learning
- [ ] In federated learning, the number of participants is always two because each client only has the federated server in its nodes (and adds itself to the count). The server should return the total number of participants for each round along with the model updates.
- [ ]  In decentralized learning, the number of participants is too high by 1 because each client adds itself to the count of participants while it is already present in its list of nodes.
- [ ] There are also occurrences of the number of participants shooting to 16 when training with only two clients.

JulienVig:
- [x] Improve the aggregator's logic
- [x] Fix the displayed number of participants in federated and decentralized learning. Closes #707 
- [x] Closes #694 and Closes #667 
- [x] Add CLI support for decentralized tasks


discojs/src/aggregator/get.ts
```diff
@@ -1,33 +1,60 @@
...
+ * @param options Options passed down to the aggregator's constructor
  * @returns The aggregator
  */
-export function getAggregator (task: Task): aggregator.Aggregator {
-  const error = new Error('not implemented')
+export function getAggregator(task: Task, scheme?: Required<Task['trainingInformation']['scheme']>,
+  options?: AggregatorOptions): aggregator.Aggregator {
+  if (options === undefined) {
+    options = {} // init empty arguments if not specified
+  }
```
tharvik: default arguments behave in the same way

```suggestion
  options: AggregatorOptions = {}): aggregator.Aggregator {
```

discojs/src/aggregator/get.ts
```diff
@@ -1,33 +1,60 @@
...
+ * If `scheme` is undefined, we rely on task.trainingInformation.scheme to infer `scheme`.
+ * 
+ * @param task The task object associated with the current training session
+ * @param scheme The training scheme. If not specified, the task training information scheme is used
+ * @param options Options passed down to the aggregator's constructor
  * @returns The aggregator
  */
-export function getAggregator (task: Task): aggregator.Aggregator {
-  const error = new Error('not implemented')
+export function getAggregator(task: Task, scheme?: Required<Task['trainingInformation']['scheme']>,
```
tharvik: why not having the scheme inside the options argument? it is not required anyway

discojs/src/aggregator/get.ts
```diff
@@ -1,33 +1,60 @@
...
- */
-export enum AggregatorChoice {
-  MEAN,
-  SECURE,
-  BANDIT
+type AggregatorOptions = {
+  model?: Model,
+  roundCutOff?: number, // MeanAggregator
+  threshold?: number, // MeanAggregator
+  thresholdType?: 'relative' | 'absolute', // MeanAggregator
```
tharvik: typescript trick: you can use [`Partial`](https://www.typescriptlang.org/docs/handbook/utility-types.html#partialtype) to have an object without all it's field as optional.

discojs/src/aggregator/get.ts
```diff
@@ -1,33 +1,60 @@
...
-      if (task.trainingInformation.scheme !== 'decentralized') {
+    case 'mean':
+      return new aggregator.MeanAggregator(options.model, options.roundCutOff, options.threshold, options.thresholdType)
+    case 'secure':
+      if (scheme !== 'decentralized') {
         throw new Error('secure aggregation is currently supported for decentralized only')
       }
-      return new aggregator.SecureAggregator()
+      return new aggregator.SecureAggregator(options.model, task.trainingInformation.maxShareValue)
     default:
```
tharvik: if the aggregator is not defined in the task, we disregard the options altogether, that's unexcepted.

discojs/src/aggregator/mean.ts
```diff
@@ -4,33 +4,87 @@ import { AggregationStep, Base as Aggregator } from "./base.js";
...
+    else if (threshold > 1) {
+      // Print a warning if threshold and thresholdType are conflicting
+      if (thresholdType === 'relative') {
+        console.warn(
+          "[WARN] Changing the aggregator's thresholdType from 'absolute' to 'relative' because threshold < 1 " +
+          `Make sure that the aggregator is indeed only expecting ${threshold} contributions.`
+        )
+      }
+      this.#thresholdType = 'absolute'
+    }
```
tharvik: hum, warning instead of throwing when passed bad arguments feels icky. consider failing instead. we can always show it more nicely in the various user interfaces but the library itself shouldn't be too smart about it.

discojs/src/client/decentralized/base.ts
```diff
@@ -20,80 +20,18 @@ export class Base extends Client {
    */
   private pool?: PeerPool
   private connections?: Map<NodeID, PeerConnection>
+  // set True when disconnected is called
+  // Used to handle timeouts and promise resolving after calling disconnect
+  private disconnected = false
```
tharvik: `Client._server` (and `Client._ownID`) is undef when disconnected. consider reading it direclty or making a small helper `isDisconnected() { return this._server === undefined }`

discojs/src/client/decentralized/base.ts
```diff
@@ -134,26 +92,115 @@ export class Base extends Client {
...
+    // Wait for the server to answer with the list of peers for the round
+    try {
+      const receivedMessage = await waitMessageWithTimeout(
+        this.server, type.PeersForRound, undefined,
+        "Timeout waiting for the round's peer list"
+      )
+      
+      if (this.nbOfParticipants > 0) {
+        throw new Error('got new peer list from server but was already received for this round')
+      }
```
tharvik: can this ever happen? we only wait for a single `PeersForRound` message from the server per round.

discojs/src/client/decentralized/base.ts
```diff
@@ -134,26 +92,115 @@ export class Base extends Client {
...
+
+      if (this.ownId !== undefined && peers.has(this.ownId)) {
+        throw new Error('received peer list contains our own id')
+      }
+      // Store the list of peers for the current round including ourselves
+      this.aggregator.setNodes(peers.add(this.ownId))
+
+      if (this.pool === undefined) {
+        throw new Error('waiting for peers but peer pool is undefined')
+      }
```
tharvik: consider moving it earlier, as this is only happening if the server isn't connected.

discojs/src/client/decentralized/base.ts
```diff
@@ -174,15 +221,17 @@ export class Base extends Client {
...
                 const encoded = await serialization.weights.encode(payload)
-                this.sendMessagetoPeer(connection, {
+                const msg = {
                   type: type.Payload,
                   peer: id,
                   round: r,
                   payload: encoded
-                })
+                }
+                peer.send(msg as messages.PeerMessage)
```
tharvik: avoid cast if it can be. I'm now trying to comment everywhere I'm using `as` (bad keyword, bad)

you can instead have `const msg: messages.PeerMessage = …`

discojs/src/client/decentralized/base.ts
```diff
@@ -193,13 +242,19 @@ export class Base extends Client {
...
+      try { 
+        result = await Promise.race([
+          this.aggregationResult,
+          timeout(undefined, "Timeout waiting on the aggregation result promise to resolve")
+        ])
+      } catch (e) {
+        if (this.disconnected) {
+          return
+        }
+        console.error(e instanceof Error ? e.message : e)
```
tharvik:
```suggestion
        console.error(e)
```

discojs/src/client/decentralized/base.ts
```diff
@@ -193,13 +242,19 @@ export class Base extends Client {
...
-      result = await Promise.race([
-        this.aggregationResult,
-        timeout(undefined, "Timeout waiting on the aggregation result promise to resolve")
-      ])
+      try { 
+        result = await Promise.race([
+          this.aggregationResult,
+          timeout(undefined, "Timeout waiting on the aggregation result promise to resolve")
+        ])
+      } catch (e) {
```
tharvik, line 246: hum, so if this timeouts, we will still procede to the next communication round, but the payload will be invalid (already used by current round and very probably changed by it). I think breaking would be safer.

also, by creating the timeout in there, it is a timeout per communication round, not for the whole onRoundEnd, which is a bit unexcepted.

discojs/src/client/decentralized/peer.ts
```diff
@@ -96,15 +96,7 @@ export class Peer {
...
-
-    // in the perfect world of bug-free implementations
-    // we would return this.bufferSize
-    // sadly, we are not there yet
-    //
-    // based on MDN, taking 16K seems to be a pretty safe
-    // and widely supported buffer size
-
-    return 16 * (1 << 10)
+    return this.bufferSize
```
tharvik, line 99: soo, does this work? :eyes:

discojs/src/client/event_connection.ts
```diff
@@ -79,8 +79,11 @@ export class PeerConnection extends EventEmitter<{ [K in type]: NarrowMessage<K>
     this.peer.send(msgpack.encode(msg))
   }
 
-  async disconnect (): Promise<void> {
-    await this.peer.destroy()
+  async disconnect(): Promise<void> {
+    await Promise.race([
+      this.peer.destroy(),
+      new Promise((res, _) => setTimeout(() => res('PeerConnection disconnection timed out'), 3000))
+    ])
```
tharvik: that's not our purview. if it's taking too long, we need to work on the wrtc stack itself. either with a documented upstream issue or even a PR.

```suggestion
    await this.peer.destroy()
```

discojs/src/client/federated/base.ts
```diff
@@ -94,57 +103,13 @@ export class Base extends Client {
...
-  }
-
   /**
    * Fetch the metadata values maintained by the federated server, for a given metadata key.
    * The values are indexed by node id.
    * @param key The metadata key
    * @returns The map of node id to metadata value
    */
-  async receiveMetadataMap(
+  async receiveMetadataMap( // TODO: never used
```
tharvik: clic clic, blast it

discojs/src/default_tasks/mnist.ts
```diff
@@ -23,13 +23,14 @@ export const mnist: TaskProvider = {
       trainingInformation: {
         modelID: 'mnist-model',
         epochs: 20,
-        roundDuration: 10,
+        roundDuration: 2,
```
tharvik, line 26: debug? it seems too frequent now.

tharvik (changes requested):
great work on making decentralized work in browser! :tada:

a few logic/safety errors to fix and it can be merged :+1:

discojs/src/aggregator/mean.ts
JulienVig: Indeed I'll throw exceptions. It's a remnant of a previous implementation where thresholdType had a default value.

discojs/src/aggregator/get.ts
JulienVig: Good catch, I'm changing this

discojs/src/default_tasks/mnist.ts
JulienVig: I don't think so, running the task in the webapp is still fast enough and I think that the more frequent the updates the better (if we can afford it time-wise)

discojs/src/client/federated/base.ts
JulienVig: 🫡

discojs/src/client/decentralized/base.ts
JulienVig: I don't think so, this is a copy paste from existing code. I'll remove it

discojs/src/client/decentralized/peer.ts
JulienVig: At least on chrome it does. Do you know a way to test if it works reliably?

discojs/src/client/decentralized/base.ts
JulienVig: Ah yes I agree we should break/return here. However I don't see why a timeout per communication round is unexpected. Shouldn't we timeout whenever we may wait on a promise forever?

tharvik: > However I don't see why a timeout per communication round is unexpected. Shouldn't we timeout whenever we may wait on a promise forever?

yes, we should timeout for sure, but as each aggregator has a various number of communication round, the "onRoundEnd" timeout varies here which I find a bit weird. you can create the timeout promise at the beginning of the function and reference the same timeout in every communication round.
but yeah, maybe I'm overthinking this (won't be the first time), if you don't find it strange, let's keep it as it is (we don't really have a policy on timeouts anyway).

discojs/src/client/decentralized/peer.ts
tharvik: meh, no, there are some services to try out with various browsers (cypress kinda does that) but testing two browsers communicating with each other is not easily provided.

JulienVig: Also works on safari and firefox (don't know how the previous ICEconnection issue got solved but it seems to be)
However it seems that using different browsers at the same time doesn't work, I get "connection closed" after they receive a contribution (and then "cannot send after peer is destroyed")

tharvik (approved):

PR #711, opened 2024-07-18
Fix preprocessing memory leaks

JulienVig:
Address preprocessing memory leaks found and fixed by Klavdiia.


cli/src/args.ts
```diff
@@ -27,7 +27,7 @@ const unsafeArgs = parse<BenchmarkUnsafeArguments>(
     task: { type: String, alias: 't', description: 'Task: titanic, simple_face, cifar10 or lus_covid', defaultValue: 'simple_face' },
     numberOfUsers: { type: Number, alias: 'u', description: 'Number of users', defaultValue: 1 },
     epochs: { type: Number, alias: 'e', description: 'Number of epochs', defaultValue: 10 },
-    roundDuration: { type: Number, alias: 'r', description: 'Round duration', defaultValue: 10 },
+    roundDuration: { type: Number, alias: 'r', description: 'Round duration', defaultValue: 2 },
```
martinjaggi: in the description can we say what the unit is of the round duration, local steps or local epochs?

martinjaggi (approved):
many thanks to @klavdiiaN , this is super valuable and tricky

Issue #613, opened 2024-01-22
Integrate Differential Privacy


JulienVig:
Differential Privacy is implemented but never used. It seems that it was intentionally removed during a code refactoring (in [this PR](https://github.com/epfml/disco/pull/578/files#diff-bb5fd2b79ca040c5cdd1e0c4ec57de0ef4cfe3413a482d6c9b4b35a38b0e93e1), l.179 of decentralized/base.ts)

- [ ] Integrate DP again in the new architecture code
- [ ] Look into how DP is implemented. It seems it is epsilon-DP that is implemented, which has a very strict privacy bound at the cost of lower performance. Most implementations use (epsilon, delta)-DP for a better trade-off between privacy and utility. However, the latter needs to track the privacy budget delta for each user. This is what the MomentAccount tracks in [Deep learning with differential privacy](https://arxiv.org/abs/1607.00133).
- [ ] Create test for the differential privacy feature

martinjaggi:
@ysfalh expressed some interest in this, as they have a [decentralized DP scheme](https://openreview.net/forum?id=5JrlywYHRi) which is easy to implement (only needs pairwise agreement on each edge, which we have anyway already from the secAgg implementation)

ysfalh:
> @ysfalh expressed some interest in this, as they have a [decentralized DP scheme](https://openreview.net/forum?id=5JrlywYHRi) which is easy to implement (only needs pairwise agreement on each edge, which we have anyway already from the secAgg implementation)

Yes, I am happy to help with adapting our [code](https://github.com/elfirdoussilab1/DECOR).

hleepfl:
Seems to have been implemented
Edit: my bad I didn't read well enough

hleepfl:
Closing since it's *mostly* implemented. Will open a separate issue for keeping track of privacy budgeting.

PR #712, opened 2024-07-21
readd some privacy

tharvik:
readding what was called differential privacy which is quite a bit harder to implement correctly (we need to know the variability of every features for one, and have a dedicated paper-based way to implement it, see #613).
I got lost along the way and changed more things, oupsi

* reexpose `privacy` module, with `addNoise` and `clipNorm` doing what was doing goo'old DP
  * with tests™©®
  * enable it in mnist task via `TrainingInformation.privacy`
* flatten `DistributedTrainer`, `LocalTrainer`, `TrainerBuilder` into a simple non-abstract `Trainer`
* change `discojs.Disco` to a more straightfoward object
  * exposing only the contained `Trainer` and not `Logger`, `Task` and others
  * have a clearer code on where the model is taken from
  * ensure that it gets passed some valid client options
* upgrade `Aggregator` to a real event emitter
  * avoid weird promises use (`makeResult`, `receiveResult` :eyes:)
  * drop `Model` link, everything goes through events now
  * drop mocking in server's tests
  * drop unused `AsyncInformant`


discojs/src/privacy.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+      await weights
+        .map((w) => w.square().sum())
+        .reduce((a, b) => a.add(b))
+        .data()
+    )[0],
+  );
+}
+
+describe("addNoise", () => {
+  it("returns the same weight with zero noise", async () => {
```
martinjaggi: would say `weight vector`

discojs/src/privacy.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+});
+
+describe("clipNorm", () => {
+  it("reduce norm for a single dimension", async () => {
+    const result = await clipNorm(1, WeightsContainer.of([2]));
+
+    expect(await WSIntoArrays(result)).to.deep.equal([[1]]);
+  });
+
+  it("keeps the same orientation", async () => {
```
martinjaggi: keeps direction unchanged

discojs/src/privacy.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  it("adds noise following a normal distribution", async () => {
+    const result = addNoise(1, WeightsContainer.of([5]));
+
+    expect((await WSIntoArrays(result))[0][0]).to.be.within(2, 8); // 99.7% of success
+  });
+});
+
+describe("clipNorm", () => {
+  it("reduce norm for a single dimension", async () => {
```
martinjaggi: for a one-dimensional vector

discojs/src/privacy.ts
```diff
@@ -1,46 +1,39 @@
...
- * The current round's weights are obtained after a single round of training, from the previous round's weights.
- * @param updatedWeights weights from the current round
- * @param staleWeights weights from the previous round
- * @param task the task
- * @returns the noised weights for the current round
- */
-export function addDifferentialPrivacy (updatedWeights: WeightsContainer, staleWeights: WeightsContainer, task: Task): WeightsContainer {
-  const noiseScale = task.trainingInformation?.noiseScale
-  const clippingRadius = task.trainingInformation?.clippingRadius
+async function frobeniusNorm(weights: WeightsContainer): Promise<number> {
```
martinjaggi, line 5: isn't this one a duplicate of norm in the spec one now?

discojs/src/privacy.ts
```diff
@@ -1,46 +1,39 @@
...
 
-  if (clippingRadius !== undefined) {
-    // Frobenius norm
-    const norm = weightsDiff.frobeniusNorm()
+/** Scramble weights */
+export function addNoise(
+  scale: number,
+  weights: WeightsContainer,
+): WeightsContainer {
+  const deviation = Math.pow(scale, 2);
```
martinjaggi: naming: `variance` instead of std dev as it's squared?

discojs/src/training/trainer.ts
```diff
@@ -0,0 +1,142 @@
...
+  }
+}
+
+async function applyPrivacy(
+  previous: WeightsContainer | undefined,
+  current: WeightsContainer,
+  options: Exclude<Task["trainingInformation"]["privacy"], undefined>,
+): Promise<WeightsContainer> {
+  const previousRoundWeights = previous ?? current.map((w) => tf.zerosLike(w));
+  const weightsProgress = current.sub(previousRoundWeights);
```
martinjaggi: excellent, clean way to keep the difference

martinjaggi (approved):
amazing PR, thanks so much. looks so much cleaner now and will be much easier to extend in the future. well done!

discojs/src/aggregator/base.ts
```diff
@@ -13,8 +13,10 @@ export enum AggregationStep {
 /**
  * Main, abstract, aggregator class whose role is to buffer contributions and to produce
  * a result based off their aggregation, whenever some defined condition is met.
+ *
+ * Emits an event whenever an aggregation step is performed.
```
JulienVig, line 17:
```suggestion
 * Emits an event whenever an aggregation step is performed. Users wait for this event to fetch the aggregation result.
```

discojs/src/aggregator/mean.ts
```diff
@@ -108,7 +108,6 @@ export class MeanAggregator extends Aggregator<WeightsContainer> {
 
     this.contributions = this.contributions.setIn([0, nodeId], contribution);
```
JulienVig, line 106: Not related to your changes but why are we storing the history of previous contributions rather than only the current one? Seems unnecessary and inefficient

discojs/src/training/trainer.ts
```diff
@@ -0,0 +1,142 @@
...
+        localWeights = await applyPrivacy(
+          previousRoundWeights,
+          localWeights,
+          this.#privacy,
+        );
+
+      const networkWeights = await this.#client.onRoundEndCommunication(
+        localWeights,
+        round,
+      );
```
JulienVig, line 97:
```suggestion
      // networkWeights is the same as localWeights when training locally
      const networkWeights = await this.#client.onRoundEndCommunication(
        localWeights,
        round,
      );
```
Very neat! No more `this.aggregator.weights.model`

discojs/src/training/trainer.ts
```diff
@@ -0,0 +1,142 @@
...
+
+      if (networkWeights !== undefined) this.model.weights = networkWeights;
```
JulienVig: Is networkWeights expected to be undefined at some point?
If my previous comment on networkWeights and locaWeights being the same when training in local is correct, maybe check if they are already equal too?

discojs/src/training/disco.ts
```diff
@@ -23,56 +28,91 @@ export interface DiscoOptions {
...
+  }
 
-      if (typeof options.url === 'string') {
-        options.url = new URL(options.url)
+  /**
+   * Connect to the given task and get ready to train.
+   *
+   * Will load the model from memory if available or fetch it from the server.
+   *
+   * @param clientConfig client to connect with or informations on how to create one.
```
JulienVig:
```suggestion
   * @param clientConfig client to connect with or parameters to create one.
```
information is always singular

discojs/src/privacy.ts
JulienVig:
```suggestion
// The Frobenius norm of a matrix is the square root of the sum of squared weights
async function frobeniusNorm(weights: WeightsContainer): Promise<number> {
```
Also tfjs has `tf.norm(x, 'fro')` which already does frobenius norm, though this would not be backend-agnostic. https://js.tensorflow.org/api/latest/#norm

JulienVig: tf.norm could be a way to test the hand-made implementation

discojs/src/privacy.ts
```diff
@@ -1,46 +1,39 @@
...
+
+/**
+ * Keep weights' norm within radius
+ *
+ * @param radius maximum norm
+ **/
+export async function clipNorm(
+  radius: number,
+  weights: WeightsContainer,
+): Promise<WeightsContainer> {
```
JulienVig:
```suggestion
export async function clipNorm(
  weights: WeightsContainer,
  radius: number,
): Promise<WeightsContainer> {
```
There is also tf.clipByValue or x.clipByValue
Nitpicking but I'm used to libraries expecting weights/data as first argument and then the actual parameters (e.g. tf.clipByValue(x, 0, 1))

server/tests/e2e/federated.spec.ts
```diff
@@ -136,17 +131,15 @@ describe("end-to-end federated", function () {
...
     await disco.close()
 
-    if (aggregator.model === undefined) {
-      throw new Error('model was not set')
-    }
     const validationLogs = logs.last()?.epochs.last()?.validation
     expect(validationLogs?.accuracy).to.be.greaterThan(0.6)
-    return aggregator.model.weights
+
+    return disco.trainer.model.weights
```
JulienVig, line 142: what do you think of adding a getter to disco to simplify this expression? Like `disco.weights`

JulienVig (approved):
Great work! The codebase is getting cleaner and cleaner

discojs/src/privacy.ts
tharvik: yep indeed. my personnal duplication-limit is at 3 for small functions, I'm extract it into a common one if we need it somewhere else.

discojs/src/privacy.ts
tharvik: ho right, that's clearer. changed the "scale" argument to "deviation" too

discojs/src/aggregator/mean.ts
tharvik: yeah, I totally agree. the issue is that this mean aggregator does more than it should, combining two features: doing a mean of weights & enabling to use weights from other rounds. we should split it. and I think that the second feature shouldn't even be in an aggregator but the responsability of the user of the aggregator (`Client` currently).
I tried to rethink a bit how such aggregation should happen, here are a few notes:
* `makePayload` and `communicationRound` are very weird, we need to get rid of it
* secure aggregation can also work across rounds, we just need to tell other nodes which weights we wants
* we need to have a clear(er) way to synchronize nodes
  * in federated, that's trivial: the server sends a `NextRound { round: number, weights: WeightContainer }` message everytime, with the aggregation of the previous round
  * in decentralized, we can use the server to orchestrate the rounds but without the weights contained in the message

discojs/src/training/trainer.ts
tharvik: > Is networkWeights expected to be undefined at some point?

~~if the aggregation failed, with a timeout for eg, it'll return `undefined`.~~

ho, you're right, it was a previous idea to return undef on failure but that's not the case anymore, it will return the local weights upon failure, thanks for pointing that out! I just added a comment on `Client.onRoundEndCommunication` stating that.

> If my previous comment on networkWeights and locaWeights being the same when training in local is correct, maybe check if they are already equal too?

correct, the local client simply returns the local weights.
hum, but I don't get the reason behind checking if the weights are already equal, WDYM?

discojs/src/training/disco.ts
tharvik: thanks for being my english Bescherelle :pray:

discojs/src/privacy.ts
tharvik: this is clearly more for ML's afficionados: the closer to know libraries, the better!

server/tests/e2e/federated.spec.ts
tharvik: weights are not really supposed to be exposed to the outside, only the model is relevant to users, so it should be a bit hard to get. at first, I didn't even wanted the trainer to be available (but some tests needed it).
I can meet you halfway by exposing `disco.model`, WDYT?
note: in my mind, `Disco` is an easy access object, for newcomers and not-so-ML people. if people want more control, they can use the trainer directly.

discojs/src/training/trainer.ts
JulienVig: I meant to check for equality to not unnecessarily copy weights if they are already the same, but I guess it's only a shallow copy so it does take much time

server/tests/e2e/federated.spec.ts
JulienVig: Oh I see no need to change it then. I was thinking this would be the API to run inference with models but it's through the validator actually.

PR #720, opened 2024-07-23
Display landing page before server boots

JulienVig:
* Rather than blocking the website to load until the server booted and pushed the discollaboratives, I'm moving the loading indicator to where the discollaboratives are actually needed: the task list page and the testing page (which list models to download)
Screenshot 2024-07-23 at 13 36 03
Screenshot 2024-07-23 at 13 36 33
* Make the DISCO GIF responsive (currently larger than the screen on small devices)
* Add a note on the collaborative button to explain how collaborative training behaves if there is a single participant.
Screenshot 2024-07-23 at 18 21 13


tharvik (approved):
perfect!

PR #724, opened 2024-07-24
release v3

tharvik:
bump major version of discojs* pkgs & update deps.
will tag commit when merged.

thanks all of you for the great work :tada:


JulienVig (approved):
All good!

PR #727, opened 2024-07-24
Fix 404 error and wrong "upload" button

### LINKED ISSUE #680: Connecting folder displays alarming "Upload" warning message
The webapp allows connecting files in two different modes:
- connecting files by label, with one drag and drop zone for each task label
- connecting a csv linking filenames to labels and then all files at once:
Screenshot 2024-05-30 at 12 00 14

<br/><br/>
When using the 2nd option, the user is allowed to select the folder containing all the files (rather than selecting the files themselves). Selecting a folder currently poses two problems:
1. The selection menu displays an "Upload button" (at least on Mac OS) while we would prefer something like "Select" or "Connect" as there is no upload happening
Screenshot 2024-05-30 at 12 00 14
<br/><br/>
2. Selecting a folder with many files displays a warning message on Firefox and Chrome (not on Safari) asking the user to confirm whether they actually want to **upload** that many files.

Screenshot 2024-05-30 at 12 00 53
<br/><br/>
Screenshot 2024-05-30 at 11 59 22

JulienVig:
* Fixes #680 which was displaying wrong "upload" button when selecting folders.
* Fix 404 error happening when refreshing the website on a task page
* Add hover effect when drag and dropping
Screenshot 2024-07-24 at 18 40 42

The progress bars need a rework, they are not resilient and it is very hard to enforce consistency as they are many endpoints to modify their state

### DIFF
--- a/webapp/src/components/data/dataset_input/DatasetInput.vue
+++ b/webapp/src/components/data/dataset_input/DatasetInput.vue
@@ -87,6 +87,7 @@
         v-show="connectImagesByGroup"
         class="grid grid-cols-1 lg:grid-cols-2 gap-4 md:gap-8"
       >
+      <!-- If we are training or evaluating and need labels -->
         <div
           v-if="requireLabels"
           class="contents"
@@ -101,6 +102,8 @@
             <template #content>
               <FileSelection
                 :label="label"
+                :is-multiple="true"
+                :info-text="true"
                 :dataset-builder="props.datasetBuilder"
                 :task="props.task"
                 :csv-rows="csvRows"
@@ -109,6 +112,7 @@
             </template>
           </IconCard>
         </div>
+        <!-- If it's for inference and we don't need labels -->
         <FileSelection
           v-else
           :dataset-builder="props.datasetBuilder"
@@ -176,7 +180,8 @@
                 @csv="setCsv($event)"
               >
                 <template #text>
-                  Either drag and drop the <b>images</b> directly or select the <b>folder</b> containing all the images referenced in the connected CSV file.
+                  Drag and drop or browse for the images referenced in the connected CSV file. 
+                  <br/>{{ browsingTip }}
                 </template>
               </FileSelection>
             </template>
@@ -202,9 +207,9 @@ interface Props {
   datasetBuilder: data.DatasetBuilder<File>
   isOnlyPrediction: boolean //is true when predicting with a trained model
 }
-
 const props = defineProps<Props>()
 
+const browsingTip = "Tip: when browsing files you can use the keyboard shortcut Ctrl + A to select all images."
 const csvRows = ref<{ filename: string, label: string }[]>()
 
 const isLabelLengthSmall = (props.task.trainingInformation.LABEL_LIST?.length ?? 0) <= 2
--- a/webapp/src/components/data/dataset_input/DatasetInput.vue
+++ b/webapp/src/components/data/dataset_input/DatasetInput.vue
@@ -108,7 +108,11 @@
                 :task="props.task"
                 :csv-rows="csvRows"
                 @csv="setCsv($event)"
-              />
+              >
+                <template #text>
+                  {{ browsingTip }}
+                </template>
+              </FileSelection>
             </template>
           </IconCard>
         </div>
--- a/webapp/src/components/data/dataset_input/FileSelection.vue
+++ b/webapp/src/components/data/dataset_input/FileSelection.vue
@@ -1,126 +1,89 @@
 <template>
-  <!-- Upload File Card-->
-  <div>
-    <article
-      aria-label="File Upload Model"
-      class="
-        h-full
-        flex flex-col
-        bg-white
-        rounded-lg
-      "
-      @drop.prevent
-      @dragover.prevent
-      @dragenter.prevent
-    >
-      <!-- scroll area -->
-      <section>
-        <header
-          v-if="!hideConnectField"
-          class="
-            border-dashed rounded-xl border-2 border-disco-cyan
-            flex flex-col
-            justify-center
-            items-center
-          "
-          @drop="async (e: DragEvent) => await dragFiles(e)"
-        >
-          <div class="flex flex-col
-            justify-center
-            items-center">
-            <p
-              class="
-                p-4
-                text-lg
-                text-disco-blue
-                flex-wrap
-                justify-center
-              "
-            >
-              <span>Drag and drop the</span>&nbsp;<span> {{ fileType }}{{ isMultiple ? 's' : '' }} or</span>
-            </p>
-            <label class="mb-6">
-              <span
-                class="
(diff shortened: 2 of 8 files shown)


webapp/src/components/training/Training.vue
```diff
@@ -38,22 +38,42 @@ import Finished from '@/components/training/Finished.vue'
...
+// Force the training store to synch with the task specified in the url
+watch(() => route.fullPath, () => {
+  if (route.fullPath !== '/list' && (trainingStore.step == 0 || route.params.id !== props.id)) {
+    setupTrainingStore()
   }
-  return task
 })
 
+onMounted(setupTrainingStore)
+
```
JulienVig: The tasksStore and trainingStore need a more time-consuming rework. I'm pushing this dirty implementation for the release until we have the time to address this.

webapp/src/components/data/dataset_input/DatasetInput.vue
```diff
@@ -202,9 +207,9 @@ interface Props {
   datasetBuilder: data.DatasetBuilder<File>
   isOnlyPrediction: boolean //is true when predicting with a trained model
 }
-
 const props = defineProps<Props>()
 
+const browsingTip = "Tip: when browsing files you can use the keyboard shortcut Ctrl + A to select all images."
```
tharvik, line 216: let's inline it (only used once, is not changing)

webapp/src/components/data/dataset_input/FileSelection.vue
```diff
@@ -145,30 +107,32 @@ interface Props {
   task: Task
   datasetBuilder: data.DatasetBuilder<File>
   csvRows: CSV,
-  isDirectory?: boolean, // is this input field accepts a directory
```
tharvik, line 148: haa thank you, this "directory vs multiple" concept was always bothering me!

webapp/src/components/training/Training.vue
```diff
@@ -38,22 +38,42 @@ import Finished from '@/components/training/Finished.vue'
 import Data from '@/components/data/Data.vue'
 
 const router = useRouter()
+const route = useRoute()
```
tharvik, line 41: same as above? I think you can merge both

webapp/src/components/training/Training.vue
```diff
@@ -38,22 +38,42 @@ import Finished from '@/components/training/Finished.vue'
...
+  // This happens when refreshing the page, every task are reset when fetched
+  if (route.name !== 'task-list') {
+    router.replace({ name: 'task-list' })
+  }
+  return undefined
+})
+
+// Addresses the case when users enter a url manually
+// Force the training store to synch with the task specified in the url
+watch(() => route.fullPath, () => {
```
tharvik, line 70: hum, from a quick look at the doc, there seems to be [a nicer way to react to route changes](https://router.vuejs.org/guide/essentials/dynamic-matching.html#Reacting-to-Params-Changes)

webapp/src/components/training/Training.vue
```diff
@@ -38,22 +38,42 @@ import Finished from '@/components/training/Finished.vue'
...
+  if (route.fullPath !== '/list' && (trainingStore.step == 0 || route.params.id !== props.id)) {
```
tharvik: full link look like too much information, and harder to keep track of (don't forget to update the watch target)

```suggestion
  if (route.name !== 'task-list' && (trainingStore.step == 0 || route.params.id !== props.id)) {
```

also, nitpicking on code style: I do like to have flatter functions' body, where `if` are there to "remove obstacles from the happy path". so, if most of the body of a function is wrapped inside an `if`, I would refactor it to rather return with the inverse condition. (arf, hard to explain, here goes some code)
```ts
function f() {
  if (cond) {
    …
  }
}
```
into
```ts
function f() {
  if (!cond) return
  …
}
```

so in our case here
```ts
() => {
  if (route.name === "task-list") return;
  if (trainingStore.step !== 0 && route.params.id === props.id) return;
  setupTrainingStore();
}
```
that reads a bit better IMO

webapp/src/components/training/Training.vue
tharvik: yeah agree, we have to rework all the store usages, most of it isn't needed at all and can simply be replaced by correctly forwarding state from one step to the other.

webapp/src/store/tasks.ts
```diff
@@ -16,8 +16,12 @@ export const useTasksStore = defineStore('tasks', () => {
   
   // used to display loading indicator in the webapp when loading tasks
   const loading = ref(false)
```
tharvik: haa, my typing demons are coming back to haunt me!

consider using a more fun type, like `const loading = ref<'successed' | 'failed' | false>`.
no need to play with many variables, and it ensure state consistency.

tharvik (approved):
well done, that's way more stressful for users now! :closed_lock_with_key:
only some cosmetic comments as usual

webapp/src/components/data/dataset_input/DatasetInput.vue
JulienVig: Good catch! it was supposed to be used twice! It got lost in some change reversion.

You were saying your limit is 3, mine is at 2 in this case. I believe that someone other than me would not necessarily notice it needs to be changed in 2 different places, especially if it's a quick change to reword or fix a typo.

webapp/src/components/training/Training.vue
JulienVig: They are different objects!
From the [vue-router doc](https://router.vuejs.org/guide/): "The key thing to notice is that the composables useRouter() and useRoute() are used to access the router instance and current route respectively."

webapp/src/components/training/Training.vue
JulienVig: I was actually using `route.name` but it's sometimes undefined, while `route.fullPath` is always defined. Which means I can be on the list page and route.name is sometimes 'task-list' and sometimes 'undefined' which is just calling for future issues.
I don't have a strong opinion on the function flatness, I'm in lining the method!

webapp/src/components/training/Training.vue
JulienVig: Well, I just tried it and I can't get onBeforeRouteUpdate to trigger so I'll leave it as it is. There are also some technical differences between the two that I don't have time to test.

webapp/src/components/training/Training.vue
JulienVig: Even seems that using route.fullPath triggers on mount and let's me remove the onMount hook

webapp/src/store/tasks.ts
JulienVig: Very neat! I don't like that  `if(loading) console.log(loading)` would print 'success' or 'failed'. This is unexpected since loading is true (it should be loading)
I'm changing it to `const status = ref<'success' | 'failed' | 'loading'>` to make everything explicit

PR #734, opened 2024-07-30
add LiGHT branding

tharvik:
only MLO was on the webapp (thanks @Annie-LiGHT for pointing it out).

* rewrote the About Us page
* change the footer to add LiGHT
* remove some dead code & simplify some other
  * flatten `IconCard`'s templates (avoid wrapping into content specific)
  * extract `IconCardHeader` for use in `DropdownCard`
  * rm dead `TitleCard`

notes:
 * `Card` is very sparsly used (only once) but look like is should be the parent of the others cards related component. we should refactor this.
 * links are restylied in various places; from [tailwind's doc](https://tailwindcss.com/docs/reusing-styles#avoiding-premature-abstraction), we should probably create a component to avoid that

Footer
image

About Us page (oc LiGHT logo is moving :sparkles:)
image


martinjaggi (approved):

JulienVig:
What do you think of something like this for the About Us page? (logos are hyperlink)

Screenshot 2024-08-05 at 13 18 58

tharvik:
> What do you think of something like this for the About Us page?

wow, the full name of Annie's lab is huuge!
go, look good to me, except that "the disco team" makes me expect a list of people rather than a list of labs (now that I think of it: it should be also kept in sync with onhover of the footer and the filename)

JulienVig:
> "the disco team" makes me expect a list of people rather than a list of labs

Yea I agree, but I think it's a small inconsistency I'm willing to pay to have to keep track of members as they come and go

JulienVig (approved):
All good thanks for the old code cleanups!

PR #735, opened 2024-07-31
scraps from #682

### LINKED ISSUE #201: console.log only when in dev mode
Using console.log can be very slow (similar to how print() in python or other languages console output can create a big overhead). This is important only in the training_manager as there are a lot of console.log. We should do something like:

if (dev mode)
  console.log('some info')

[https://stackoverflow.com/questions/11426185/will-console-log-reduce-javascript-execution-performance](https://stackoverflow.com/questions/11426185/will-console-log-reduce-javascript-execution-performance)

tharvik:
#682 was postponed until the v3 release, so much that it's very cumbersome to rebase it on the current code. this PR contains some scraps of this PR, commits which were a bit out of scope, and some code massaging to make it fit nicer. consider it a split.

* rework `DatasetInput` into specialized components
  * simplify `FileSelection`
    * only takes care of entering files and make it nice, no parsing (handled by parents)
    * changed to be a [v-model](https://vuejs.org/guide/components/v-model.html) instead of emitting events (better represent that it keeps data ~inside it)
      * we can invalidate it from the outside by setting its model prop to undef
  * split into `LabeledImageDatasetInput` (with wrapped `ByCSV` & `ByGroup`), `ImageDatasetInput` (without labels), `TabularDatasetInput` & `TextInput`
    * for now, no much special handling of passed `DatasetBuilder` but will change when adding types to represented datasets in next PR
  * inline `Data` at call sites, as labelled datasets (for training) and unlabelled ones (for evaluation) are quite different  
  * rename "webapp/src/components/data" to "…/dataset_input"
  * with tests™©®
* rework how to start the server
  * single class `Server` that get's constructed with the wanted tasks: no more name clash with `discojs.Disco`, clearer `Task` setup, no more mutability
  * single CLI entrypoint as `main.ts` which also print some infos for humans
    * no more garbage output in tests
  * inline config: single callsite, most fields unused
* webapp changes
  * avoid unneeded dynamic components
  * simplify `ImageCard`
* fix "missing tippy directive" warning in tests
* add dependabots tracking for docs/examples
* rewrite `console.*` calls to use [debug](https://www.npmjs.com/package/debug), a widely used logging library in JS
* update some types here and there


cli/src/cli.ts
```diff
@@ -31,10 +31,13 @@ async function runUser(
...
 
-async function main (task: Task, numberOfUsers: number): Promise<void> {
+async function main (provider: TaskProvider, numberOfUsers: number): Promise<void> {
+  const task = provider.getTask()
   console.log(`Started ${task.trainingInformation.scheme} training of ${task.id}`)
   console.log({ args })
-  const [server, url] = await startServer()
+
+  const disco = await Server.of(provider)
+  const [server, url] = await disco.serve()
```
JulienVig:
```suggestion
  const discoServer = await Server.of(provider)
  const [server, url] = await discoServer.serve()
```
What about renaming the variable names to not use the same one as in `runUser` (where `const disco` is a `Disco` object, not `Server`)

discojs-web/src/memory/memory.ts
```diff
@@ -9,11 +9,11 @@
 import { Map } from 'immutable'
 import * as tf from '@tensorflow/tfjs'
 
-import type { Path, Model, ModelInfo, ModelSource } from '@epfml/discojs'
+import type { Model, ModelInfo, ModelSource } from '@epfml/discojs'
 import { Memory, models } from '@epfml/discojs'
 
 export class IndexedDB extends Memory {
-  override getModelMemoryPath (source: ModelSource): Path {
+  override getModelMemoryPath (source: ModelSource): string {
```
JulienVig, line 16: Why change Path to string?

discojs/src/aggregator/mean.ts
```diff
@@ -60,7 +63,8 @@ export class MeanAggregator extends Aggregator<WeightsContainer> {
     else {
       // Print a warning regarding the default behavior when thresholdType is not specified
       if (thresholdType === undefined) {
-        console.warn(
+	// TODO enforce validity by splitting features instead of warning
```
JulienVig, line 66: I'm not sure I understand, do you mean to prevent leaving thresholdType unspecified?

discojs-web/src/memory/memory.ts
JulienVig: Ah I see it's to inline the type

docs/examples/training.ts
```diff
@@ -25,7 +25,8 @@ async function main (): Promise<void> {
   const NAME: string = 'titanic'
 
   // Launch a server instance
-  const [server, url] = await startServer()
+  const disco = await Server.of(defaultTasks.simpleFace, defaultTasks.titanic)
+  const [server, url] = await disco.serve()
```
JulienVig:
```suggestion
  const discoServer = await Server.of(defaultTasks.simpleFace, defaultTasks.titanic)
  const [server, url] = await discoServer.serve()
```
Same comment about the naming conflict with `runUser`

webapp/src/components/dataset_input/FileSelection.vue
```diff
@@ -0,0 +1,171 @@
...
+  defineProps<{
+    type: "tabular" | "image" | "text";
+    multiple?: boolean; // accept one or multiple files
+  }>(),
+  {
+    multiple: false,
+  },
+);
+
+const model = defineModel<Set<File>>();
```
JulienVig: I think using the name `model` can be a bit confusing with an ML model. I understand it references the vus.js v-model but I don't think it is necessary to include it in the name. Something like `files` would make more sense to me.

webapp/src/components/dataset_input/FileSelection.vue
```diff
@@ -0,0 +1,171 @@
...
+  return `${name}${props.multiple ? "s" : ""}`;
+});
+const acceptFilter = computed(() => {
+  switch (props.type) {
+    case "tabular":
+      return ".csv";
+    case "image":
+      return "image/*";
+    case "text":
+      return "text/plain";
```
JulienVig, line 133: It seems I can upload an image or a csv in a text field without any error (and vice-versa), is it expected?

cli/src/cli.ts
tharvik: good idea, that's more explicit

discojs-web/src/memory/memory.ts
tharvik: it's called `Path` but isn't really, it is used an ID for the models rather than a filesystem path. I found it confusing (which is kinda against the idea of having a dedicated type).

discojs/src/aggregator/mean.ts
tharvik: yeah, kinda, it's more general than that: this aggregator is trying to do way too much:
* do a mean of received weights
* allow weights from previous rounds
  * from a fixed number of peers
  * from a relative percentage of peers

splitting it in various type, one per feature allow us to get rid of `thresholdType` by having its meanings "embedded into the type" (for eg, splitting it in `MeanAggregator` (only doing mean), `AbsoluteContributers` & `RelativeContributors`). chaining such aggregator (and related) makes it safer (warning are usually not read, except in the CLI maybe).

webapp/src/components/dataset_input/FileSelection.vue
tharvik: hum, on firefox ESR, when it opens the file upload modal, it filters the shown dir with "Text File" (or "Image Files" , …). one can choose to "show all files" and in that case, select anything.
so yeah, you can upload anything but you'll have to actively do it wrong with my setup. is it not the case in your browser?

if we want to check that the files are indeed valid, we need to read the whole dataset to know that (which is not currently done with `datasetBuilder`, only partially when building it). I can add such check in #682 but yeah, garbage-in garbarge-out.

webapp/src/components/dataset_input/LabeledImageDatasetInput/__tests__/ByGroup.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+    props: {
+      labels: Set.of("file"),
+      datasetBuilder,
+    },
+    global: { directives: { Tippy } },
+  });
+
+  await wrapper
+    .getComponent(FileSelection)
+    .setValue(Set.of(new File([], "file.csv")));
```
JulienVig:
```suggestion
    .setValue(Set.of(new File([], "image.png")));
```
If I'm not mistaken file selection by group is only available for images

webapp/src/components/dataset_input/LabeledImageDatasetInput/__tests__/ByCSV.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+  await flushPromises(); // handler is async
+
+  await folderSelector.setValue(
+    Set.of(
+      new File([], "first.csv"),
+      new File([], "second.csv"),
+      new File([], "third.csv"),
+      new File([], "unrelated.csv"),
+    ),
+  );
```
JulienVig:
```suggestion
  await folderSelector.setValue(
    Set.of(
      new File([], "first.png"),
      new File([], "second.png"),
      new File([], "third.png"),
      new File([], "unrelated.png"),
    ),
  );
```
The second field is for images

JulienVig: Also we are allowing connecting more images that specified in the CSV, should be add a warning about this? This scenario would probably not be intentional so a notice could be useful

webapp/src/components/dataset_input/__tests__/TabularDatasetInput.spec.ts
```diff
@@ -0,0 +1,37 @@
...
+import { directive as Tippy } from "vue-tippy";
+import { flushPromises, mount } from "@vue/test-utils";
+
+import { data, defaultTasks } from "@epfml/discojs";
+import { WebTabularLoader } from "@epfml/discojs-web";
+
+import TabularDatasetInput from "../TabularDatasetInput.vue";
+import FileSelection from "../FileSelection.vue";
+
+it("adds to builder when inputing a file", async () => {
```
JulienVig:
```suggestion
it("adds to builder when inputting a file", async () => {
```

JulienVig (approved):
Impressive work as always!! I really like the Server refactor, the name conflict has been bothering me for a while. And the `debug` library is really cool.
A few comments here and there but nothing blocking!

## FileSelection
* Connecting by csv and not plugging all the images listed in the csv triggers the usual "some images listed in the CSV are missing" toaster but now there is an additional "Something went wrong on our side" that appears.
* Extra feature: Now that we are using v-model can we clear already connected images  when clearing the csv file? O.w. this allows bypassing the file check: 1) connect a CSV file 2) connect corresponding images 3) clear CSV file (the image field is now blocked but the images are not cleared) 4) connect a new CSV that has nothing to do with the current connected images -> no error while there should be
* right now users can connect more images than specified in the csv files, should be allow that? Or add a warning when this happens?

webapp/src/components/dataset_input/FileSelection.vue
JulienVig: Oh I was drag and dropping files so there was no modal to prevent me from connecting wrong things

webapp/src/components/dataset_input/LabeledImageDatasetInput/__tests__/ByCSV.spec.ts
tharvik: good idea, added a toaster warning. also added an error if the CSV is matching the same filename to multiple labels.

tharvik:
> * Connecting by csv and not plugging all the images listed in the csv triggers the usual "some images listed in the CSV are missing" toaster but now there is an additional "Something went wrong on our side" that appears.

right, I forgot some `throw`s without try-catch wraps. should be fixed now.

> * Extra feature: Now that we are using v-model can we clear already connected images  when clearing the csv file?

good catch! yes, added (soo satisfying to add a feature with a one liner).

> * right now users can connect more images than specified in the csv files, should be allow that? Or add a warning when this happens?

toaster warning it is!

Issue #718, opened 2024-07-23
New users are not able to join once training started


JulienVig:
Currently, participants need to join within a few seconds of each other, otherwise their contribution is dropped out (by the server in federated and by other peers in decentralized).

### Federated
- [x] The first thing to do is to make the server return the current communication round so that peer can get up to date with the network. This is currently not done so users that join a bit too late keep trailing behind and sending outdated updates.
- [x] If the server receives an outdated contribution, it should send back the latest model  so that the user can catch up.

### Decentralized
- [x] A peer joining the network after the training started doesn't update its round to the current round so its contributions are always dropped because outdated.
- [x] As soon as `minReadyPeer` peers joined a task, the collaborative training starts and sometimes starts so quickly that there isn't enough time for other peers to join and contribute before the first round is already finished. For example if `minReadyPeer = 3`, as soon as 3 peers joined the network, they start aggregating their contributions. Even if a 4th peer joins right after the 3rd, its contribution may be dropped because the first 3 peers already passed to a new round. This should be fixed by the previous checkbox, enabling outdated peers to catch up to the latest round but the first round may still finish before every peer that wanted to join could contribute. 
The solution implemented is described in #775

PR #741, opened 2024-08-06
Let users join federated training mid session

JulienVig:
Fixes the federated case of #718
- [x] Have server store the last aggregated model and send it back to participants joining late, or that have fallen behind
- [x] Participants notify server when disconnecting from a task
- [x] Implement a "waiting room" to have at least 2 participants before starting a federated training
- [x] Handle the case when a user ends up alone during in a training session (all the others either stopped or already finished). This user could pause their training and wait until someone else joins
- [x] Handle receiving an invalid server update. Currently, a client waits for a single server update per round. If it receives an old or invalid server updates, the client drops it and skips to the next round. However, the update may simply be delayed. Should the client skip to the next round or wait for another potential update?
- [x] Display the current status in the trainingboard, especially when waiting for new participants.
Screenshot 2024-08-14 at 17 30 34

Refactoring:
* Split `server/router` into `routes` and `controllers` [as commonly done in Express](https://developer.mozilla.org/en-US/docs/Learn/Server-side/Express_Nodejs/routes). `routes` setup the router API and `controllers` handle the actual federated and decentralized logic (via `initTask` and `handle`)
* Renamed some files into more explicit names imo (e.g. `client/client.ts`, `client/federated_client.ts`, `client/decentralized_client.ts`)
* Renamed 'feai' and 'deai' occurrences (e.g. server API) into 'federated' and 'decentralized'
* Renamed the TasksAndModels object to TaskInitializer and added doc


martinjaggi:
BTW for delayed updates during training, the easiest way (and algorithmically valid) is to define a maximum threshold delay, so a time delta (in number of steps/rounds): if an update arrives which was computed based on a model older than this delta, the update is dropped. otherwise it's included. 

the aggregation is simply triggered once the number of present valid update candidates (buffer size) is sufficient (this part is already implemented). 

that would work well both in decentralized and federated

JulienVig:
Yes! This should already be implemented, although current defaults always set the time delta to 0. We can probably increase the cutoff to 1 or 2.

discojs/src/aggregator/base.ts
```diff
@@ -134,6 +154,16 @@ export abstract class Base<T> extends EventEmitter<{'aggregation': T }> {
...
   }
 
+  /**
+   * Remove a node's id from the set of active nodes.
+   * @param nodeId The node to be removed
+   */
+  removeNode (nodeId: client.NodeID): void{
+    if (this.nodes.has(nodeId)) {
+      this._nodes = this._nodes.delete(nodeId)
+    }
```
tharvik: `delete` doesn't fail if the value is not present (and it avoid double lookup)

```suggestion
    this._nodes = this._nodes.delete(nodeId)
```

discojs/src/aggregator/base.ts
```diff
@@ -69,14 +69,34 @@ export abstract class Base<T> extends EventEmitter<{'aggregation': T }> {
...
    */
   abstract add (nodeId: client.NodeID, contribution: T, round: number, communicationRound?: number): boolean
 
+  /**
+   * Evaluates whether a given participant contribution can be used in the current aggregation round
+   * the boolean returned by `this.add` is obtained via `this.isValidContribution`
+   */
+  isValidContribution(nodeId: client.NodeID, round: number): boolean {
+    if (!this.nodes.has(nodeId)) {
+      console.warn("Contribution rejected because node id is not registered")
```
tharvik: we can use `debug` now instead :)

```suggestion
      debug("Contribution rejected because node id is not registered")
```

discojs/src/aggregator/base.ts
```diff
@@ -69,14 +69,34 @@ export abstract class Base<T> extends EventEmitter<{'aggregation': T }> {
...
+      return false;
+    }
+    if (!this.isWithinRoundCutoff(round)) {
+      console.warn(`Contribution rejected because round ${round} is not within round cutoff`)
```
tharvik:
```suggestion
      debug(`Contribution rejected because round ${round} is not within round cutoff`)
```

discojs/src/client/event_connection.ts
```diff
@@ -127,10 +127,10 @@ export class WebSocketServer extends EventEmitter<{ [K in type]: NarrowMessage<K
     })
   }
 
-  disconnect (): Promise<void> {
+  disconnect(): Promise<void> {
     return new Promise((resolve, reject) => {
-      this.socket.once('close', resolve)
-      this.socket.once('error', reject)
+      this.socket.onclose = () => resolve()
+      this.socket.onerror = () => reject()
```
tharvik: let's add context to errors

```suggestion
      this.socket.onerror = (e) => reject(e.message)
```

discojs/src/client/federated/federated_client.ts
```diff
@@ -0,0 +1,247 @@
...
+import {
+  serialization,
+  type WeightsContainer,
+} from "../../index.js";
+import { Client } from "../client.js";
+import { type, type ClientConnected, type ClientDisconnected } from "../messages.js";
+import {
+  type EventConnection,
+  waitMessageWithTimeout,
+  // waitMessage,
```
tharvik:
```suggestion
```

discojs/src/client/federated/federated_client.ts
```diff
@@ -0,0 +1,247 @@
...
+ * Client class that communicates with a centralized, federated server, when training
+ * a specific task in the federated setting.
+ */
+export class FederatedClient extends Client {
+  /**
+   * Arbitrary node id assigned to the federated server which we are communicating with.
+   * Indeed, the server acts as a node within the network. In the federated setting described
+   * by this client class, the server is the only node which we are communicating with.
+   */
+  public static readonly SERVER_NODE_ID = "federated-server-node-id";
```
tharvik: it's not used outside, we can have it as a file-local `const` (not in the class I mean).

discojs/src/client/federated/federated_client.ts
```diff
@@ -0,0 +1,247 @@
...
+  
+  // Total number of other federated contributors, including this client, excluding the server
+  // E.g., if 3 users are training a federated model, nbOfParticipants is 3
+  #nbOfParticipants: number = 1;
+  
+  /**
+   * Flag set by the server to tell clients to pause and wait until more
+   * participants join the session
+   */
+  #waitingForMoreParticipants = false;
```
tharvik: this seem to be closely tight to the promise. consider avoiding keeping state in sync by deriving this one from the promise, ie `get #waitingForMoreParticipants(): boolean { this.#promiseForMoreParticipants !== undefined }`

discojs/src/client/federated/messages.ts
```diff
@@ -1,11 +1,17 @@
...
 
-import { type, hasMessageType, type AssignNodeID, type ClientConnected } from '../messages.js'
+import {
+  type, hasMessageType, type AssignNodeID, type ClientConnected, type ClientDisconnected,
+ } from '../messages.js'
 
+ // See ../messages.ts for doc
 export type MessageFederated =
   ClientConnected |
+  ClientDisconnected |
```
tharvik: can we avoid `ClientDisconnected` altogether? the server knows when one disconnect with `WebSocket.once('close', …)` (same idea for `ClientConnected` with `WebSocket.once('open', …)`)

the less messages types, the less possible state has to be handled. for example, what happen if the client closes the connection but never send a `ClientDisconnected` message (such as when it crashes).

server/src/controllers/federated_controller.ts
```diff
@@ -0,0 +1,246 @@
...
+import { TrainingController } from "./training_controller.js";
+
+import Messages = client.messages
+import MessageTypes = Messages.type
+import FederatedMessages = client.federated.messages
+
+const debug = createDebug("server:controllers:federated")
+
+// Arbitrarily set a minimum of two participants in a federated training
+const MIN_PARTICIPANTS = 2
```
tharvik: ~~what's the difference with `Task.trainingInformation.minimumReadyPeers`?~~

can we expand the meaning of `Task.trainingInformation.minimumReadyPeers` to also be for the federated case?

discojs/src/client/federated/federated_client.ts
```diff
@@ -0,0 +1,247 @@
...
+
+    // Need to await the resulting global model right after sending our local contribution
+    // to make sure we don't miss it
+    this.server.send(msg);
+    const serverResult = await this.receiveServerResult();
+
+    if (
+      serverResult !== undefined &&
+      this.aggregator.add(FederatedClient.SERVER_NODE_ID, serverResult, this.aggregator.round)
+    ) {
```
tharvik: I'm drifting a bit out of scope: why do we even need a local aggregator when doing federated? shouldn't we simply trust what the server is telling us and not use our own aggregator? that look to me more like classical federated and under a honest-but-curious server threat model, safe.

discojs/src/client/federated/federated_client.ts
```diff
@@ -0,0 +1,247 @@
...
+    }
+    return await this.aggregationResult
+  }
+
+  /**
+   * Wait for the server to reply with the most recent aggregated weights
+   * after having sent out local weight update to the federated server
+   */
+  private async receiveServerResult(): Promise<WeightsContainer | undefined> {
+    // Try receiving the latest update multiple times in case we receive outdated updates
```
tharvik: can we receive outdated updates? I mean, the server knows best what is the current round as it is the one orchestrating the process. trying in a loop looks weird to me. if the server is sending us some unexpected results, it's that we're out of sync and should react to that rather than trying again.

discojs/src/client/federated/federated_client.ts
```diff
@@ -0,0 +1,247 @@
...
+    const serverResult = await this.receiveServerResult();
+
+    if (
+      serverResult !== undefined &&
+      this.aggregator.add(FederatedClient.SERVER_NODE_ID, serverResult, this.aggregator.round)
+    ) {
+      // Regular case: the server sends us its aggregation result which will serve our
+      // own aggregation result.
+    } else {
+      // Unexpected case: for some reason, the server result is stale.
```
tharvik: yeah well, if the server is not answering, then there isn't much we can do: if we continue training, we won't be able to get a network model but only a local one, which is unexpected from federated learning. maybe we should just fail?

discojs/src/client/federated/federated_client.ts
```diff
@@ -0,0 +1,247 @@
...
+      type: type.ClientDisconnected,
+    };
+    this.server.send(msg); // notify the server we are disconnecting but don't wait for an answer
+    await this.server.disconnect();
+    this._server = undefined;
+    this._ownId = undefined;
+
+    this.aggregator.setNodes(this.aggregator.nodes.delete(FederatedClient.SERVER_NODE_ID));
+
+    return Promise.resolve();
```
tharvik: it was added when functions were not `async` (eslint was telling me that the function didn't needed to be async, but to respect the return type, I had to return a `Promise`)

```suggestion
```

server/src/controllers/federated_controller.ts
```diff
@@ -0,0 +1,246 @@
...
+      this.#participants.delete(clientId)
+      this.#aggregator.removeNode(clientId)
+      debug("client leaving: %o", clientId)
+
+      // Check if we dropped below the minimum number of participant required 
+      if (this.#participants.size >= MIN_PARTICIPANTS) return
+
+      this.#waitingForMoreParticipants = true
+      // Tell remaining participants to wait until more participants join
+      for (const [participantId, participantWs] of this.#participants.entries()) {
```
tharvik: if two clients disconnect, we will send the `WaitingForMoreParticipants` message twice to the remaining ones (not an issue with `MIN_PARTICIPANTS ==  2`)

server/src/routes/training.ts
```diff
@@ -59,18 +82,19 @@ export abstract class Server {
...
-
-  protected abstract readonly description: string
-
-  protected abstract buildRoute (task: TaskID): string
+  protected buildRoute (task: TaskID): string {
+    return `/${task}`
+  }
 
-  protected abstract initTask (task: TaskID, model: Model): void
+  public isValidUrl (url: string | undefined): boolean {
```
tharvik: while at it, can you inline the other `is*` methods? all are used at most once

server/src/routes/training.ts
```diff
@@ -1,23 +1,33 @@
...
+ */
+export class TrainingRouter {
   private readonly ownRouter: expressWS.Router
 
-  private readonly tasks: string[] = new Array<string>()
+  private readonly tasks = new Set<string>()
   private readonly UUIDRegexExp = /^[0-9a-fA-F]{8}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{12}$/gi
 
-  constructor (wsApplier: expressWS.Instance, tasksAndModels: TasksAndModels) {
+  protected readonly description: string
```
tharvik: can you also inline it? it's only used in the constructor

discojs/src/training/trainer.ts
```diff
@@ -33,11 +35,13 @@ export class Trainer {
     task: Task,
     public readonly model: Model,
     client: Client,
+    logger: Logger
```
tharvik: can we keep rather keep it in `Disco`? I wanted the `Trainer` to be more library oriented (not involved with user interface) and `Disco` to be more user-focused. that way, we can more easily reuse `Trainer` in other libraries by reducing coupling (no `Memory` here either).

discojs/src/logging/mock_logger.ts
```diff
@@ -0,0 +1,12 @@
+import { Logger, TrainingStatus } from "./logger.js";
+
+/**
+ * MockLogger not doing anything
+ *
+ * @class Logger
+ */
+export class MockLogger implements Logger {
```
tharvik: "Mock" means more to ~simulate what an object would do will keeping track of what happened. one would expected it to return the previously passed message and so on.
would `NullLogger` or `NopLogger` or `SinkLogger` be more fit?

discojs/src/logging/logger.ts
```diff
@@ -9,4 +14,10 @@ export interface Logger {
    * @param message - message to be displayed
    */
   error(message: string): void;
+  /**
+   * Displays the current training status
+   * @param status training status to display
+   * e.g., 'Waiting for more participants', 'Training locally', 'Updating the model'
+   */
+  setStatus(status: TrainingStatus): void;
```
tharvik: ho, I see what you want to do with that. sorry I didn't came up with a nice solution at our last meeting but now I think I've got smth good, tell me what you think:
let's implement `abstract class Client extends EventEmitter<{'round status': 'waiting' | 'training' | 'updating' }>`, and somethink similar on `Disco`. this way, one can simply `Disco.on('round status', (status) => …)` to do wathever (such as updating the DOM with some nice messages, or printing to the logger, or both).

webapp/src/components/containers/IconCard.vue
```diff
@@ -18,5 +21,6 @@ import IconCardHeader from "./IconCardHeader.vue";
 
 defineProps<{
   titlePlacement?: "left" | "center" | "right";
+  fillSpace?: boolean;
```
tharvik: it is treated as `false` in code, better to use a default instead of a tri-state

and I think I'm in my head there, let me explain: `fillSpace` here can have one of three values: `true`, `false` or `undefined` but the two last are in fact used in the same way. better to reduce the type to reflect that and only have two possible values: `true` or `false`. this way, every possible value as a different meaning, making it easier to reason about.

server/src/controllers/decentralized_controller.ts
```diff
@@ -4,50 +4,31 @@ import msgpack from 'msgpack-lite'
...
 
-import { Server } from '../server.js'
+import { TrainingController } from './training_controller.js'
 
 import messages = client.decentralized.messages
 import AssignNodeID = client.messages.AssignNodeID
 import MessageTypes = client.messages.type
 
-const debug = createDebug("server:router:decentralized")
+const debug = createDebug("server:routes:decentralized")
```
tharvik:
```suggestion
const debug = createDebug("server:controllers:decentralized")
```

server/src/controllers/decentralized_controller.ts
```diff
@@ -70,28 +51,20 @@ export class Decentralized extends Server {
             // Answer with client id in an AssignNodeID message
             const msg: AssignNodeID = {
               type: MessageTypes.AssignNodeID,
-              id: peerId
+              id: peerId,
+              waitForMoreParticipants: this.readyNodes.size  < minimumReadyPeers
```
tharvik: it's unused in the decentralized client, is it wanted? if yes, consider splitting into two different messages, if no, I don't think that the client is actually waiting.

server/src/controllers/federated_controller.ts
```diff
@@ -0,0 +1,246 @@
...
+  /**
+   * Promise containing the current round's results. To be awaited on when providing clients
+   * with the most recent result.
+   */
+  #result: Promise<WeightsContainer> | undefined = undefined
+  /**
+   * The last global model aggregated. The model is already serialized and 
+   * can be sent to participants joining mid-training or contributing to previous rounds
+   */
+  #latestGlobalModel: serialization.weights.Encoded | undefined = undefined
```
tharvik: "Model" makes me think that it also contains the architecture.

```suggestion
  #latestGlobalWeights: serialization.weights.Encoded | undefined = undefined
```

server/src/controllers/federated_controller.ts
```diff
@@ -0,0 +1,246 @@
...
+   * It registers what the server will do upon receiving messages from the participant.
+   * Note that `this.handle` is only called once to setup the logic. It is `ws.on()`
+   * that is called upon receiving messages (and not `this.handle`)
+   * 
+   * @param task the task associated with the current websocket (= participant)
+   * @param ws the websocket connection through which the participant and the server communicate
+   */
+  handle(ws: WebSocket): void {
+    if (this.#aggregator === undefined)
+      throw new Error(`no aggregator for task ${this.task.id}`)
```
tharvik: this can't happen (there is an eslint rule for that which I really want to enable but I'm waiting for upstream to update the stack)

```suggestion
```

server/src/controllers/federated_controller.ts
```diff
@@ -0,0 +1,246 @@
...
+          waitForMoreParticipants
+        }
+        ws.send(msgpack.encode(msg))
+
+        debug("Wait for more participant flag: %o", waitForMoreParticipants)
+        
+        // If we were previously waiting for more participants to join and we now have enough,
+        // broadcast to previously waiting participants that the training can start
+        if (this.#waitingForMoreParticipants && !waitForMoreParticipants) {
+          for (const [participantId, participantWs] of this.#participants.entries()) {
```
tharvik: this can be a filter & forEach #nitpicking

server/src/controllers/federated_controller.ts
```diff
@@ -0,0 +1,246 @@
...
+      * A client sends a weight update to the server
+      */
+      } else if (msg.type === MessageTypes.SendPayload) {
+
+        const { payload, round } = msg
+        if (this.#aggregator.isValidContribution(clientId, round)) {
+          // We need to create a promise waiting for the global model before adding the contribution to the aggregator
+          // (so that the aggregation and sending the global model to participants
+          // doesn't happen before the promise is created)
+          this.createPromiseForWeights(ws)
```
tharvik: soo, I'll bother a bit on this, disregard it if you think that it's not the PR to do that. I might also be quite wrong as it's quite uncharted territories for me.

as this is the only place where the aggregator is being added to, we can avoid all this shenanigans with creating and storing `Promise`s. we can have `Aggregator.add` return the result of the aggregation (it works for both fixed and relative number of peers). this way, no need to wait for the result or playing around with Promises and a clearer way to know when the aggregation results are computed.

server/src/controllers/federated_controller.ts
```diff
@@ -0,0 +1,246 @@
...
+   * Boolean used to know if we have enough participants to train or if 
+   * we should be waiting for more
+   */
+  #waitingForMoreParticipants = true
+  /**
+   * List of active participants along with their websockets
+   * the list allows updating participants about the training status 
+   * i.e. waiting for more participants or resuming training
+   */
+  #participants = new Map<string, WebSocket>()
```
tharvik: to be consistent, please use the one from immutable, not the ECMA one. [immutability has great benefits](https://immutable-js.com/#the-case-for-immutability), especially in mutable-everywhere/script languages.
beware that with this change, `#participants.delete` doesn't modify the collection in place

server/src/controllers/federated_controller.ts
```diff
@@ -0,0 +1,246 @@
...
+    const serializedModel =  await serialization.weights.encode(globalModel)
+    this.#latestGlobalModel = serializedModel
+
+    // Create a new promise for the next round
+    // TODO weird usage, should be handled inside of aggregator
+    void this.storeAggregationResult()
+  }
+
+  initTask(): void {
+    // start the perpetual promise loop
```
tharvik: :cry:

server/src/controllers/federated_controller.ts
```diff
@@ -0,0 +1,246 @@
...
+        this.#waitingForMoreParticipants = waitForMoreParticipants // update the attribute
+
+      /* 
+      * A participant leaves the task 
+      */
+      } else if (msg.type === MessageTypes.ClientDisconnected) {
+        debug(`client ${clientId} left ${this.task.id}`)
+
+        this.#participants.delete(clientId)
+        this.#aggregator.removeNode(clientId)
```
tharvik: not the same logic as in `ws.once('close', …)`

server/src/controllers/training_controller.ts
```diff
@@ -0,0 +1,32 @@
...
+ * 
+ * More info on controllers:
+ * https://developer.mozilla.org/en-US/docs/Learn/Server-side/Express_Nodejs/routes
+ * 
+ */
+export abstract class TrainingController {
+
+  constructor(protected readonly task: Task) { }
+
+  abstract initTask (model: Model): void
```
tharvik: subclasses never uses the passed model. also, from what I saw in code, it's always called just after construction. consider remove it altogether

server/src/controllers/training_controller.ts
```diff
@@ -0,0 +1,32 @@
...
+
+  abstract handle(
+    ws: WebSocket,
+    model: Model,
+    req: express.Request,
```
tharvik: both unused in subclasses. in JS, if a function receives more argument than needed, it's not an issue (compared to stack based function calls in compiled languages)

```suggestion
```

server/src/server.ts
```diff
@@ -37,8 +50,27 @@ export class Server {
     app.use(express.json({ limit: "50mb" }));
     app.use(express.urlencoded({ limit: "50mb", extended: false }));
 
-    const baseRouter = new Router(wsApplier, this.#tasksAndModels);
-    app.use("/", baseRouter.router);
+    const taskRouter = new TaskRouter(this.#tasksAndModels)
+    const federatedRouter = new TrainingRouter('federated', wsApplier, this.#tasksAndModels)
+    const decentralizedRouter = new TrainingRouter('decentralized', wsApplier, this.#tasksAndModels)
+
+    process.nextTick(() =>
```
tharvik: do we need the next tick?

webapp/src/components/training/Trainer.vue
```diff
@@ -8,30 +8,70 @@
     </div>
     <!-- Train Button -->
     <div class="flex justify-center">
-      <IconCard title-placement="center">
+      <IconCard title-placement="center" :fill-space="true">
```
tharvik: one of the nice thing that you have with boolean props is that you don't even need to set it to true, its mere presence is enough.

```suggestion
      <IconCard title-placement="center" fill-space>
```

webapp/src/components/training/Trainer.vue
```diff
@@ -171,17 +217,23 @@ async function startTraining(distributed: boolean): Promise<void> {
...
+      setStatus: (status: TrainingStatus) => trainingStatus.value = status
     },
     memory: memoryStore.useIndexedDB ? new IndexedDB() : new EmptyMemory(),
-    scheme: distributed ? props.task.trainingInformation.scheme : "local",
+    scheme: isTrainingAlone.value ? "local": props.task.trainingInformation.scheme,
   });
 
+  // Store the cleanup function such that it can be ran if users
+  // manually interrupt the training
+  cleanupTrainingSessionFn.value = async () => {
```
tharvik: ho, that's for capturing `disco` innit. hum, tricky, a bit hard to comprehend. consider moving the other cleanups (`displayModelCaching`, `*training*`, …) to `cleanupTrainingSession` and update the name to smth like `cleanupDisco` or `cleanupServerConnection`.

webapp/src/components/training/Trainer.vue
```diff
@@ -219,14 +271,24 @@ async function startTraining(distributed: boolean): Promise<void> {
...
 
+/**
+ * 
+ */
+async function cleanupTrainingSession() {
+  // check if the function has been initialized
+  if (cleanupTrainingSessionFn.value === undefined) return
+  // Calling the cleanup function returns a promise
+  // awaiting the promise notifies the network that we are disconnecting
+  await cleanupTrainingSessionFn.value()
```
tharvik: that might fire multiple times (TTCTTU). consider moving the cleanup function locally and setting the outside one to undef before the `await` point so that no other ~threads can pass the `undef` check

tharvik (changes requested):
incredible! that's some massive work, well done! no need to synchronize at sub-seconds level now :tada: 

I'm might have gone a bit our of the road on the comments, oupsi; nothing too bothering, it touches some files I hardly changed so many are for my own understanding.
can you add also some tests for the disconnect and reconnect handling, see how well it advances under stress?

discojs/src/aggregator/base.ts
JulienVig: Ah yes that's left from when I couldn't manage to print debug statements in the browser

discojs/src/client/federated/federated_client.ts
JulienVig: Yes definitely, I've been fighting the urge to rework the client aggregation

server/src/controllers/training_controller.ts
JulienVig: Nice, I hesitated removing them but my editor was telling me you had added these lines so I didn't want to break something I didn't get

discojs/src/client/federated/federated_client.ts
JulienVig: It's to handle unreliable network events like message swaps or an old update arriving late. But I guess an unreliable network would break many other things in disco, I'm mostly worried about only waiting for a single message because if it's not the right one then the participant can't recover for this round.

discojs/src/client/federated/federated_client.ts
JulienVig: So far we were just timing out and continuing to the next round which I think makes sense if we expect the server to recover at some point. Maybe we could instead wait indefinitely until the server recovers (and display a status message in the UI) ? Failing would also make sense after waiting long enough

discojs/src/client/federated/federated_client.ts
JulienVig: I've implemented it in a loop but the idea simply is "wait for the expected update for a given amount of time and discard any unexpected message". But because we are manually "waiting" for server messages, "discarding unexpected messages" translates into a loop of manually waiting.

discojs/src/client/federated/federated_client.ts
JulienVig: Yes, just need to make sure to reset the promise back to undefined once it resolved

server/src/controllers/federated_controller.ts
JulienVig: Good catch, I'm adding a check to see if we were already waiting for more participants

server/src/controllers/decentralized_controller.ts
JulienVig: It's indeed unused in decentralized. I added this while doing the same for federated. This is just some "foundation" work for future work.

JulienVig: Would you prefer to remove this change altogether from the PR?

discojs/src/client/federated/federated_client.ts
JulienVig: Well in this case we should still check for the message "freshness" and discard any server updates from previous rounds

server/src/server.ts
JulienVig: No!

discojs/src/client/federated/federated_client.ts
tharvik: > I've been fighting the urge to rework the client aggregation

very hard to not succumb to the rewrite urges :smiling_imp:

> Well in this case we should still check for the message "freshness" and discard any server updates from previous rounds

~~hum, I'm not sure that I follow: how would the server send a message of the previous round if that's the responsability of the server to set at which round we're at? a client shouldn't advance if the server hasn't send a `ReceiveServerPayload` message for the current round.
it could advance by itself (and I think that I might do so on timeout) but I don't think that should be the case.~~

saw your reponse on the comment below, continuing there.

server/src/controllers/training_controller.ts
JulienVig: This made me realize that clients get their initial model from the server when initializing a Disco object (`Disco.fromTask` calls `client.getLatestModel`) and not when starting training (`disco.train`)
This explains why clients joining mid session still start with the initial model rather than the latest model from the session: `getLatestModel` returns the model the server has on disk but not latest model currently training

A solution would be to return the initial model when clients connect to the server within the AssignNodeID message which is used both in federated and decentralized (one less round trip compared to the current implementation in bonus)

discojs/src/client/federated/federated_client.ts
tharvik: > I'm mostly worried about only waiting for a single message because if it's not the right one then the participant can't recover for this round.

haa right, I see what you mean. that shouldn't really happen: websocket do not reorder messages (as it's over TCP). I guess as you added the loop, it got reorder anyway so that might be hinting that the message processing stack is missbehaving.

don't worry to much about it, if you think that's needed, you know best :)

discojs/src/client/federated/federated_client.ts
tharvik: yeah, waiting indefinitely (or a looong time) makes more sense to me. if the server crashes, I'm not even sure that it can restart a training. and the newer round will get rejected by the server anyway.

server/src/controllers/decentralized_controller.ts
tharvik: hum, as you wish; a comment/todo is welcome to let others know that it's baking.

server/src/controllers/training_controller.ts
tharvik: ho, nicely found! model origin is muddy but it's getting clearer with every PR!

> return the initial model when clients connect to the server

good idea, let's do that, that's both clearer and goes in the direction of dynamics tasks on the server :1st_place_medal:

server/src/controllers/training_controller.ts
tharvik: I don't even remember adding theses lines :upside_down_face:

discojs/src/client/federated/federated_client.ts
JulienVig: Aaah I was assuming we were working with UDP that makes things much easier great, then indeed then any message reorder should indicate something is wrong with the logic rather than the network

discojs/src/client/federated/federated_client.ts
JulienVig: Indeed with a reliable network and if we choose to wait indefinitely we don't need these checks, I'm simplifying all the logic

webapp/src/components/training/Trainer.vue
JulienVig: Good catch!

webapp/src/components/training/Trainer.vue
JulienVig: That's the fix I found to ensure we still called the cleanup even when the stop error was not caught, maybe it's not useful anymore with the changes in your PR

server/src/controllers/federated_controller.ts
JulienVig: As long as waiting on the `aggregator.add` doesn't block and prevent the server from receiving other messages and contributions, I think this would be a great solution! This PR is getting too big so I'll put this off to the next one

discojs/src/training/trainer.ts
JulienVig: The only way I found to circumvent this is to set the "training locally" status at the very end of the client's `onBeginRoundCommunication`. However, this relies on the trainer calling `client.onBeginRoundCommunication()` and `model.train()` back to back and requires duplicating the statement in federated client and decentralized client...

discojs/src/logging/logger.ts
JulienVig: Ah nice good idea!

webapp/src/components/training/Trainer.vue
JulienVig: I'm not sure if I misunderstood or if there is a typo but the other cleanups like displayModelCaching, isTraining etc are already included in cleanupTrainingSessionFn, is this what you meant?

server/src/controllers/training_controller.ts
JulienVig: I'm digging into this further and we also have to take into account the browser model caching, which currently allows users to start a training session with a cached model rather than the server's base model.
This raises the question: do we allow users to start training with their local models? (which implies allowing them to continue training previously trained models) Or always overwrite it with the server's base model?

I'm leaning towards always pulling the model from the server and overwriting the local working model (as it can be saved). I don't think being able to continue training a previously trained model is important

webapp/src/components/training/Trainer.vue
tharvik: not exactly, let me explain: `cleanupTrainingSessionFn` is doing more than it should. excepted `disco.close`, the other lines of the function are always doing the same thing. moving theses other lines in `function cleanupTrainingSession()` would make it a bit easier to read. also, with this change, `cleanupTrainingSessionFn` can be renamed to `cleanupDisco` or smth like that, which is a more precise name.

```ts
async function startTraining() {
  // …
  cleanupDisco.value = disco.close
  // …
}

async function cleanupTrainingSession() {
  const discoCloser = cleanupDisco.value
  if (discoCloser !== undefined)
    await discoCloser()

  cleanupDisco.value = undefined
  displayModelCaching.value = true
  trainingGenerator.value = undefined
  isTraining.value = false
}
```

server/src/controllers/training_controller.ts
tharvik: yeah, I'm clearly leaning in the same direction, don't really understand the need to train with a local model, let's overwrite it.

webapp/src/components/training/Trainer.vue
JulienVig: I see thanks!

JulienVig: Follow up: I had to make sure to write `cleanupDisco.value = async () => await disco.close()` rather than `cleanupDisco.value = disco.close` otherwise `this` was undefined when trying to access it within the `disco.close()` method

JulienVig:
Sorry for the ugly commit history I had to debug a test case that failed only in the github actions but passed locally.
FYI it was because of network delay letting a client send its contribution before they received the server message to wait for more participants (which should prevent them from sending their contribution). I've solved it by letting the server accept contributions even when below the minimum number of participants but preventing the aggregation until above the threshold.

.github/workflows/lint-test-build.yml
```diff
@@ -288,7 +288,7 @@ jobs:
           cache: npm
       - run: npm ci
       - run: npm --workspace={discojs,discojs-node} run build
-      - run: npm --workspace=server test
+      - run: DEBUG='server:controllers:federated|discojs:client:federated' npm --workspace=server test
```
tharvik: debug? or if we want to have a verbose CI, we should set it on every job

discojs/src/aggregator/mean.ts
```diff
@@ -122,6 +128,7 @@ export class MeanAggregator extends Aggregator<WeightsContainer> {
     this.log(AggregationStep.AGGREGATE);
 
     const result = aggregation.avg(currentContributions.values());
+    this._round = this._round++
```
tharvik: this is a double operation which is confusing

```suggestion
    this._round++
```

discojs/src/client/messages.ts
```diff
@@ -1,24 +1,40 @@
 import type * as decentralized from './decentralized/messages.js'
 import type * as federated from './federated/messages.js'
-import { type NodeID } from './types.js'
+// import { type NodeID } from './types.js'
```
tharvik:
```suggestion
```

server/src/task_initializer.ts
```diff
@@ -0,0 +1,190 @@
...
+  on(_: 'newTask', callback: (t: Task, m: EncodedModel) => Promise<void>): void {
+    // Apply the callback to already initialized task-model pairs
+    this.#initializedTasks.forEach(async ([t, m]) => { await callback(t, m) })
+    // Register the callback that will be ran when new tasks are added
+    this.listeners = this.listeners.push(callback)
+  }
+
+  // Emit a 'newTask' event, 
+  // It runs all the registered callbacks with the new task and model
+  emit(_: 'newTask', task: Task, model: EncodedModel): void {
```
tharvik: evetns can should only be emitted via `addTask`

```suggestion
  #emit(_: 'newTask', task: Task, model: EncodedModel): void {
```

server/src/task_initializer.ts
```diff
@@ -0,0 +1,190 @@
...
+ * from disk if they exists, downloading them from a URL or 
+ * initializing the model from its architecture definition. 
+ *  
+ * We work with EncodedModels rather than Models because they are sent encoded 
+ * to clients. Since the server doesn't need to use the Model, we
+ * simply leave it already encoded and ready to be sent to clients
+ * 
+ * Due to its asynchronous nature, TaskInitializer acts like an EventEmitter, 
+ * by registering callbacks on new tasks and emitting a 'newTask' event 
+ * when a new task has been added.
```
tharvik: why not directly using the EventEmitter from `discojs/utils`? I restricted the return type of the callbacks there to `void` but we can expand it to `void | PromiseLike<void>`.
the asynchronus nature of the callbacks are not really relevant to `TaskInitializer` itself, consumer are using it this way but it's not required for its correct use
note that it doesn't replay old events by itself

server/src/routes/task_router.ts
```diff
@@ -0,0 +1,102 @@
...
+import { serialization, isTask } from '@epfml/discojs'
+
+import type { TaskInitializer } from '../task_initializer.js'
+
+const debug = createDebug("server:router:task_router");
+
+export class TaskRouter {
+  readonly #expressRouter: express.Router
+
+  #tasksAndModels = Set<[Task, EncodedModel]>()
```
tharvik: same content as in `TaskInitializer` (after stabilisation) which might be kinda huge. consider exposing a field in `TaskInitializer` and referencing it instead.

server/src/task_initializer.ts
```diff
@@ -0,0 +1,190 @@
...
+        debug("removing model files at %s", modelPath)
+        await fs.rm(modelPath, { recursive: true, force: true })
+        throw e
+      }
+    }
+    
+    return model
+  }
+
+  // TODO: never used, seems that no task ever provide a digest
```
tharvik: dead feature I guess, we can remove it at a later point

server/tests/client/federated.spec.ts
```diff
@@ -28,7 +28,8 @@ describe("federated client", () => {
       TASK,
       aggregators.getAggregator(TASK),
     );
-    await client.connect();
+    const model = await client.connect();
+    expect(model).does.not.equal(undefined)
```
tharvik: that couldn't be the case, no? the return type is `Model` not `Model | undefined` so it should fail to compile

server/tests/client/federated.spec.ts
```diff
@@ -48,16 +49,19 @@ describe("federated client", () => {
...
       },
       aggregators.getAggregator(TASK),
     );
 
+    let model
     try {
-      await client.connect();
+      model = await client.connect();
     } catch {
+      expect(model).equal(undefined)
```
tharvik: undef is soo common, chai added [a special way to check it](https://www.chaijs.com/api/bdd/#method_undefined)

```suggestion
      expect(model).to.be.undefined
```

server/tests/e2e/federated.spec.ts
```diff
@@ -163,10 +163,153 @@ describe("end-to-end federated", () => {
...
+     * In this test the timeline is:
+     * - User 1 joins the task by themselves
+     * - User 2 joins
+     * - User 1 leaves
+     * - User 3 joins
+     * - User 2 & 3 leave
+     */
+    const TRAINING: RoundStatus = "Training the model on the data you connected"
+    const WAITING: RoundStatus = "Waiting for more participants"
+    const UPDATING: RoundStatus = "Updating the model with other participants' models"
```
tharvik, line 215: yeah, consider having `RoundStatus` the other way around: `type RoundStatus = "training" | "waiting" | "updating"` and put on the UI the burden of how to format it (humans need more text)

server/tests/e2e/federated.spec.ts
```diff
@@ -163,10 +163,153 @@ describe("end-to-end federated", () => {
...
+    const statusUpdateTime = 500
+
+    // Create User 1
+    const discoUser1 = new Disco(lusCovidTask, url, { scheme: "federated" });
+    let statusUser1: RoundStatus | undefined;
+    discoUser1.on("status", status => {
+      statusUser1 = status
+      console.log("user 1 status", status)
```
tharvik: no need to clutter test output

```suggestion
```

server/tests/e2e/federated.spec.ts
```diff
@@ -163,10 +163,153 @@ describe("end-to-end federated", () => {
...
+    })
+    const generatorUser1 = discoUser1.trainByRound(["image", dataset])
+    
+    // Have User 1 join the task and train locally for one round
+    const logUser1Round1 = await generatorUser1.next()
+    if (logUser1Round1.done) throw Error("User 1 finished training at their 1st round")
```
tharvik: makes for nicer error message

```suggestion
    expect(logUser1Round1.done).to.be.false
```

server/tests/e2e/federated.spec.ts
```diff
@@ -163,10 +163,153 @@ describe("end-to-end federated", () => {
...
+    // User 1 did a) and b) so their status should be Training
+    expect(logUser1Round1.value.participants).equal(1)
+    expect(statusUser1).equal(TRAINING)
+
+    // Calling next() a 2nd time makes User 1 go to c) where the client should
+    // stay stuck awaiting until another participant joins
+    const logUser1Round2Promise = generatorUser1.next()
+    await new Promise((res,_) => setTimeout(res, statusUpdateTime)) // Wait some time for the status to update
+    expect(statusUser1).equal("Waiting for more participants")
```
tharvik:
```suggestion
    expect(statusUser1).equal(WAITING)
```

server/tests/e2e/federated.spec.ts
```diff
@@ -163,10 +163,153 @@ describe("end-to-end federated", () => {
...
+
+    // Create User 2
+    const discoUser2 = new Disco(lusCovidTask, url, { scheme: "federated" });
+    let statusUser2: RoundStatus | undefined;
+    discoUser2.on("status", status => {
+      statusUser2 = status
+      console.log("user 2 status", status)
```
tharvik:
```suggestion
```

server/tests/e2e/federated.spec.ts
```diff
@@ -163,10 +163,153 @@ describe("end-to-end federated", () => {
...
+    })
+    const generatorUser2 = discoUser2.trainByRound(["image", dataset])
+
+    // Have User 2 join the task and train for one round
+    const logUser2Round1 = await generatorUser2.next()
+    if (logUser2Round1.done) throw Error("User 2 finished training at their 1st round")
```
tharvik:
```suggestion
    expect(logUser2Round1.done).to.be.false
```

server/tests/e2e/federated.spec.ts
```diff
@@ -163,10 +163,153 @@ describe("end-to-end federated", () => {
...
+    // and for the server to aggregate the local updates
+    expect(statusUser1).equal(UPDATING)
+
+    // Proceed with round 2
+    // the server should answer with the new global weights
+    // and users should train locally on the new weights
+    const logUser2Round2 = await generatorUser2.next()
+    const logUser1Round2 = await logUser1Round2Promise // the promise can resolve now
+    if (logUser2Round2.done || logUser1Round2.done)
+      throw Error("User 1 or 2 finished training at the 2nd round")
```
tharvik:
```suggestion
    expect(logUser1Round2.done).to.be.false
    expect(logUser2Round2.done).to.be.false
```

server/tests/e2e/federated.spec.ts
```diff
@@ -163,10 +163,153 @@ describe("end-to-end federated", () => {
...
+    const logUser2Round3Promise = generatorUser2.next()
+    await new Promise((res, _) => setTimeout(res, statusUpdateTime)) // Wait some time for the status to update
+    expect(statusUser2).equal(WAITING)
+
+    // Create User 3
+    const discoUser3 = new Disco(lusCovidTask, url, { scheme: "federated" });
+    let statusUser3: RoundStatus | undefined;
+    discoUser3.on("status", status => {
+      statusUser3 = status
+      console.log("user 3 status", status)
```
tharvik:
```suggestion
```

server/tests/e2e/federated.spec.ts
```diff
@@ -163,10 +163,153 @@ describe("end-to-end federated", () => {
...
+    })
+    const generatorUser3 = discoUser3.trainByRound(["image", dataset])
+
+    // User 3 joins mid-training and trains one local round
+    const logUser3Round1 = await generatorUser3.next()
+    if (logUser3Round1.done) throw Error("User 3 finished training at their 1st round")
```
tharvik:
```suggestion
    expect(logUser3Round1.done).to.be.false
```

tharvik (approved):
wouhou, great work, that's a very neat fix for an important feature. thanks for tests also, very useful to avoid breaking subtle status changes :tada: 

a few beautifulization comments, nothing really important

.github/workflows/lint-test-build.yml
JulienVig: debug indeed thanks

server/tests/client/federated.spec.ts
JulienVig: Indeed I was thinking of cases where something goes wrong with the connection but an error would be thrown elsewhere in those cases

server/src/task_initializer.ts
JulienVig: TasksAndModels didn't rely on `discojs/utils`'s EventEmitter so I didn't either I see if it's convenient
By asynchronous nature I meant the one of `addTask`, not the callbacks, does it make more sense? I can also just remove this paragraph if it's wrong/confusing

tharvik: no, no, the pharagraph helped me understand, I was just reusing it as a context for the comment. yeah, I should've also updated it when implementing `EventEmitter`, forgot about it that time

server/tests/e2e/federated.spec.ts
JulienVig: I agree! I'll push that to another PR though

server/src/task_initializer.ts
JulienVig: I didn't manage to adapt EventEmitter quickly so I'll defer that to another time!

martinjaggi:
impressive work, big thanks!

PR #764, opened 2024-08-28
rework memory

tharvik:
* mv memory from discojs* to webapp
  * it's only used there and was more bothering than useful.
* rework memory store into models store, with persistence in local storage
* drop ModelLibrary side panel, now the only way to use stored models is via predict/test
  * drop downloading of model, which was only working for TFJS models (not for GPT, nor any other type of model). we can probably reintroduce it's use when we have a standard way to represent weights ([HDF5]?)(https://en.wikipedia.org/wiki/Hierarchical_Data_Format). if people want to access that anyway, they can do so using CLI/examples
* also fix model serialisation when crossing node & browser boundaries
* also add icon into button card


martinjaggi:
thanks a lot, looking excellent! model store or model library indeed are much better names than memory here

webapp/src/components/testing/Testing.vue
```diff
@@ -30,39 +30,23 @@
                 "
                 class="shadow shadow-disco-cyan"
               >
-                <template #title>
-                  {{ taskTitle(metadata.taskID) }}
+                <template #title> {{ taskTitle(infos.taskID) }} </template>
+                <template #icon>
+                  <button @click="removeModel(id)"><Bin2Icon /></button>
```
JulienVig: Very nice! A small hover feedback would be great
```suggestion
                  <button 
                    @click="removeModel(id)"
                    class="hover:text-red-500 transition duration-200"
                  >
                    <Bin2Icon />
                  </button>
```

webapp/src/components/progress_bars/TestingBar.vue
```diff
@@ -61,7 +60,6 @@ import TestingButtons from "./TestingButtons.vue";
 
 const toaster = useToaster();
 const validationStore = useValidationStore();
-const memoryStore = useMemoryStore();
 
 const testing = computed(() => {
```
JulienVig: Related but not introduced by this PR: `testing` seems to be undefined when mounting the Testing page, so the TestingBar displays the title `Model ()`, it would be great to address the case when  `testing` is undefined more gracefully

JulienVig (changes requested):
Great job!! No more discojs memory!

* In the Testing page, clicking the next button displays an empty page. Maybe we should remove the next button to only allow going to the next page by clicking `Test` or `Predict`. (The title also displays `Model()`)
Screenshot 2024-08-29 at 16 59 32
* Training for a few epochs on Lus covid, stopping training and then saving the model display contradictory toasters. I guess we should be able to save the working model. Saving at the end of the training works great
Screenshot 2024-08-29 at 18 39 30

webapp/src/components/testing/Testing.vue
tharvik: hoo, that's so fancy, thanks for the tip!

JulienVig (approved):
All good thanks for your work!!

PR #765, opened 2024-08-29
Remove html injections and undoing model deletion

### LINKED ISSUE #258: Remove html injections
Currently in the tasks.json, the description is html code, so when we load it we are then injecting it into the vue template. It is bad practice to inject html code, especially since we will allow users to create tasks (and add information), this could be exploited.

The injection happens in line 24 in `TaskList.vue`:
```
 <span v-html="task.displayInformation.summary"></span>
```

Once this is done we can remove the corresponding error that is being muted in the `.eslintrc.json` (something about no-v-html).

### LINKED ISSUE #756: Implement undo feature when deleting a model
Thanks to #710 by @tomasoignons we now ask for confirmation when deleting a model from the model library:
Screenshot 2024-08-21 at 13 12 13

In my opinion, an improvement on this would be to let users undo their deletion by clicking on the toaster (e.g. `"The model has been deleted successfully. Click here to undo the deletion."`) rather than ask them to confirm systematically.

tomasoignons:
Solve the old issue #258 (the change was really simple, as the concerned field was already defined as text in the form, and in the json definitions of all the example tasks)

And also solved the new issue #756 (I implemented a temporary variable that stores the model during five seconds, and if needed, put it again in the map)

closes : #258, closes : #756


webapp/src/components/sidebar/ModelLibrary.vue
```diff
@@ -157,6 +157,8 @@ memoryStore.$subscribe((_, state) => {
 
 const memory = memoryStore.useIndexedDB ? new IndexedDB() : new EmptyMemory();
 
+let tempDeletedModel : { id : string; model : Model, metadata : ModelMetadata} | undefined;
```
JulienVig: @tharvik do you recommend using a ref here?

webapp/src/components/sidebar/ModelLibrary.vue
```diff
@@ -210,11 +212,34 @@ function sortModels(a: ModelMetadata, b: ModelMetadata): number {
...
-    onClick: () => deleteModel(modelID),
-    duration: 10000,
-  });
+async function deleteModelConfirm(modelID: string){
+  const modelInfo = models.value.find(([id]) => id === modelID);
+  const model = await memory.getModel(modelID);
+  if (!modelInfo) return
+
+  deleteModel(modelID).then(() => {
+    console.log("ici")
```
JulienVig:
```suggestion
```

webapp/src/components/sidebar/ModelLibrary.vue
```diff
@@ -210,11 +212,34 @@ function sortModels(a: ModelMetadata, b: ModelMetadata): number {
 }
 
 
-function deleteModelConfirm(modelID: string){
-  toaster.default(`Click here to confirm the model deletion`, {
-    onClick: () => deleteModel(modelID),
-    duration: 10000,
-  });
+async function deleteModelConfirm(modelID: string){
+  const modelInfo = models.value.find(([id]) => id === modelID);
```
JulienVig:
```suggestion
  const modelInfo = models.value.find(id => id === modelID);
```
I think you can unwrap the argument if you want

webapp/src/components/sidebar/ModelLibrary.vue
```diff
@@ -210,11 +212,34 @@ function sortModels(a: ModelMetadata, b: ModelMetadata): number {
...
+  const model = await memory.getModel(modelID);
+  if (!modelInfo) return
```
JulienVig:
```suggestion
  if (modelInfo === undefined) return
```
It's always better to triple equal check when testing undefined, JS has a lot of twisted logic with booleans

webapp/src/components/sidebar/ModelLibrary.vue
```diff
@@ -210,11 +212,34 @@ function sortModels(a: ModelMetadata, b: ModelMetadata): number {
...
+
+  deleteModel(modelID).then(() => {
+    console.log("ici")
+    tempDeletedModel = { id : modelID, model : model, metadata : modelInfo[1]};
+    toaster.info("Model deleted. Click here to undo", { 
+      onClick: () => undoDeleteModel(),
```
JulienVig:
```suggestion
      onClick:  undoDeleteModel,
```
If you want to be fancy

webapp/src/components/sidebar/ModelLibrary.vue
```diff
@@ -210,11 +212,34 @@ function sortModels(a: ModelMetadata, b: ModelMetadata): number {
...
+      duration : 10000,
+    });
+  })
+}
+
+async function undoDeleteModel() {
+  try {
+    if (!tempDeletedModel) throw new Error("No model to restore");
```
JulienVig:
```suggestion
    if (tempDeletedModel === undefined) throw new Error("No model to restore");
```

JulienVig (approved):
Great work!! Thanks a lot again. No blocking comment.

In the PR description, you can link issues that will automatically be closed when the PR is merged, e.g. `Closes #258, closes #756` : [github doc](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword). The keywords also work in commit messages

webapp/src/components/sidebar/ModelLibrary.vue
tharvik: hum, that shouldn't matter as it is not derived upon (no `compute`, no ui changes.

but it can be avoided if used as an argument to `undoDeleteModel` which would be more readable

tomasoignons: it was resolved by another branch this week

PR #775, opened 2024-09-11
Allow decentralized users to join late and catch up

### LINKED ISSUE #718: New users are not able to join once training started

JulienVig:
Closes #718 

#### Decentralized issues
Currently, we can only train decentralized with the exact number of peers specified in `minNbOfParticipants`. 
- No peer can leave because leaving mid training makes the round drop below the min number of participant and fail (have to abort the round)
- No more than `minNbOfParticipants` peers can join one round because the server sends the list of peers as soon as the threshold is reached, so the `minNbOfParticipants + 1`nth peer joining doesn't get the current round's peer list but is included in the next round's
Essentially, once the peers list has been sent, joining and leaving is not possible anymore.

#### Solution Implemented
1. During `onRoundBeginCommunication`, peers manifest their interest to join the current round: they send a `PeerJoinsRound` to the server. The server keeps a list of peers wanting to join (but doesn't reply with a peer list as it currently does).
2. The peers start training locally without the round's peer list
3. When peers are done training locally, they notify the server with a `PeerIsReady`, i.e. ready to exchange weight updates. The server waits until all the peers that sent a `PeerJoinsRound` sends their `PeerIsReady` and then send the round's peer list.
  i. This allows for some time for peers to join the round. To prevent new peers from continually joining and waiting for new peers to be ready, the server can stop including peers in this round as soon as one peer is ready (and include them in the next).
  ii. Peers can leave and notify the server before. As long as the peer list hasn't been sent, peers can join and leave without it being a problem.
4. Upon receiving the peer list, peers establish p2p connection and start exchanging weight updates.

- [x] Allow a peer to join a session that has already started.
- [x] If the number of peers drops below the `minNbOfParticipants` threshold, the peers wait for more participants
- [x] Peers leaving notifies the server which in turn notifies the remaining peers to wait.
- [x] More than exactly `minNbOfParticipants` peers can participate in the same round (because the `PeerJoinsRound` step allows some time for participants to join, instead of directly starting the weight update when `minNbOfParticipants` peers joined)

Refactoring
- `aggregator.add` returns a promise for the aggregated weights (no more need for the perpetual promise loop for the server controller) + new aggregator tests
- TaskInitializer implements discojs utils's `EventEmitter`
- Drop unused `trainingInformation.decentralizedSecure` and add `aggregationStrategy` ('mean' or 'secure')


discojs/src/aggregator/aggregator.ts
```diff
@@ -62,24 +62,53 @@ export abstract class Base<T> extends EventEmitter<{'aggregation': T }> {
...
 
-    // On every aggregation, update the object's state to match the current aggregation
-    // and communication rounds.
-    this.on('aggregation', () => this.nextRound())
+    // On each aggregation, increment
+    // updates the aggregator's state to proceed to the next communication round.
+    // If all communication rounds were performed, proceeds to the next aggregation round
+    // and empties the collection of stored contributions.
+    this.on('aggregation', () => {
+      if (++this._communicationRound === this.communicationRounds) {
```
tharvik: mixing test and actions is error prone (and makes my brain melt)

```suggestion
      this._communicationRound++;
      if (this.communicationRound === this.communicationRounds) {
```

discojs/src/aggregator/aggregator.ts
```diff
@@ -62,24 +62,53 @@ export abstract class Base<T> extends EventEmitter<{'aggregation': T }> {
...
    * Within one aggregation round there may be multiple communication rounds (such as for the decentralized secure aggregation
-   *  which requires multiple steps to obtain a global model)
+   * which requires multiple steps to obtain a global model)
    * The contribution will be aggregated during the next aggregation step.
    * @param nodeId The node's id
    * @param contribution The node's contribution
-   * @param round aggregation round of the contribution was made
    * @param communicationRound communication round the contribution was made within the aggregation round
-   * @returns boolean, true if the contribution has been successfully taken into account or False if it has been rejected
+   * @returns undefined, if the contribution is invalid, o.w. a promise for the weight update
```
tharvik:
```suggestion
   * @returns the aggregated weights
```

discojs/src/aggregator/aggregator.ts
```diff
@@ -62,24 +62,53 @@ export abstract class Base<T> extends EventEmitter<{'aggregation': T }> {
...
+   */
+  abstract add(nodeId: client.NodeID, contribution: WeightsContainer, communicationRound?: number): Promise<WeightsContainer>
```
tharvik: hum, we do have two way to get the aggregated weights, one via this promise, one via the event emitter. we should only keep one. (but yeay, no need to check for this weird boolean!)
I would argue that in many places, the promise gets `void`-ed instead of used
to quote [the zen of python](https://peps.python.org/pep-0020/): "There should be one-- and preferably only one --obvious way to do it."

EDIT: and one more via the `aggregate` method, consider making this one protected

discojs/src/aggregator/aggregator.ts
```diff
@@ -62,24 +62,53 @@ export abstract class Base<T> extends EventEmitter<{'aggregation': T }> {
...
+  protected createAggregationPromise(): Promise<WeightsContainer> {
+    if (this.isFull()) {
+      const aggregatedWeights = this.aggregate()
+      // Emitting the 'aggregation' communicates the aggregation to other clients and
+      // takes care of incrementing the round
+      this.emit('aggregation', aggregatedWeights)
+      return Promise.resolve(aggregatedWeights);
+    }
+    // Wait for the aggregation event to be emitted
+    return new Promise<WeightsContainer>((resolve) => this.once('aggregation', resolve))
```
tharvik: this is effectively equivalent, but a bit more readable (single exit).

```suggestion
    // Wait for the aggregation event to be emitted
    const ret = new Promise<WeightsContainer>((resolve) => this.once('aggregation', resolve));
    
    if (this.isFull()) {
      const aggregatedWeights = this.aggregate()
      // Emitting the 'aggregation' communicates the aggregation to other clients and
      // takes care of incrementing the round
      this.emit('aggregation', aggregatedWeights)
    }
    
    return ret
```

but I'm reading a bit more and looks like it's only when exiting the `add` method. consider splitting the `add` call, one public just a now but not abstract, one protected abstract (smth like `_add`) which does the internal work. the public `add` will then ensure that it uses `createAggregationPromise` (and potentially log before calling the internal `_add`)

discojs/src/aggregator/mean.spec.ts
```diff
@@ -11,44 +11,115 @@ async function WSIntoArrays(ws: WeightsContainer): Promise<number[][]> {
...
 
 describe("mean aggregator", () => {
-  it("updates only within round cutoff", () => {
-    const aggregator = new MeanAggregator(1, 3);
-    aggregator.setNodes(
-      Set.of("first client", "second client", "third client"),
-    );
+  it("updates only within round cutoff", async () => {
+    const aggregator = new MeanAggregator(1, 1, 'relative');
+    const [id1, id2, id3] = ["client 1", "client 2", "client 3"]
```
tharvik: consider inlining theses values. yes, it will duplicate the strings and might make more bugs than wanted when changing a test. but, a string is faster to read than a variable, and a test should be dead simple and not make fancy stuff (or delegate to a helper to do so). the easier a test is to read, the more correct it'll be.
the updability of a test shouldn't be hard, if it is, that indicate other issues: hard to use API, too complex test, unit-test that should be integration one, …

it's not important to fix it in theses tests, it's more of a general opinion on the matter

discojs/src/aggregator/mean.ts
```diff
@@ -98,38 +98,26 @@ export class MeanAggregator extends Aggregator<WeightsContainer> {
...
     nodeId: client.NodeID,
     contribution: WeightsContainer,
-    round: number,
-    currentContributions: number = 0,
-  ): boolean {
-    if (currentContributions !== 0)
-      throw new Error("only a single communication round");
-
-    if (!this.isValidContribution(nodeId, round)) return false
+  ): Promise<WeightsContainer> {
```
tharvik: we don't check that the contribution is valid anymore, is it wanted? if some devs forgets to check `Aggregator.isValidContribution` before, one can insert unwanted contribution in the aggregation

for the responsability of a given class, I usually consider that its methods should ensure that no invalid state can be entered, keeping it safe internally (in the spirit of [defensive programming](https://en.wikipedia.org/wiki/Defensive_programming)). devs/users will forget to call the check method at some point.

discojs/src/aggregator/mean.spec.ts
```diff
@@ -11,44 +11,115 @@ async function WSIntoArrays(ws: WeightsContainer): Promise<number[][]> {
...
+    const result1 = aggregator.add(id1, WeightsContainer.of([0], [1]));
+    // Make sure that the aggregation isn't triggered by waiting some time
+    let timeoutTriggered = false
+    await Promise.race([result1,
+      new Promise<void>(resolve => setTimeout(() => {
+        timeoutTriggered = true
+        resolve()
+      }, 300))
+    ])
+    expect(timeoutTriggered).to.be.true
```
tharvik: yeah, time based tests are cluncky (halting problem here we are), better to check that here, round is still zero and after awaiting the results, it is one

discojs/src/aggregator/mean.spec.ts
```diff
@@ -11,44 +11,115 @@ async function WSIntoArrays(ws: WeightsContainer): Promise<number[][]> {
...
```
tharvik: time based tests, same fix idea

(and I played with it a bit, you can avoid the `let` with that)

```suggestion
    const timeoutTriggered = await Promise.race([
      result1,
      new Promise<void>((resolve) => setTimeout(resolve, 300)),
    ]);
    expect(timeoutTriggered).to.be.undefined;
```

discojs/src/client/client.ts
```diff
@@ -6,24 +7,35 @@ import type { NodeID } from './types.js'
...
+
   /**
-   * The aggregator's result produced after aggregation.
+   * When the server notifies the client that they can resume training
+   * after waiting for more participants, we want to be able to display what
+   * we were doing before waiting (training locally or updating our model).
+   * We use this attribute to store the status to rollback to when we stop waiting
    */
-  protected aggregationResult?: Promise<WeightsContainer>
+  protected previousStatus: RoundStatus | undefined = undefined;
```
tharvik: only used in here, consider making it private. you can also avoid the `= undefined` as it is by default.

discojs/src/client/client.ts
```diff
@@ -95,4 +197,15 @@ export abstract class Client extends EventEmitter<{'status': RoundStatus}>{
...
+   * Whether the client should wait until more
+   * participants join the session, i.e. a promise has been created
+   */
+  get waitingForMoreParticipants(): boolean {
+    return this.promiseForMoreParticipants !== undefined 
+  }
+
+  protected shortId(id: string): string{
+    return id.slice(0, 4)
+  }
```
tharvik: doesn't need access to `this`, consider extracting it as a standalone function

discojs/src/client/client.ts
```diff
@@ -63,23 +175,13 @@ export abstract class Client extends EventEmitter<{'status': RoundStatus}>{
...
-  // If federated, it should the number of participants excluding the server
-  // If local it should be 1
+  * Number of contributors to a collaborative session
+  * If decentralized, it should be the number of peers
+  * If federated, it should the number of participants excluding the server
+  * If local it should be 1
+  */
   get nbOfParticipants(): number {
-    return this.aggregator.nodes.size // overriden by the federated client
+    return 1 // overriden by the federated and decentralized clients
```
tharvik: would be a good place for `abstract` but abstract accessor doesn't exist in this language.. either move it to a full method `getNbOfParticipants(): number` or at least make it throw a "should be overriden by subclass". better to fail loudly rather than returning an erroneous value.

discojs/src/client/client.ts
```diff
@@ -45,8 +69,96 @@ export abstract class Client extends EventEmitter<{'status': RoundStatus}>{
...
+   * @param setMessageInversionFlag function flagging whether a message inversion occurred
+   * between a NewNodeInfo message and an EnoughParticipant message.
+   */
+  protected setupServerCallbacks(setMessageInversionFlag: () => void) {
+    // Setup an event callback if the server signals that we should 
+    // wait for more participants
+    this.server.on(type.WaitingForMoreParticipants, () => {
+      debug(`[${this.ownId.slice(0, 4)}] received WaitingForMoreParticipants message from server`)
+      // Display the waiting status right away
+      this.emit("status", "NOT ENOUGH PARTICIPANTS")
```
tharvik: shouldn't it be saveAndEmit here?

discojs/src/client/client.ts
```diff
@@ -45,8 +69,96 @@ export abstract class Client extends EventEmitter<{'status': RoundStatus}>{
...
+   * 1. Client 1 connect to the server
+   * 2. Server answers with message A containing "not enough participants"
+   * 3. Before A arrives a new client joins. There are enough participants now.
+   * 4. Server updates client 1 with message B saying "there are enough participants"
+   * 5. Due to network and message sizes message B can arrive before A. 
+   * i.e. "there are enough participants" arrives before "not enough participants"
+   * ending up with client 1 thinking it needs to wait for more participants.
+   * 
+   * To keep track of this message inversion, `setMessageInversionFlag` 
+   * tells us whether a message inversion occurred (by setting a boolean to true)
```
tharvik, line 101: note to future: we should change the way the messages are forwarded around. it shouldn't be reordering, especially when the link isn't doing so. we should also have a nice way to "split" a message stream for specific module, such as one for the aggregation itself, one for orchestring the nodes, one for getting the list of tasks, …

discojs/src/client/client.ts
```diff
@@ -45,8 +69,96 @@ export abstract class Client extends EventEmitter<{'status': RoundStatus}>{
...
+   * participants (anymore) to start/continue training
+   * The method creates a promise that will resolve once the server notifies 
+   * the client that the training can resume via a subsequent EnoughParticipants message
+   * @returns a promise which resolves when enough participants joined the session
+   */
+  protected async createPromiseForMoreParticipants(): Promise<void> {
+    return new Promise<void>((resolve) => {
+      // "once" is important because we can't resolve the same promise multiple times
+      this.server.once(type.EnoughParticipants, () => {
+        debug(`[${this.ownId.slice(0, 4)}] received EnoughParticipants message from server`)
```
tharvik: can use `shortId` (two more in this file)

discojs/src/client/client.ts
```diff
@@ -45,8 +69,96 @@ export abstract class Client extends EventEmitter<{'status': RoundStatus}>{
...
+    // Setup an event callback if the server signals that we should 
+    // wait for more participants
+    this.server.on(type.WaitingForMoreParticipants, () => {
+      debug(`[${this.ownId.slice(0, 4)}] received WaitingForMoreParticipants message from server`)
+      // Display the waiting status right away
+      this.emit("status", "NOT ENOUGH PARTICIPANTS")
+      // Upon receiving a WaitingForMoreParticipants message,
+      // the client will await for this promise to resolve before sending its
+      // local weight update
+      this.promiseForMoreParticipants = this.createPromiseForMoreParticipants()
```
tharvik, line 118: `this.promiseForMoreParticipants` should be undef before that, no? can you enforce it by checking/throw before it? (can help to notice if the server is misbehaving)

discojs/src/client/client.ts
```diff
@@ -45,8 +69,96 @@ export abstract class Client extends EventEmitter<{'status': RoundStatus}>{
...
+      this.server.once(type.EnoughParticipants, () => {
+        debug(`[${this.ownId.slice(0, 4)}] received EnoughParticipants message from server`)
+        // Emit the last status emitted before waiting if defined
+        if (this.previousStatus !== undefined) this.emit("status", this.previousStatus)
+        resolve()
+      })
+    })
+  }
 
+  protected async checkIfWaitForParticipants(): Promise<void>{
```
tharvik: hum, "check" makes me think of simply returning a boolean, a rapid test. wdyt of "waifForParticipantsIfNeeded" or "waitForEnoughParticipants"? it's an action which makes more sense for me

discojs/src/client/decentralized/decentralized_client.ts
```diff
@@ -22,9 +22,14 @@ export class DecentralizedClient extends Client {
...
    */
-  private pool?: PeerPool
-  private connections?: Map<NodeID, PeerConnection>
-  
+  #pool?: PeerPool
+  #connections?: Map<NodeID, PeerConnection>
+  #roundParticipants = 1
+
+  get nbOfParticipants(): number {
+    return this.#roundParticipants
```
tharvik: maybe we can avoid `#roundParticipants` entirely?

```suggestion
    return this.aggregator.nodes.size
```

discojs/src/client/decentralized/decentralized_client.ts
```diff
@@ -177,60 +208,69 @@ export class DecentralizedClient extends Client {
    * @param connections 
    * @param round 
    */
-  private receivePayloads (connections: Map<NodeID, PeerConnection>, round: number): void {
+  private receivePayloads (connections: Map<NodeID, PeerConnection>): void {
     connections.forEach(async (connection, peerId) => {
-      let currentCommunicationRounds = 0
       debug(`waiting for peer ${peerId}`);
-      do {
+      for (let _ = 0; _ < this.aggregator.communicationRounds; _++) {
```
tharvik: better to use more readable variable name, not the somewhat standard "ignore" variable name

```suggestion
      for (let r = 0; r < this.aggregator.communicationRounds; r++) {
```

discojs/src/client/decentralized/decentralized_client.ts
```diff
@@ -177,60 +208,69 @@ export class DecentralizedClient extends Client {
...
     // the aggregator to define any complex multi-round aggregation mechanism.
     let result = weights;
-    for (let r = 0; r < this.aggregator.communicationRounds; r++) {
+    for (let communicationRound = 0; communicationRound < this.aggregator.communicationRounds; communicationRound++) {
       // Generate our payloads for this communication round and send them to all ready connected peers
-      if (this.connections !== undefined) {
+      if (this.#connections !== undefined) {
         const payloads = this.aggregator.makePayloads(result)
         try {
           await Promise.all(payloads.map(async (payload, id) => {
```
tharvik: ouch, this doesn't do what's seems to be advertised
`payloads.map(…) -> Map<NodeID, Promise<…>>` which when iterated upon via `Promise.all` gives `Promise<Array<[NodeID, Promise<…>]>>` so not waiting for the payloads' new values.

I think that it was meant to be a forEach (and to `void` the `aggregator.add`). if that's correct, the try/catch isn't doing a thing as the promises' potential errors are never actually read

```suggestion
          payloads.forEach(async (payload, id) => {
```

discojs/src/default_tasks/mnist.ts
```diff
@@ -32,7 +32,7 @@ export const mnist: TaskProvider = {
         preprocessingFunctions: [data.ImagePreprocessing.Resize, data.ImagePreprocessing.Normalize],
         LABEL_LIST: ['0', '1', '2', '3', '4', '5', '6', '7', '8', '9'],
         scheme: 'decentralized',
-        decentralizedSecure: true,
+        aggregationStrategy: 'mean',
```
tharvik: only "mean" strategy is actually set, maybe a "secure" one here?

discojs/src/training/disco.ts
```diff
@@ -22,11 +22,10 @@ interface DiscoConfig {
...
 
-export type RoundStatus =
-  "Waiting for more participants" |
-  "Retrieving peers' information" |
-  "Updating the model with other participants' models" |
-  "Training the model on the data you connected"
+export type RoundStatus = 'NOT ENOUGH PARTICIPANTS' | // Server notification to wait for more participants
+  'UPDATING MODEL' | // fetching/aggregating local updates into a global model
+  'TRAINING' | // Training the model locally
+  'RETRIEVING PEERS' // for decentralized only, fetch the server's list of participating peers
```
tharvik: SCREAMING CASE #nitpicking it reads very loud for me

discojs/src/utils/event_emitter.ts
```diff
@@ -2,17 +2,21 @@
...
  */
 export class EventEmitter<I extends Record<string, unknown>> {
-  private listeners: {
+  // List of callbacks to run per event
+  #listeners: {
     [E in keyof I]?: List<[once: boolean, _: Listener<I[E]>]>;
   } = {}
+  // Keep a list of the previously emitted events
+  // to allow subscribers to run callbacks on past events if needed
+  #pastEvents: { [E in keyof I]?: List<I[E]> } = {}
```
tharvik: yeah, I was thinking of doing that at some point but I don't think that's a good idea: it uses an ever growing amount of memory so it will exhaust the system it's running on. to avoid that, the listeners should either register before it starts emitting event or find another way to get the previous state from the implementing class.
there is a similar case that get's solved in `server.TaskInitializer`: the list of `Task` is kept around so that new listener can still know what are the current state is when registering. (now that I read this class, one can also replay the tasks when some listener joins and avoid exposing the state, but yeah, at a latter point)

server/src/controllers/decentralized_controller.ts
```diff
@@ -14,23 +14,22 @@ import MessageTypes = client.messages.type
...
-   * Set of nodes who have contributed.
-   */
-  private readyNodes = Set<client.NodeID>()
-  /**
-   * Map associating node ids to their open WebSocket connections.
-   */
-  private connections: Map<client.NodeID, WebSocket> = Map()
+  // Map of nodes who want to join the round.
+  // The boolean value indicates if the node is ready to exchange weight updates (i.e.
+  // the node has already sent a PeerIsRead message)
```
tharvik:
```suggestion
  // the node has already sent a PeerIsReady message)
```

server/src/controllers/decentralized_controller.ts
```diff
@@ -105,5 +92,69 @@ export class DecentralizedController extends TrainingController {
...
+          participantWs.send(msgpack.encode(msg))
+        })
+    }) 
+  }
+  /**
+   * Check if we have enough participants to start the training
+   * and if all peers that joined the round are ready to exchange weight updates
+   * If so, send the list of peers for this round to all participants
+   */
+  private checkThenSendPeersForRound(): void {
```
tharvik: same argument as before: "check" doesn't look like it's doing smth big. `sendPeersForRoundIfNeeded`?

server/src/controllers/decentralized_controller.ts
```diff
@@ -105,5 +92,69 @@ export class DecentralizedController extends TrainingController {
...
+    const minNbOfParticipants = this.task.trainingInformation.minNbOfParticipants
+    const nbOfPeersReady = this.#roundPeers.filter(ready => ready).size
+    // First check there are enough participants to start the round
+    // Then check if all peers that wanted to join this round are ready
+    if (nbOfPeersReady >= minNbOfParticipants && nbOfPeersReady == this.#roundPeers.size) {
```
tharvik: #style
most of the function is inside this `if`, inversing the condition and returning would make it a bit more readable: `if (cond) { … }` into `if (!cond) return; …`

server/src/controllers/decentralized_controller.ts
```diff
@@ -14,23 +14,22 @@ import MessageTypes = client.messages.type
...
-  /**
-   * Map associating node ids to their open WebSocket connections.
-   */
-  private connections: Map<client.NodeID, WebSocket> = Map()
+  // Map of nodes who want to join the round.
+  // The boolean value indicates if the node is ready to exchange weight updates (i.e.
+  // the node has already sent a PeerIsRead message)
+  // We wait for all peers to be ready to exchange weight updates
+  #roundPeers = Map<client.NodeID, boolean>()
+  #aggregationRound = 0
```
tharvik, line 22: it is the same as the general round, no? renaming to "round" is clearer IMO

server/src/controllers/federated_controller.ts
```diff
@@ -25,95 +23,24 @@ export class FederatedController extends TrainingController {
...
-   * i.e. waiting for more participants or resuming training
-   */
-  #participants = Map<string, WebSocket>()
+  
 
   constructor(task: Task, initialWeights: EncodedWeights) {
     super(task)
     this.#latestGlobalWeights = initialWeights
-    // start the perpetual promise loop
-    void this.storeAggregationResult()
```
tharvik, line 55: :partying_face:

server/src/controllers/federated_controller.ts
```diff
@@ -25,95 +23,24 @@ export class FederatedController extends TrainingController {
...
-          nbOfParticipants: this.#participants.size
-        }
-        ws.send(msgpack.encode(msg))
-      })
-      .catch((e) => debug("while waiting for weights: %o", e))
+    // Save the latest weight updates to be able to send it to new or outdated clients
+    this.#aggregator.on('aggregation', (weightUpdate) => {
+      serialization.weights.encode(weightUpdate)
+        .then(serializedWeights => { this.#latestGlobalWeights = serializedWeights })
+        .catch((e) => debug("while encoding latest weight update: %o", e))
```
tharvik: you can simplify it to a full fledge "async function" now that you added support for it in EventEmitter (also, catching here seems weird, it's not really smth that can fail, or if it does, we really should know about it)

server/src/controllers/training_controller.ts
```diff
@@ -18,10 +24,40 @@ import type { Task } from '@epfml/discojs'
...
+   */
+  protected connections = Map<client.NodeID, WebSocket>()
 
   constructor(protected readonly task: Task) { }
 
   abstract handle(
     ws: WebSocket
   ): void
+
+  protected checkIfEnoughParticipants(waitForMoreParticipants: boolean, currentId: client.NodeID) {
```
tharvik: same "check" comment

tharvik: also, "waitForMoreParticipants" is always equals to `this.connections.size < this.task.trainingInformation.minNbOfParticipants`, consider not having the argument at all

server/src/controllers/training_controller.ts
```diff
@@ -18,10 +24,40 @@ import type { Task } from '@epfml/discojs'
...
+    // If we were previously waiting for more participants to join and we now have enough,
+    // broadcast to previously waiting participants that the training can start
+    if (this.waitingForMoreParticipants && !waitForMoreParticipants) {
+      this.connections
+        // filter out the client that just joined as 
+        // it already knows via the NewFederatedNodeInfo message
+        .filter((_, id) => id !== currentId)
```
tharvik: this is equivalent to `this.connections.delete(currentId)` (this one being O(1) rather than O(n))

server/src/controllers/federated_controller.ts
```diff
@@ -225,26 +144,26 @@ export class FederatedController extends TrainingController {
...
 
-      this.#waitingForMoreParticipants = true
+      this.waitingForMoreParticipants = true
       // Tell remaining participants to wait until more participants join
-      this.#participants
+      this.connections
         .forEach((participantWs, participantId) => {
           debug("Telling remaining client [%s] to wait for participants", participantId.slice(0, 4))
-          const msg: FederatedMessages.WaitingForMoreParticipants = {
+          const msg: client.messages.WaitingForMoreParticipants = {
```
tharvik: kind of a simliar behavior in decentralized and federated, consider commonizing it by pushing it up in the class hierarchy, as already done with checkIfEnoughParticipants

server/src/task_initializer.ts
```diff
@@ -31,31 +31,16 @@ import type { EncodedModel } from '@epfml/discojs'
  * and objects depending on tasks and models can subscribe to 
  * the 'newTask' event to run callbacks whenever a new Task and EncodedModel are initialized.
  */
-export class TaskInitializer {
-  // List of callback to apply to future task-model pairs added
-  private listeners = List<(t: Task, m: EncodedModel) => Promise<void>>()
+export class TaskInitializer extends EventEmitter<{
```
tharvik: while at it, I don't really like the name of this class, it doesn't initialize a task really, rather it loads models based on passed task. looks to me more like something containing tasks and models, so smth like LoadedTasks/TasksSet?
initializer doesn't ~sound like a container
naming is the first issue of programming :shrug:

server/tests/e2e/decentralized.spec.ts
```diff
@@ -114,4 +119,148 @@ describe('end-to-end decentralized', function () {
...
     await reachConsensus('secure', 3)
   })
+
+  it("peers emit expected statuses", async function () {
+    this.timeout(15_000);
+    const lusCovidTask = defaultTasks.lusCovid.getTask();
+    lusCovidTask.trainingInformation.aggregationStrategy = 'mean';
+    lusCovidTask.trainingInformation.epochs = 8;
+    lusCovidTask.trainingInformation.roundDuration = 2;
+    lusCovidTask.trainingInformation.minNbOfParticipants = 2;
```
tharvik: less character, mr rdbl

```suggestion
    const lusCovidTask = {
      ...defaultTasks.lusCovid.getTask(),
      aggregationStrategy: "mean",
      epochs: 8,
      roundDuration: 2,
      minNbOfParticipants: 2,
    };
```

discojs/src/training/disco.ts
tharvik: "RETRIEVING PEERS" is when we finish the round and that we're connecting to other peers, rather smth like "CONNECTING PEERS"?

server/tests/e2e/decentralized.spec.ts
```diff
@@ -114,4 +119,148 @@ describe('end-to-end decentralized', function () {
...
+     * Calling next() again will then run c), as well as a) and b) again.
+     * 
+     * In this test the timeline is:
+     * - User 1 joins the task by themselves
+     * - User 2 joins
+     * - User 1 leaves
+     * - User 3 joins
+     * - User 2 & 3 leave
+     */
+    const statusUpdateTime = 500 // allow some time for the client to update their status
```
tharvik, line 173: yeah, time based tests are broken; using [an event queue](https://github.com/epfml/disco/blob/88b20b34b14d9cecf196e081d984c6a2e4413d65/server/tests/e2e/federated.spec.ts#L26) (from #781) to check value changes is safer and faster. you can ensure that a status didn't "flicker" and that it hadn't gone through unexpected status before stabilisation.
I actually did it in there as the tests were very regularly failing on my laptop.

server/tests/e2e/decentralized.spec.ts
```diff
@@ -114,4 +119,148 @@ describe('end-to-end decentralized', function () {
...
+
+    // Create User 1, set the task's scheme to decentralized (federated by default)
+    const discoUser1 = new Disco(lusCovidTask, url, { scheme: "decentralized" });
```
tharvik: you can set the scheme when setting up the task and avoid passing it the `new Disco`

server/tests/e2e/decentralized.spec.ts
```diff
@@ -114,4 +119,148 @@ describe('end-to-end decentralized', function () {
...
+    expect(statusUser3.last()).equal("TRAINING")
+    // User 2 is still in c) waiting for user 3 to be ready to exchange waits
+    expect(statusUser2.last()).equal("RETRIEVING PEERS")
+    
+    // User 3 notifies the server that they are ready to exchange waits
+    // then user 2 and 3 exchange weight updates
+    const logUser3Round3 = await generatorUser3.next()
+    const logUser2Round3 = await logUser2Round3Promise // the promise can resolve now
+    if (logUser3Round3.done || logUser2Round3.done)
+      throw Error("User 1 or 2 finished training at the 3nd round")
```
tharvik: yeah, that's the way to check for iteration results state. TypeScript doesn't infer that `iff not done then value is valid` with function, only via `if` (it would require way more sophisticated typing means).
that allow to not cast to `RoundLogs` (as done above)

webapp/src/components/training/Trainer.vue
```diff
@@ -59,7 +59,7 @@
             <!-- Display an activity indicator depending on the training status -->
             <div class="min-h-9">
               <div v-if="roundStatus === 'Waiting for more participants' ||
-                roundStatus === 'Retrieving peers\' information'"
+                roundStatus === 'Establishing peer-to-peer connections'"
```
tharvik: checking on such strings is quite error-prone (it will silently missbehave when changing the formatted text). updating `roundStatus` to `ref<[RoundStatus, string]>` helps by having both a safe value to check (`RoundStatus`) and a formatted string to print

tharvik (changes requested):
huge work, thank you! finally got rid of the mess in the server with promises :star_struck:

some potential improvement but nothing serious

discojs/src/client/decentralized/decentralized_client.ts
JulienVig: hmm I had to rely on another variable because the aggregator's nodes are reset after exchanging the weights (and forgot to document it) but thinking about it now I think we can simply stop resetting the nodes.

discojs/src/aggregator/mean.ts
JulienVig: Adding the check back in the public add method!

discojs/src/client/client.ts
JulienVig: I don't think so, we never need to rollback to Not enough participants.  We are actually using saveAndEmit in the other places so that when we emit "not enough participants" here, we are able rollback the previous status in the future.

discojs/src/client/client.ts
JulienVig: I completely agree, do you know of paradigms/libraries/documentation on how to manage that? (I'll research it myself, just in case you already have some knowledge on the subject)

server/src/controllers/decentralized_controller.ts
JulienVig: I had to stare at it a good 1min to notice the difference haha

server/src/controllers/decentralized_controller.ts
JulienVig: I was actually getting confused with the rounds because it's hard to know when using `round` if it is always the aggregation round or if we can infer from the context that it is the communication round (or even another one), so I would prefer expliciting the round type every time we use it and never use `round`. What do you think?

discojs/src/client/client.ts
tharvik: I've been reflecting a while on how to best approach it. I'm still unsure, one would need to validate it with some code to know how well it works, but it would already help us see more clearly:

let's first remark that many parts requires a different set of message (federated vs decentralized, secure aggregator vs some other) yet are still used in the same fashion. so it seems that we can plug a class with various implementations.

we can have an abstract `NetworkClient` with various implementation, one federated, one decentralized (and one local). theses advertise their supported messages by exposing a `isValidMessage(raw: unknown): raw is Message` (with `Message` depending on the implementation). that ensure that the type of messages we are receiving are typed correctly. it then gets passed an async queue of messages (such as [`reactive-channel`](https://www.npmjs.com/package/reactive-channel)'s one).
same idea for `AggregatorClient`

`Client` should do more, and not be an abstract class but be constructed with `NetworkClient`, `AggregatorClient`, …. it should take care of splitting the message stream, ie read next msg and put it the queue where the corresponding `isValidMessage` succeeded.

that helps by having a local set of messages per implementation, and only care about these ones. we can also better check that the server is not weirdly behaving by failing when the next messages is unexpected. and message order is preserved because async channels are also ordered and there is a single reader of WebSocket.

with that, it forces us to check that we don't receive an unexpected message in the stream. that can be a bit bothering in the long run (I've other ways to fix that such as type-encoding a state-transition table, but I might be going a bit far for now). but overall, I do think that can greatly help us understand where messages are coming from and in which state we are.
WDYT for a first draft?

server/src/controllers/decentralized_controller.ts
tharvik: yeah, good idea, explicit is better than implicit.
hopefully, we will get rid of this confusing "communication round" (keeping it internal to the aggregators).

server/tests/e2e/decentralized.spec.ts
JulienVig: I wanted to explicit the fact that I was overwriting the default scheme (which is federated for lus_covid) but not very important

server/tests/e2e/decentralized.spec.ts
JulienVig: I think it should be something like:
```js
const lusCovidTask = {
      ...defaultTasks.lusCovid.getTask(),
      trainingInformation: {
        ...defaultTasks.lusCovid.getTask().trainingInformation,
        scheme: "decentralized",
        aggregationStrategy: "mean",
        epochs: 8,
        roundDuration: 2,
        minNbOfParticipants: 2,
      } as TrainingInformation
    };
```
I find this harder to read than setting parameters one by one but maybe the nested spread can be simplified.

JulienVig: Actually I find this one better:
```js
const lusCovidTask = defaultTasks.lusCovid.getTask();
lusCovidTask.trainingInformation = {
  ...lusCovidTask.trainingInformation,
  scheme: 'decentralized',
  aggregationStrategy: 'mean',
  epochs: 8,
  minNbOfParticipants: 2,
}
```

discojs/src/utils/event_emitter.ts
JulienVig: I'm pushing adding the tasks to the server until after the routers are initialized and subscribed to TaskInitializer (renamed to TaskSet)

discojs/src/client/decentralized/decentralized_client.ts
JulienVig: Good catch! I removed the try catch but I'm not sure if that's what you implied

discojs/src/client/client.ts
JulienVig: Yes that sounds good! Already using a queue would ensure sending messages in order and using different message channels makes things extra clear. And the server would also need to rely on the same message queues I think. Do you mind creating an issue to keep track of this?

discojs/src/aggregator/aggregator.ts
JulienVig: I went with the 'aggregation' event!

tharvik (approved):
superbe, thanks for the hard work!

Issue #758, opened 2024-08-26
use WebWorker for training


tharvik:
@martinjaggi pointed out that the training can freeze the UI, so much that it seems to jump from start to finish.
using [WebWorkers](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API) should make it way more reactive as computation would be separated from UI rendering.

Issue #649, opened 2024-03-12
Disco preprocessing is too constraining


JulienVig:
A recent [refactoring](https://github.com/epfml/disco/commit/0892e405036ef46f9f8948be0795dff0bf8089d7) enforced tasks' preprocessing in a lazy and streaming fashion. The preprocessing first defines stateless preprocessing functions (e.g. `resize` and `normalize` for images) and then apply them successively on the dataset one row at a time:
```js
// The preprocessing functions are applied as a map on the dataset:
this.dataset.map(this.preprocessing)
```

Limitations to address:
- [ ] This functional preprocessing doesn't allow for "stateful" preprocessing. Because of its streaming nature, it is currently not possible to normalize a tabular column since we can't compute general aggregations (e.g. mean and standard deviation of a feature). In other words, if we want to implement new preprocessing functions we can only add functions with one dataset row as sole argument which is very constraining.
- [ ] The preprocessing state learned during training should be saved to be re-used for test and inference. For example, standardizing the testing set should be done with the training set's mean and standard deviation, not the test's statistics. Therefore, the preprocessing state should be saved and this is currently not supported.

JulienVig:
#781 may allow more flexibility for preprocessing

tharvik:
> #781 may allow more flexibility for preprocessing

it should, it should. answering based on a comment [you made there](https://github.com/epfml/disco/pull/781#discussion_r1776854353) and some other comments

> choosing how to handle missing values [in tabular] should be the responsibility of the data owner so we could throw an error or drop the whole row as default behavior.

totally! for now, how I see the usage of discojs is as follow:
* for easy access, use `Disco` and let it transform data via the `preprocess` function
  * here, we can be opiniated, and actively choose what the behavior should be (default/throw/drop/…)
  * datatype specific processing are set in `Task`
* for more advance usage, direclty use `Trainer` and process the data yourself
  * it should be quite straightforward to do so yet allow any error handling

in the Disco case, I'm thinking of throwing now, letting the user know that the dataset is missing important values rather than choosing ourself how to handle it (currently, titanic's  is missing some values).

> * [ ]  This functional preprocessing doesn't allow for "stateful" preprocessing. Because of its streaming nature, it is currently not possible to normalize a tabular column since we can't compute general aggregations (e.g. mean and standard deviation of a feature). In other words, if we want to implement new preprocessing functions we can only add functions with one dataset row as sole argument which is very constraining.


[...]

PR #781, opened 2024-09-17
*: framework agnostic preprocessing

tharvik:
on the road to (closer by every PR) ONNX' land.

preprocessing revamp
* simplifing preprocessing by using flat functions rather than classes
  * allows to change type with every function rather than using the way-too-generic `tf.TensorContainer`
  * for example, force the users to normalize the images before passing theses to a model by having the latter ingesting `NormalizedImage`
* get rid of `{Image,Tabular,Text}Data` classes by replacing theses by correctly typed dataset

dataset improvements
* add `Dataset.cached` to avoid recomputation when iterating multiple times
  * is memory sensitive so can work with any dataset size (but not very useful if dataset materialization is bigger than RAM)
* add `preprocessOnce` option to `Disco` to avoid all recomputation by keeping a preprocessed version of memory (more hardcore version of `Dataset.cached`)
  * can crash the JS VM with OOM but yeah, let the user make the call
  * quite useful to get reproducable timing for tests
* `Dataset.batch` constructs batch concurrently

type safety
* make `Model`, `Task`, `Trainer`, … generic on their datatype (`"image" | "tabular" | "text"`) to check soundness _at compile time_
  * ensure that when defining a `Task<'tabular'>`, is does have the required `{input,output}Columns` 
  * ensure that a `{Disco,Trainer,Validator}<'image'>` doesn't get feed a `Dataset<{Tabular,Text}>` but only a `Dataset<[Image, label: string]>`
* single `TrainingInformation<'tabular'>.outputColumn` instead of multiple colum
  * we in fact only supported single output but it was an array
* clear separation between what is outside types ingested by `Disco` & `Validator` (`Raw`), and what is internal (`ModelEncoded`, after preprocessing and before postprocessing)

misc
* change server/tests/status to not wait for changes on timeout but with promises in order to avoid slower computer timeouts
* reduce discojs/validator/ to a single file
* bump deps

sadly, there is a performance drawback to that: as the _preprocessing_ computation is solely done in JS which is fundamentally monothread (as python is), processing is slower than the tfjs-node accelerated version we have. I tried a few ways to workaround it and push V8 to do magic but it didn't work. there is some way to make make all that work (#758) but that would require platform specific accelerators. and this PR was already a bit long so pushing it to another iteration.


JulienVig:
Encountered this error on Mac M1 while reviewing:
```bash
Error: Cannot find module @rollup/rollup-darwin-arm64. npm has a bug related to optional dependencies (https://github.com/npm/cli/issues/4828). Please try 'npm i again after removing both package-lock.json and node_modules directory.
at requireWithFriendlyError (/Users/vignoud/Disco/review/node_modules/rollup
/dist/native.js: 59:9)
at Object. ‹anonymous> (/Users/vignoud/Disco/review/node_modules/rollup/dist/
native. js: 68:76)
•. 3 lines matching cause stack trace ...
at Module._load (node:internal/modules/cjs/loader:1023:12) at cjsLoader (node:internal/modules/esm/translators:356:17)
at ModuleWrap. ‹anonymous> (node:internal/modules/esm/translators:305:7) at ModuleJob. run (node:internal/modules/esm/module_job:218:25)
at async ModuleLoader. import (node:internal/modules/esm/loader:329:24) {
  [cause]: Error: Cannot find module '@rollup/rollup-darwin-arm64'
  Require stack:
    - /Users/vignoud/Disco/review/node_modules/rollup/dist/native.js
    at Module._resolveFilename (node:internal/modules/cjs/loader:1144:15)
    at Module._load (node:internal/modules/cjs/loader:985:27) at Module. require (node:internal/modules/cjs/loader:1235:19)
    at require (node:internal/modules/helpers:176:18)
    at requireWithFriendlyError (/Users/vignoud/Disco/review/node_modules/roll up/dist/native.js:41:10)
    at Object. ‹anonymous> (/Users/vignoud/Disco/review/node_modules/rollup/dis
    t/native. js: 68:76)
    at Module._compile (node:internal/modules/cjs/loader:1376:14)
    at Module._extensions..js (node: internal/modules/cjs/Loader:1435:10) at Module. load (node:internal/modules/cjs/loader:1207:32)
    at Module._load (node:internal/modules/cjs/loader: 1023:12) {
    code: 'MODULE_NOT_FOUND' requireStack: [
    '/Users/vignoud/Disco/review/node_modules/rollup/dist/native.js'
    ]
  }
｝
```
Fixed by re-cloning the repository.

server/tests/e2e/federated.spec.ts
```diff
@@ -23,20 +23,42 @@ async function arrayFromAsync<T>(iter: AsyncIterable<T>): Promise<T[]> {
...
 
+class Queue<T> {
+  #content = List<T>();
+
+  put(e: T) {
+    this.#content = this.#content.push(e);
+  }
+
+  async next(): Promise<T> {
+    for (;;) {
```
JulienVig: Do you have a preference for an empty for over a while(true)?

discojs/src/processing/image.ts
```diff
@@ -0,0 +1,135 @@
...
+> {
+  // private as it doesn't check that array content is valid
+  private constructor(
+    public readonly data: Readonly<Float32Array>,
+    public readonly width: W,
+    public readonly height: H,
+    public readonly depth: D,
+  ) {
+    if (data.length != width * height * depth)
+      throw new Error("data isn't of excepted size");
```
JulienVig:
```suggestion
      throw new Error("data isn't of expected size");
```

discojs/src/processing/image.ts
```diff
@@ -0,0 +1,135 @@
+import { Repeat, Seq } from "immutable";
+import { createJimp } from "@jimp/core";
+import * as jimpResize from "@jimp/plugin-resize";
+
+import { Image } from "../index.js";
+
+/** Image where intensity is represented in the range 0..1 */
+export class NormalizedImage<
+  D extends 1 | 3 | 4 = 1 | 3 | 4,
```
JulienVig, line 9: If I understand correctly:
- Depth of 1: grey scale no alpha channel
- Depth of 3: color without alpha channel
- Depth of 4: color with alpha
Is this correct?
Why not include a depth of 2?

discojs/src/processing/image.ts
```diff
@@ -0,0 +1,135 @@
...
+  image: Image<3 | 4, W, H>,
+): Image<4, W, H> {
+  switch (image.depth) {
+    case 3:
+      return new Image(
+        Uint8Array.from(
+          Seq(image.data).flatMap((v, i) => {
+            if (i % 3 !== 2) return [v];
+            else return [v, 0xff];
+          }),
```
JulienVig: Could you add a comment explaining what this line is doing?

discojs/src/processing/image.ts
```diff
@@ -0,0 +1,135 @@
...
+export function removeAlpha<W extends number, H extends number>(
+  image: Image<1 | 3 | 4, W, H>,
+): Image<1 | 3, W, H> {
+  switch (image.depth) {
+    case 1:
+    case 3:
+      return new Image(image.data, image.width, image.height, image.depth);
+    case 4:
+      return new Image(
+        image.data.filter((_, i) => i % 4 !== 3),
```
JulienVig, line 84: I think it would be worth adding documentation to the Image class that `data` is represented as a 1d array where "congruence classes match dimensions", I would have expected arrays of arrays (but 1d array is more efficient)

discojs/src/processing/image.ts
```diff
@@ -0,0 +1,135 @@
...
+  image: Image<4, W, H>,
+): Image<3, W, H>;
+export function removeAlpha<
+  D extends 1 | 3,
+  W extends number,
+  H extends number,
+>(image: Image<D | 4, W, H>): Image<D, W, H>;
+export function removeAlpha<W extends number, H extends number>(
+  image: Image<1 | 3 | 4, W, H>,
+): Image<1 | 3, W, H> {
```
JulienVig, line 77: I'm not sure what's going on here. It seems you are overloading the function but without providing an implementation?

discojs/src/processing/index.ts
```diff
@@ -0,0 +1,172 @@
...
+    }
+    case "text": {
+      // cast as typescript doesn't reduce generic type
+      const d = dataset as Dataset<Raw["text"]>;
+      const t = task as Task<"text">;
+
+      const tokenizer = await models.getTaskTokenizer(t);
+      const maxLength =
+        t.trainingInformation.maxSequenceLength ??
+        (tokenizer.model_max_length as number) + 1;
```
JulienVig: Can we add a comment here explaining that the + 1 is to include the label for next token prediction? (if I'm not mistaken)

discojs/src/processing/index.ts
```diff
@@ -0,0 +1,172 @@
...
+  }
+}
+
+export async function preprocessWithoutLabel<D extends DataType>(
+  task: Task<D>,
+  dataset: Dataset<RawWithoutLabel[D]>,
+): Promise<Dataset<ModelEncoded[D][0]>> {
+  switch (task.trainingInformation.dataType) {
+    case "image": {
+      // cast as typescript doesn't reduce generic type
```
JulienVig, line 82: what do you mean by "reduce" ? deduce?

discojs/src/processing/index.ts
```diff
@@ -0,0 +1,172 @@
...
+      const tokenizer = await models.getTaskTokenizer(t);
+      const maxLength =
+        t.trainingInformation.maxSequenceLength ??
+        (tokenizer.model_max_length as number) + 1;
+
+      return d
+        .map((line) =>
+          processing.tokenizeAndLeftPad(line, tokenizer, maxLength),
+        )
+        .map((tokens) => tokens.pop());
```
JulienVig: If we don't want the label can we remove the + 1  instead of including an extra token to pop it later?

discojs/src/processing/index.ts
```diff
@@ -0,0 +1,172 @@
...
+    }
+  }
+}
+
+function extractToNumbers(columns: Iterable<string>, row: Tabular) {
+  return (
+    List(columns)
+      .map((column) => processing.extractColumn(row, column))
+      // TODO sanitization doesn't care about column distribution
+      .map((v) => (v !== "" ? v : "0"))
```
JulienVig, line 162: I chose a constant value as a temporary fix for the training failures (#615, #649) but now that the preprocessing works differently we may be able to handle it more gracefully. 
IMO choosing how to handle missing values should be the responsibility of the data owner so we could throw an error or drop the whole row as default behavior.

discojs/src/processing/text.ts
```diff
@@ -0,0 +1,47 @@
...
+    max_length: length,
+  });
+
+  if (
+    typeof tokenized !== "object" ||
+    tokenized === null ||
+    !("input_ids" in tokenized) ||
+    !isArrayOfNumber(tokenized.input_ids)
+  )
+    throw new Error("tokenizer returns unexcepted type");
```
JulienVig:
```suggestion
    throw new Error("tokenizer returns unexpected type");
```
If it's indeed a typo it might be worth doing a ctrl+f for "excepted" :)

discojs/src/dataset/dataset.ts
```diff
@@ -57,9 +62,7 @@ export class Dataset<T> implements AsyncIterable<T> {
   chain(other: Dataset<T> | DatasetLike<T>): Dataset<T> {
     if (!(other instanceof Dataset)) other = new Dataset(other);
 
-    const self = {
-      [Symbol.asyncIterator]: () => this.#content(),
-    };
+    const self = { [Symbol.asyncIterator]: () => this[Symbol.asyncIterator]() };
```
JulienVig: Do you know why we have to create another object rather than use `this`? (aside from typescript not being happy)

discojs/src/dataset/dataset.ts
```diff
@@ -42,7 +47,7 @@ export class Dataset<T> implements AsyncIterable<T> {
    */
   map<U>(mapper: (_: T) => U | Promise<U>): Dataset<U> {
     const content = {
-      [Symbol.asyncIterator]: () => this.#content(),
+      [Symbol.asyncIterator]: () => this[Symbol.asyncIterator](),
```
JulienVig: I don't think I understand why this is necessary to "allow subclass"

discojs/src/dataset/dataset.ts
```diff
@@ -174,4 +182,55 @@ export class Dataset<T> implements AsyncIterable<T> {
...
+  /** Try to keep generated elements to avoid recomputing
+   *
+   * Drops everything when memory pressure is applied.
+   */
+  cached(): Dataset<T> {
+    return new CachingDataset(this.#content);
+  }
+}
+
+class CachingDataset<T> extends Dataset<T> {
```
JulienVig, line 204: Could you add more documentation on the expected behavior? (for people like me who can't think in typescript fluently yet)
For example:
* What is being cached? Is it trying to cache the whole dataset for future epochs? can it cache the dataset partially or is it everything or nothing?
* how is the `done` boolean used? We're reading from the cache when `done` is true, I was expecting the opposite. It seems "done" means "ready to use" rather than "finished iterating" as for iterators

discojs/src/dataset/dataset.ts
```diff
@@ -174,4 +182,55 @@ export class Dataset<T> implements AsyncIterable<T> {
...
+      // eslint-disable-next-line @typescript-eslint/require-await
+      return (async function* () {
+        yield* cached[1];
+      })();
+    }
+
+    debug("cache invalid, recomputing");
+
+    this.#cache = new WeakRef([false, List()]);
+    const cache = this.#cache;
```
JulienVig: is this needed because we don't have access to `this` in the returned generator?

cli/src/benchmark_gpt.ts
```diff
@@ -90,32 +81,10 @@ async function main(args: Required<CLIArguments>): Promise<void> {
...
-                )
-                .toArray(),
-            ) as tf.Tensor3D,
-          })),
-        ),
-    );
+    const preprocessedDataset = dataset
+      .map((line) => processing.tokenizeAndLeftPad(line, tokenizer, maxLength))
+      .map((tokens) => [tokens.pop(), tokens.shift()] as [List<number>, List<number>])
+      .batch(batchSize);
```
JulienVig: Verify satisfying changes

discojs/src/types.ts
```diff
@@ -1,11 +1,38 @@
...
 
-export type TypedLabeledDataset =
-  | ["image", Dataset<[Image, label: string]>]
-  | ["tabular", Dataset<Tabular>]
-  | ["text", Dataset<Text>];
+/**
+ * The data that we handle goes through various stages.
+ * The labels also gets transformed at each stage.
+ *
+ * raw -> model encoded -> inferred
```
JulienVig: I'm not sure what ModelEncoded represents, from the name I expected it to be similar to a serialized model type (there is the EncodedModel that does exactly that) but it seems to be representing a dataset ready to be fed to a model, is this right? 
What about changing the names to RawDataset -> ModelInputDataset -> ModelOutputDataset?

discojs/src/models/gpt/index.ts
```diff
@@ -45,38 +50,31 @@ export class GPT extends Model {
...
       const batchLogs = await this.#runBatch(batch);
-      tf.dispose(batch);
 
       yield batchLogs;
       batchesLogs = batchesLogs.push(batchLogs);
     }
 
-    const validation = validationData && (await this.#evaluate(validationData));
+    const validation =
+      validationDataset && (await this.#evaluate(validationDataset));
```
JulienVig, line 75: Does this check that validationDataset is defined and return the evaluation if so?

JulienVig: If `validationDataset` is defined why does `Dataset && Record` resolves to `Record`?

discojs/src/models/gpt/index.ts
```diff
@@ -182,3 +200,12 @@ export class GPT extends Model {
...
   }
 }
+
+function intoTFDataset<T extends tf.TensorContainer>(
+  iter: AsyncIterable<T>,
+): tf.data.Dataset<T> {
+  // @ts-expect-error generator
+  return tf.data.generator(async function* () {
+    yield* iter;
+  });
```
JulienVig: Why is typescript raising an error here? Is it the probalem with tfjs not exporting their Iterator type?

discojs/src/models/gpt/index.ts
JulienVig: Aah got it: "the && operator returns the value of the first falsy operand encountered when evaluating from left to right, or the value of the last operand if they are all truthy."

discojs/src/models/tfjs.ts
```diff
@@ -164,4 +214,123 @@ export class TFJS extends Model {
...
+            b.map(([line]) => tf.tensor1d(line.toArray())).toArray(),
+          ),
+          ys: tf.stack(
+            b
+              .map(([line, next]) =>
+                tf.oneHot(line.shift().push(next).toArray(), outputSize),
+              )
+              .toArray(),
+          ),
+        };
```
JulienVig: I don't think we're ever using tfjs with text datasets right?

discojs/src/training/disco.ts
```diff
@@ -5,55 +5,70 @@ import {
...
 import { RoundLogs, Trainer } from "./trainer.js";
-import { labeledDatasetToDataSplit } from "../dataset/data/helpers.js";
 
 interface DiscoConfig {
-  scheme: TrainingInformation["scheme"];
+  scheme: TrainingInformation<DataType>["scheme"];
   logger: Logger;
+
+  // keep preprocessed dataset in memory while training
+  preprocessOnce: boolean;
```
JulienVig: It's not clear to me what this does from this comment and from the code, could you add more explanations?
I expected this to toggle the dataset caching or something. 
Does this apply the preprocessing over the whole dataset at once rather than doing it on the fly?

discojs/src/processing/index.ts
```diff
@@ -0,0 +1,172 @@
...
+      return d
+        .map((line) =>
+          processing.tokenizeAndLeftPad(line, tokenizer, maxLength),
+        )
+        .map((tokens) => tokens.pop());
+    }
+  }
+}
+
+export async function postprocess<D extends DataType>(
```
JulienVig, line 120: Could you add a docstring explaining what post-processing is doing for the different data types?

discojs/src/models/gpt/index.ts
```diff
@@ -122,17 +112,45 @@ export class GPT extends Model {
...
+
+  override async predict(
+    batch: Batched<ModelEncoded["text"][0]>,
+  ): Promise<Batched<ModelEncoded["text"][1]>> {
+    const predictNext = async (tokens: List<number>) => {
+      const generated = await this.model.generate(tokens.toArray(), {
+        maxNewTokens: 1,
+        temperature: 1.0,
+        doSample: false,
+      });
```
JulienVig: Predict doesn't allow choosing the generation parameters, is it intended?

discojs/src/validator.ts
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  /** infer every line of the dataset and check that it is as labelled */
+  async *test(dataset: Dataset<Raw[D]>): AsyncGenerator<boolean, void> {
+    const results = (await processing.preprocess(this.task, dataset))
+      .batch(this.task.trainingInformation.batchSize)
+      .map(async (batch) =>
+        (await this.#model.predict(batch.map(([inputs, _]) => inputs)))
+          .zip(batch.map(([_, outputs]) => outputs))
+          .map(([infered, truth]) => infered === truth),
```
JulienVig:
```suggestion
          .map(([inferred, truth]) => inferred === truth),
```

cli/src/args.ts
```diff
@@ -67,6 +67,6 @@ export const args: BenchmarkArguments = {
 
       return task;
     },
-    getModel: provider.getModel,
+    getModel: () => provider.getModel(),
```
JulienVig, line 70: Is this better than just writing the function name?

webapp/cypress/support/e2e.ts
JulienVig, line 1: My IDE is not happy with Task and TaskProvider's genericity, I'm not sure why nothing fails though. It seems that you removed the default DataType type 

Screenshot 2024-09-26 at 17 43 15

JulienVig (changes requested):
Huge amount of work! You seem to be pushing typescript to its limits haha

I have an error when trying to build the webapp:
```
Error: Cannot find module @rollup/rollup-darwin-arm64. npm has a bug related to optional dependencies (https://github.com/npm/cli/issues/4828). Please try 'npm i again after removing both package-lock.json and node_modules directory.
at requireWithFriendlyError (/Users/vignoud/Disco/review/node_modules/rollup
/dist/native.js: 59:9)
at Object. ‹anonymous> (/Users/vignoud/Disco/review/node_modules/rollup/dist/
native. js: 68:76)
•. 3 lines matching cause stack trace ...
at Module._load (node:internal/modules/cjs/loader:1023:12) at cjsLoader (node:internal/modules/esm/translators:356:17)
at ModuleWrap. ‹anonymous> (node:internal/modules/esm/translators:305:7) at ModuleJob. run (node:internal/modules/esm/module_job:218:25)
at async ModuleLoader. import (node:internal/modules/esm/loader:329:24) {
  [cause]: Error: Cannot find module '@rollup/rollup-darwin-arm64'
  Require stack:
    - /Users/vignoud/Disco/review/node_modules/rollup/dist/native.js
    at Module._resolveFilename (node:internal/modules/cjs/loader:1144:15)
    at Module._load (node:internal/modules/cjs/loader:985:27) at Module. require (node:internal/modules/cjs/loader:1235:19)
    at require (node:internal/modules/helpers:176:18)
    at requireWithFriendlyError (/Users/vignoud/Disco/review/node_modules/roll up/dist/native.js:41:10)
    at Object. ‹anonymous> (/Users/vignoud/Disco/review/node_modules/rollup/dis
    t/native. js: 68:76)
    at Module._compile (node:internal/modules/cjs/loader:1376:14)
    at Module._extensions..js (node: internal/modules/cjs/Loader:1435:10) at Module. load (node:internal/modules/cjs/loader:1207:32)
    at Module._load (node:internal/modules/cjs/loader: 1023:12) {
    code: 'MODULE_NOT_FOUND' requireStack: [
    '/Users/vignoud/Disco/review/node_modules/rollup/dist/native.js'
    ]
  }
｝
```
I didn't manage to solve the issue even by re-creating the package-lock.json (it actually solves this error but then I can't build the server anymore because of new errors)

I couldn't review everything yet because of this but I already left a lot of questions (nothing blocking except the ropllup error so far)
If you can't solve the rollup error without having access to a Mac we can organize a call next week if you want

webapp/cypress/support/e2e.ts
JulienVig: This may be an error that arose after re-creating the package-lock.json

server/tests/e2e/federated.spec.ts
tharvik: actually, I prefer the latter but [eslint was bothering me](https://eslint.org/docs/v8.x/rules/no-constant-condition). look like the rule wasn't very well done and is changed is the newer eslint 9.

discojs/src/processing/image.ts
tharvik: right on!
it's based on the image format I know, I only added RGB (depth 3) & RGBA (depth 4). in a previous PR, you told me that grayscale (depth 1) was common (I though it was antique and not used anymore). but I never heard of a format with only two channel, but if we stumble on it at some point, we can add support for it

discojs/src/processing/image.ts
tharvik: added a paragraph explaining the layout.
in image manipulation/format, it is very common to represent images this way (interleaved RGBA). indeed 1D array are a tad more efficient, but the most enjoyable is the `Uint8Array` feature, that ensure that only a byte is used per value (reduced memory footprint) and that each value is integral in range 0..255, which is the only range of value valid for RGB(A).

discojs/src/processing/image.ts
tharvik: correct, it's the [function overloads](https://www.typescriptlang.org/docs/handbook/2/functions.html#function-overloads) syntax of typescript. it allows to specify more precisely how types get transformed by a function. in this case:
* `Image<4> -> Image<3>`
* `Image<1 | 3 | 4> -> Image<1 | 3>` (more exactly, it'll drop the 4 from the type)

and the last one is the actual implementation, which should behave how the overload are stating (but it is not actually enforced :roll_eyes:).

it couldn't have been done using a single function as otherwise, the type of the inputed image would get widden without reason. for example, `Image<3> -> Image<1 | 3>` (no alpha channel should keep the depth) & `Image<4> -> Image<1 | 3>` (but we know it's 3). it's a bit magic I agree :sparkles: but makes it very nice for the users.

discojs/src/processing/index.ts
tharvik: added a clarification by using a better variable name (`totalTokenCount`).
in general (and that's quite opiniated), I prefer to change code to be more readable than using comment to explain what is happening. using one-off function, clearly separating logical blocks and descriptive variable name (where it matter) are all means to have a closer-to-english language than a cryptic one.
sadly, it doesn't works in all cases, especially if the implemented algorithm is convoluted, but even there, one should strive to make it the most readable, and have a general overview of the logic at the top of the file.

discojs/src/processing/index.ts
tharvik: hum, it's how I represent myself the types when going in this scope: "D" of type `'image' | 'tabular' | 'text'` is reduced to `'image'` in this case.
in the documentation, [narrowing is used](https://www.typescriptlang.org/docs/handbook/2/narrowing.html), would it be clearer?

discojs/src/processing/index.ts
tharvik: ho yes, good catch!

discojs/src/processing/index.ts
tharvik: good point, thanks for linking the issue; followup on #649

discojs/src/processing/text.ts
tharvik: haaa, I'm always make this type, it's so close!
8 corrected all around

discojs/src/dataset/dataset.ts
tharvik: [`this` is a nightmare](https://github.com/Microsoft/TypeScript/wiki/'this'-in-TypeScript#what-is-this-in-javascript). to make it short, `this` is defined by how it is called rather than where it is scoped (allow for `element.onclick = function () { this /* of type HTMLElement */ }` but makes it quite hard to reason when implement smth)

in here, it's because `function` actually defines a new scope so a new `this` (which isn't the case with arrow function for example). so we need to keep the `this` of `Dataset` around instead of using the one of the function. and that's the way I found to do so at the time (it's also very explicit, which helps with `this` handling).

that being told, now that I read a bit more on it, I changed it up to append `bind(this)` to the function to forward `Dataset`'s `this`. it does look nicer (wasn't able to pass `super` in `CachingDataset` tough, that would be magic for another day)

discojs/src/dataset/dataset.ts
tharvik: by using a method call, it will defer to the actual implementation of `Symbol.asyncIterator`. as it gets overriden in `CachingDataset`, we can now mix multiple implementation of the class and call the correct method. (note also that in `CachingDataset`, `this.#content` isn't valid because it's private to the parent class)

discojs/src/dataset/dataset.ts
tharvik: right, I used a bit of JS magic there (it's my first time using WeakRef outside of a theory course :eyes:), added a few lines to explain what its purpose.

I do recommand the [wikipedia page](https://en.wikipedia.org/wiki/Weak_reference) the general idea and the [javascript.info page on the matter](https://javascript.info/weakmap-weakset) for more code.

> What is being cached? Is it trying to cache the whole dataset for future epochs?

all elements, it builds a List of the generated parent elements and keep it around (if not evicted). of course, if the cached dataset is bigger than memory, it serves not purpose (but at least, it doesn't crash)

> can it cache the dataset partially or is it everything or nothing?

it is everything or nothing. the issue is that the iterator is only sequential, so even if we had some partial cache, we would still need to advance the iterator (thus computing the values) until we get to the uncached values.

this way of implementing dataset while very easy to use might be revised, especially if we want to compute data in parallel.

> how is the done boolean used?

ha yes, that not really obvious, "done" is quite an unprecise term: it is as you pointed out used as "finished iterating". I renamed it to "filled" and added a comment explaining its (tri)state.

discojs/src/dataset/dataset.ts
tharvik: correct. but it's not needed anymore as I'm now binding `this` in the async generator function!

cli/src/benchmark_gpt.ts
tharvik: héhé thanks! I'm kinda proud of how readable it turned out :sparkles:

discojs/src/types.ts
tharvik: > but it seems to be representing a dataset ready to be fed to a model, is this right?

yeah, pretty much. `ModelEncoded` is here both the input and the output of the model: the two are needed for training, `predict` takes the first one and return the second, validator might only need the input, …

> What about changing the names to RawDataset -> ModelInputDataset -> ModelOutputDataset?

I was using previously smthDataset, but it began exploding when dealing with Batched and/or WithLabel; also, it is not a Dataset in itself but a type of data _inside_ a Dataset.
instead of having it as a ~root type, I can push it into a module, as to have `data.Raw`, `data.ModelEncoded`, … I don't like the name "data" as it is quite undescriptive, maybe you've a better one?

discojs/src/models/gpt/index.ts
tharvik: ha yes, I find it closely related to `??` in my mind, it's a bit magic but remember me so much of POSIX sh (which was my first language, after [Logo](https://en.wikipedia.org/wiki/Logo_(programming_language)))

discojs/src/models/gpt/index.ts
tharvik: `tf.data.generator` except a `() => Iterator<T>| Promise<Iterator<T>>` but we pass it `() => AsyncIterator<T>` (akin to an iterator of promises). eventough it is in fact supported internally.
I've openned [an upstream PR](https://github.com/tensorflow/tfjs/pull/8408) to fix that.

so not really their internal iterator, rather, they typed something in a restrictive fashion.

discojs/src/models/tfjs.ts
tharvik: hoo indeed, indeed! dropping text support for TFJS.

btw, very comfortable change to implement, as types gets refined, one can simply change `class TFJS<D extends DataType>` to `class TFJS<D extends "image" | "tabular">` and works their way through the errors rather than letting it crash somewhere down the road.

discojs/src/processing/index.ts
tharvik: hum, it seems that I'm not even using `postprocess`; well, I update the validator to now use it, and to output inferred values rather than numbers.
also added a few lines to `types.Inferred` to explain more what each values should be

discojs/src/models/gpt/index.ts
tharvik: well, that's a good point. I'm trying to think of a generic way to allow predict to do that.
* having it as the `Model` level doesn't work as every implementation might have a different config (temperature for 'tabular' models?)
  * note: if we find a config usable across a given DataType, we can have that instead
* having it as the implementation (GPT/TFJS) level looks right, with an extra argument to `predict` which is only valid after a `instanceof` check

I'm having an issue with returning many tokens from GPT, as that would break the types (a single token is used as output in `ModelEncoded`). I would argue that it's weird to ask for many tokens. IMO, in a loop, one should ask for the next one and append it to the input (as it is already done in `GPTForCausalLM.generate`).

after losing myself in code, I choose to only have a single token generated, and to expose the config in GPT direclty. that put a bit a burden on the users to pre/post-process the lines, but `Model` isn't made for users.
at some later point, I was thinking of merging `Validator` into `Disco` (as to have a single entry-point), and thus to provide a nice predict method there, with a selectable number of token.

tl;dr no, it wasn't intended, put it back but nicer

cli/src/args.ts
tharvik: I wanted the first one to be working as I found it nicer, but [eslint didn't want me to](https://typescript-eslint.io/rules/unbound-method) as `this` can be unexpectly bound (not an issue in this case but the fix is simple and more explicit)

webapp/cypress/support/e2e.ts
tharvik: ha yes, we discussed it last time but for anyone stumbling on that:
cypress's support files weren't typedcheck in the CI (but is with LSP in IDEs). added a project reference in webapp's tsconfig to ensure that, and oc fixing the actual error.

tharvik:
> You seem to be pushing typescript to its limits haha

héhé, yeah, it's what I do in pretty much any coding languages, pushing it to its limits then complaining about it (:
 
> I couldn't review everything yet because of this but I already left a lot of questions

yeah,I think that the more we progress, the long the review will be, as we both know more of the code, what can get wrong, and are getting opiniated on various components. I feel quite good about it in fact, that means we care more for disco :mirror_ball: 

> If you can't solve the rollup error without having access to a Mac we can organize a call next week if you want

fwiw, it was indeed a mac specific issue, that gets only triggered there, and will probably happen again if there is an update of rollup that I do on my side. well, now we know

discojs/src/processing/index.ts
JulienVig: Now that I understand it makes sense, both reduce or narrow are okay with me!

discojs/src/models/tfjs.ts
JulienVig: Haha nice reaping the benefits of your work!

discojs/src/types.ts
JulienVig: Maybe rename data to DatasetFormat? and have Raw, Training, Prediction and Validator formats?

discojs/src/training/disco.ts
tharvik: `Dataset.cached` is kinda free for us to use, as it is quite fragile (will drop everything if asked for), so I'm using it unconditionnaly. but then tests runtime were quite unpredictable, as sometime it was recomputing, sometime not. so adding this switch allowed to keep the preprocessed dataset around (and incidently, also preprocessing before training).

made the comment clearer

discojs/src/types.ts
tharvik: the issue with splitting `ModelEncoded` into many types is that it decouple the link between input & output of a model.

* `Model.train` needs a dataset with both input and output
* `Model.predict` takes an (batched) input and returns a (batched) output

otherwise, I would need a common type to link the two and enforce that the same type which is used in training is also the one in predict.

note: also, I hope to get rid of the specifics for Validator at some later point, so hopefully merging `RawWithoutLabel` & `Inferred` into `Raw` (and of course, theses types will probably evolve, I already got ideas to make tabular task safer!)

anyway, moving everything into a `DataFormat` type, but keeping the "old" names

discojs/src/types/data_format.ts
```diff
@@ -0,0 +1,32 @@
+import { List } from "immutable";
+
+import type { Image, processing, Tabular, Text } from "../index.js";
+
+/** what get's ingested by Disco */
```
JulienVig:
```suggestion
/** what gets ingested by Disco */
```
Same for the other comments

discojs/src/types/data_format.ts
```diff
@@ -0,0 +1,32 @@
...
+}
+/** what get's ingested by the Validator */
+export interface RawWithoutLabel {
+  image: Image;
+  tabular: Tabular;
+  text: Text;
+}
+
+type Token = number;
+export interface ModelEncoded {
```
JulienVig, line 33: Can you add something like "Raw is processed into ModelEncoded. Data format expected to train models"

discojs/src/dataset/dataset.ts
```diff
@@ -174,4 +185,62 @@ export class Dataset<T> implements AsyncIterable<T> {
...
+    if (cached !== undefined && cached[0]) {
+      debug("valid cache, reading from it");
+
+      // eslint-disable-next-line @typescript-eslint/require-await
+      return (async function* () {
+        yield* cached[1];
+      })();
+    }
+
+    debug("cache invalid, recomputing");
```
JulienVig:
```suggestion
    debug("cache invalid, reading from dataset");
```
Nitpick: can we tweak this debug statement a bit? I found it a bit misleading when it's the first time calling the dataset to see "recomputing" as if it had already been computed before but failed.

discojs/src/models/tfjs.ts
```diff
@@ -30,69 +39,58 @@ export class TFJS extends Model {
...
 
   async #runBatch(
-    batch: tf.TensorContainer,
+    batch: Batched<DataTypeToPreprocessedLabeled[D]>,
   ): Promise<Omit<BatchLogs, "batch">> {
-    let logs: tf.Logs | undefined;
-    await this.model.fitDataset(tf.data.array([batch]), {
+    const { xs, ys } = this.#batchToTF(batch);
+
+    const { history } = await this.model.fit(xs, ys, {
```
JulienVig, line 67: Is there a reason why we're not use fit rather than fitDataset for GPT? (I rather dislike using a method called fitDataset when it's in fact just a batch)

server/src/task_set.ts
```diff
@@ -64,13 +69,10 @@ export class TaskSet extends EventEmitter<{
...
       encodedModel = model // don't do anything if already encoded
     } else {
-      let tfModel: Model
+      let tfModel: Model<DataType>
       if (model === undefined) { 
         // Get the model if nothing is provided
         tfModel = await this.loadModelFromTask(taskOrProvider)
-      } else if (model instanceof URL) {
-        // Downloading the model if a URL is given
-        tfModel = new models.TFJS(await tf.loadLayersModel(model.href))
```
JulienVig, line 74: Is this never used? We have some tasks using remote models, but they maybe load the models in task.getModel

JulienVig (approved):
Amazing refactor!!

* Mainly, I noticed that testing an image model always shows the images as red (while the accuracy is 100% in this case). That can be a simple UI mistake but maybe a deeper problem in the validation.
Screenshot 2024-10-24 at 14 02 43
* Not sure if this was introduced in this PR or the dark theme but when there is only one datapoint the chart now shows a bar instead of a dot
Screenshot 2024-10-24 at 11 06 07

* Also the x-axis labels are getting crowded
Screenshot 2024-10-24 at 14 01 09

* Testing with a language model shows a "Download CSV" button that doesn't work 
Screenshot 2024-10-24 at 11 08 20

discojs/src/models/tfjs.ts
tharvik: because GPT is misusing TFJS quite a lot :face_with_diagonal_mouth: instead of reusing TFJS's training loop, it is redoing it, by overriding `fitDataset` (and forgetting to also implement `fit`)… I don't think there is any real reason.
if we want to put a bunch of hours on fixing it, we can probably get rid of it entirely (it remaind of smth I tried to do a while back) but I'm not sure that the gain will be enormous, especially when ONNX would be better,faster,stronger

server/src/task_set.ts
tharvik: yeah, we don't have a model URL in `Task` anymore, it was move in `getModel` some time back (and in fact, one can quick easily have the old behavior by making a `TaskProvider` where it will `tf.loadLayersModel`)

discojs/src/types/data_format.ts
tharvik: yeah, that's not obvious, I'm in it too deep!
added a few comments to clarify the relation w/ processing

tharvik:
> * Mainly, I noticed that testing an image model always shows the images as red (while the accuracy is 100% in this case). That can be a simple UI mistake but maybe a deeper problem in the validation.

ouf, it was a simple UI mistake, I inversed red & green in #682

> * Not sure if this was introduced in this PR or the dark theme but when there is only one datapoint the chart now shows a bar instead of a dot
> * Also the x-axis labels are getting crowded

ha yes, good point, I reworked it to make it a bit nicer. the curve is now "cubic" instead of "smooth" which makes it less jittery at the expense of precision.

> * Testing with a language model shows a "Download CSV" button that doesn't work

right, it is not implemented. I wasn't sure of what to show in it so I prefered to throw a TODO. now, do we want to even have this feature? I feel that it is not really useful (for testing that is, for prediction it is useful). we need a way to present testing results but CSV is too technical.
* images: we already have a nice way to show the results
* tabular: yes, not really nice but at least it is somewhat readable (colors would be nicer)
* text: I can think of way to present it: for every input line (styled as gray), show the next word in green/red

WDYT? (for a next iteration)

PR #796, opened 2024-09-30
*: rm slack

tharvik:
as per our slack guru (@Annie-LiGHT), we don't want to have a direct join link for random users.


JulienVig (approved):
All good!!

Issue #800, opened 2024-10-07
GPT-tfjs model is not persisted in the browser local storage


JulienVig:
Download the GPT model in the Evaluation page doesn't persist the model in the browser's cache: 
- refreshing the page makes it disappear.
- trying to use the model after downloading fails
Inspecting the page, we can see that the local storage (or indexedDB) doesn't show any new model when downloading the GPT model, while it does for others. This happens both in local and for discolab.ai
@tharvik do you have an idea of what's happening? I suspect it may come from #764

tharvik:
after some investigation, it seems that a serialized GPT model is 40MiB which is [way larger than how much we can put in local storage](https://developer.mozilla.org/en-US/docs/Web/API/Storage_API/Storage_quotas_and_eviction_criteria#web_storage). sadly, moving to using a larger storage requires to use async storage which [isn't support by the persist layer we use now](https://prazdevs.github.io/pinia-plugin-persistedstate/guide/limitations.html#storage-must-be-synchronous).

I'm moving to use another persist layer, [`pinia-plugin-persistedstate-2`](https://www.npmjs.com/package/pinia-plugin-persistedstate-2), which allow for async storage. and implementing an [OPFS](https://developer.mozilla.org/en-US/docs/Web/API/File_System_API/Origin_private_file_system) interface.

JulienVig:
I see, not very surprising. This is something we will have to care about as language models grow larger. What about IndexedDB?

tharvik:
IndexedDB is also a possibility. if our case, we don't need to search for data, so OPFS looked more fit.

PR #803, opened 2024-10-08
webapp: fix GPT storage

tharvik:
as outlined in #800, the storage space of the browser does break when huge models (such as GPT) are stored.

* change the persist backend to support larger stores as [the old one doesn't support theses](https://prazdevs.github.io/pinia-plugin-persistedstate/guide/limitations.html#storage-must-be-synchronous)
* add a storage backend based on [OPFS](https://developer.mozilla.org/en-US/docs/Web/API/File_System_API/Origin_private_file_system)
* a smaller storage backend based on LocalStorage, to still work if OPFS isn't supported (such as in the tests)
  * stores the smaller models until capacity, where it discards the large ones
* also add a layer to check that deserialized values are valid
* also reduce the serialized size of weights & models by using `UInt8Array` instead of array of numbers
* change the msgpack library to a one not eating up all memory (and recent, and standard)
* update server tests to wait for closing the server, and reduce the number of loaded task per test


tharvik:
finally managed to find a way to avoid clogging up memory: don't trust `msgpack-lite` to do the job correctly and instead use the official JS library

JulienVig (commented):
Thanks for the great work!

Using Chrome or Firefox (131) smaller models are saved with LocalStorage but GPT never seems to be persisted (refreshing clears it), is there anything to do to enable OPFS?
Edit: that was my mistake

discojs/src/client/client.ts
```diff
@@ -172,8 +171,9 @@ export abstract class Client extends EventEmitter<{'status': RoundStatus}>{
     }
     url.pathname += `tasks/${this.task.id}/model.json`
 
-    const response = await axios.get<ArrayBuffer>(url.href, { responseType: 'arraybuffer' })
-    return await serialization.model.decode(new Uint8Array(response.data))
+    const response = await fetch(url);
+    const encoded = new Uint8Array(await response.arrayBuffer())
+    return await serialization.model.decode(encoded)
```
JulienVig, line 176: What was the problem with axios?

discojs/src/index.ts
```diff
@@ -1,7 +1,5 @@
 export * as data from './dataset/index.js'
 export * as serialization from './serialization/index.js'
-export { Encoded as EncodedModel } from './serialization/model.js'
-export { Encoded as EncodedWeights } from './serialization/weights.js'
```
JulienVig, line 4: I liked having the explicit difference between serialized weights and model, is this not relevant anymore? Is there no difference when `serialization.Encoded` is a model or weights?

discojs/src/serialization/coder.ts
```diff
@@ -0,0 +1,51 @@
...
+  // `Buffer.slice` (subclass of Uint8Array on Node) doesn't copy
+  // thus doesn't respect Liskov substitution principle
+  // https://nodejs.org/api/buffer.html#bufslicestart-end
+  // here we call the correct implementation
+  return Uint8Array.prototype.slice.call(arr);
+}
+
+const CODEC = new msgpack.ExtensionCodec();
+CODEC.register({
+  type: 0x17,
```
JulienVig, line 29: Can you add a comment explaining what are these two extensions and why you chose these values (0x17 and 0x1a) for the extension types?
From what I understood Float32Array is the weights format, and is ArrayBuffer for the fetch response?

webapp/package.json
```diff
@@ -13,12 +13,13 @@
...
     "@epfml/discojs": "*",
     "@epfml/discojs-web": "*",
+    "@msgpack/msgpack": "^3.0.0-beta2",
     "apexcharts": "3",
     "cypress": "13",
     "d3": "7",
     "immutable": "4",
     "pinia": "2",
-    "pinia-plugin-persistedstate": "3",
+    "pinia-plugin-persistedstate-2": "2",
```
JulienVig, line 22: Why switch to pinia-2? It has 100x less users than the original

webapp/src/store/models/local_storage.ts
```diff
@@ -0,0 +1,131 @@
...
+  // https://developer.mozilla.org/en-US/docs/Web/API/Storage_API/Storage_quotas_and_eviction_criteria#web_storage
+  static readonly #MAX_SIZE = 5 * 1024 * 1024;
+
+  readonly storage = localStorage;
+
+  serialize(state: State): string {
+    let size = 0;
+    const keptModels = state.idToModel
+      .sortBy((infos) => infos.encoded.length)
+      .takeWhile((infos) => {
```
JulienVig, line 23: Very interesting, didn't know about `takeWhile`'s existence

webapp/src/store/models/local_storage.ts
```diff
@@ -0,0 +1,131 @@
...
+    encoded,
+    dateSaved,
+  }: Partial<Record<keyof LocalStorage.Serialized[0][1], unknown>> = raw;
+
+  if (
+    typeof taskID !== "string" ||
+    !(
+      typeof encoded === "string" &&
+      encoded.length % 2 === 0 &&
+      encoded.match(/[^0-9a-f]/) === null
```
JulienVig: Can you add a comment describing what is the expected format that the regex is testing?
The encoding should not start with a digit or a lowercase letter between a and f? So it's expecting a capital letter or a lower case letter starting from g?

webapp/src/components/testing/TestSteps.vue
JulienVig, line 1: With regard to this file, I noticed that testing using GPT in the webapp showed a success toaster "Model testing successfully completed" right after starting the test while the number of visited sample was still increasing so obviously the testing had not completed.
There is an `await` missing l.284 before `startTextTest(toRaw(dataset.value)[1]);`

JulienVig (approved):
Great work!! I left mostly questions and pleas for more comments in the code, nothing blocking :)

discojs/src/client/client.ts
tharvik: there isn't a fundamental problem, it happened when trying to find what was using up all the memory.
also, `fetch` is implement by all platform now ([took nodejs a while to do so](https://github.com/nodejs/node/blob/main/doc/changelogs/CHANGELOG_V18.md#fetch-experimental)) so axios is not really needed anymore and less deps is nice.

discojs/src/index.ts
tharvik: correct, everything is `Uint8Array` now. previously, it was kinda the case, one being `number[]` (but with only 0-255 values), the other `Uint8Array`. I can reintroduce the two if you prefer but a `isEncoded{Model,Weights}` function can't as it would require actually decoding the whole array to differenciate between Model & Weights.

I actually kept the `EncodedModel` in server/task_set to make it more explicit, which can help where the meaning of "Encoded" isn't obvious.

discojs/src/serialization/coder.ts
martinjaggi: yes would be good to clarify in the code which models become int8 (only LLMs i guess?) or also others? because classic ML ones might not work very well with int8. LLMs can be ok

tharvik: ho right, that's quite cryptic indeed. added a few comments explaining the CODEC

the values are taken from msgpack-lite for no special reason apart that I needed random value.
Float32Array are for the weights, ArrayBuffer are for the TFJS's saved model

webapp/package.json
tharvik: yeah, that bothers me too, but the original [doesn't support async/large storage]((https://prazdevs.github.io/pinia-plugin-persistedstate/guide/limitations.html#storage-must-be-synchronous))

added a comment in webapp/main.ts to explain that choice.

webapp/src/store/models/local_storage.ts
tharvik: functional on iterators all the way :rocket:

webapp/src/store/models/local_storage.ts
tharvik: ha yes, regex are an acquired taste! here, it checks that no character outside of the hex (`[0-9a-f]`) range is found in `encoded`, to ensure that it is indeed a valid base16 encoding.

added a comment

webapp/src/components/testing/TestSteps.vue
tharvik: ho nice, good catch! added a commit for it

discojs/src/serialization/coder.ts
tharvik: to answer @martinjaggi: here it's simply the serialization we use to communicate weights & models on the network, not the types used for the tensors inside the models. theses are float32 (because WeightsContainer maps everything to float32, which should be changed at some point)
now that you mention it, if the LLM model can work with int8, we should be able to divide by four its serialized size, which is a very nice gain!

JulienVig: We've been assuming that JS could only support 32 bits but can we actually train a model on int8?

tharvik: I think so, at least, there is support in TFJS for [various type of tensor](https://github.com/tensorflow/tfjs/blob/master/tfjs-core/src/types.ts#L40)

PR #807, opened 2024-10-16
Fix and rework GPT-TF.js

### LINKED ISSUE #654: Improve and rework GPT-tfjs

JulienVig:
Closes #654 

* Fix weight initialization from zero to random uniform -> Fixes GPT2 NaN loss
* Implement weight sharing between token embeddings and the language modeling head -> -20% memory footprint
* Improve generation with top k sampling option
* Add seed for deterministic runs
* Implement text loaders by byte chunk rather than by line which doesn't require to pad each line to the context length
* Waiting on Transformers.js [#984](https://github.com/huggingface/transformers.js/issues/984) and [#1019](https://github.com/huggingface/transformers.js/issues/1019#issuecomment-2465803843) to migrate to @huggingface/transformers v3.

### DIFF
--- a/cli/src/benchmark_gpt.ts
+++ b/cli/src/benchmark_gpt.ts
@@ -79,7 +79,7 @@ async function main(args: Required<CLIArguments>): Promise<void> {
     task.trainingInformation.maxSequenceLength = contextLength
     const dataset = loadText('../datasets/wikitext/wiki.train.tokens')
       .map(text => processing.tokenize(tokenizer, text))
-      .unbatch()
+      .flatten()
       .batch(config.blockSize + 1, 1)
 
     const preprocessedDataset = dataset
--- a/cli/src/benchmark_gpt.ts
+++ b/cli/src/benchmark_gpt.ts
@@ -107,9 +107,9 @@ async function main(args: Required<CLIArguments>): Promise<void> {
     
     // Benchmark parameters
     const prompt = 'The game began development in 2010 , carrying over a large portion, The game began development in 2010 , carrying over a large portion, The game began development in 2010 , carrying over a large portion,'
-    const nbNewTokens = 200
+    const maxNewTokens = 200
     const iterations = 10
-    console.log("Generating", nbNewTokens, "new tokens")
+    console.log("Generating", maxNewTokens, "new tokens")
 
     let tokens = List(
       (tokenizer(prompt, { return_tensor: false }) as { input_ids: number[] })
@@ -119,13 +119,13 @@ async function main(args: Required<CLIArguments>): Promise<void> {
     let inferenceTime = 0
     for (let i = 0; i < iterations; i++) {
       const timeStart = performance.now()
-      for (let n = 0; n < nbNewTokens; n++) {
+      for (let n = 0; n < maxNewTokens; n++) {
         const next: number = (await model.predict(List.of(tokens))).first();
-	tokens = tokens.push(next)
+        tokens = tokens.push(next)
       }
       inferenceTime += performance.now() - timeStart
     }
-    console.log(`Inference time: ${(inferenceTime/ nbNewTokens / iterations).toFixed(2)} ms/token`)
+    console.log(`Inference time: ${(inferenceTime/ maxNewTokens / iterations).toFixed(2)} ms/token`)
   }
   await new Promise((resolve, reject) => {
     server.once('close', resolve)
--- a/cli/src/benchmark_gpt.ts
+++ b/cli/src/benchmark_gpt.ts
@@ -1,3 +1,4 @@
+import '@tensorflow/tfjs-node';
 import { List } from "immutable";
 import { parse } from "ts-command-line-args";
 import { AutoTokenizer } from "@xenova/transformers";
@@ -41,7 +42,7 @@ const args = { ...defaultArgs, ...parsedArgs }
  * Benchmark results are reported in https://github.com/epfml/disco/pull/659
  */
 
-async function main(args: Required<CLIArguments>): Promise<void> { 
+async function main(args: Required<CLIArguments>): Promise<void> {
   const { inference: benchmarkInference, modelType,
     contextLength, batchSize, modelPath } = args
 
@@ -77,10 +78,11 @@ async function main(args: Required<CLIArguments>): Promise<void> {
     task.trainingInformation.batchSize = batchSize
     task.trainingInformation.maxSequenceLength = contextLength
     const dataset = loadText('../datasets/wikitext/wiki.train.tokens')
+      .map(text => processing.tokenize(tokenizer, text))
+      .unbatch()
+      .batch(config.blockSize + 1, 1)
 
-    const maxLength = task.trainingInformation.maxSequenceLength ?? (tokenizer.model_max_length as number) + 1
     const preprocessedDataset = dataset
-      .map((line) => processing.tokenizeAndLeftPad(line, tokenizer, maxLength))
       .map((tokens) => [tokens.pop(), tokens.last()] as [List<number>, number])
       .batch(batchSize);
     
@@ -111,10 +113,7 @@ async function main(args: Required<CLIArguments>): Promise<void> {
     const iterations = 10
     console.log("Generating", maxNewTokens, "new tokens")
 
-    let tokens = List(
-      (tokenizer(prompt, { return_tensor: false }) as { input_ids: number[] })
-        .input_ids,
-    );
+    let tokens = processing.tokenize(tokenizer, prompt);
 
     let inferenceTime = 0
     for (let i = 0; i < iterations; i++) {
--- a/cli/src/benchmark_gpt.ts
+++ b/cli/src/benchmark_gpt.ts
@@ -69,18 +69,18 @@ async function main(args: Required<CLIArguments>): Promise<void> {
     const config: models.GPTConfig = {
       modelType: modelType as models.GPTConfig['modelType'],
       maxIter: iterationsPerEpoch,
-      blockSize: contextLength,
       lr: 0.0001,
+      contextLength,
     }
 
     // Load the dataset after setting the Task batch size and max sequence length
     // to make sure the dataset is batched and tokenized correctly
     task.trainingInformation.batchSize = batchSize
-    task.trainingInformation.maxSequenceLength = contextLength
+    task.trainingInformation.contextLength = contextLength
     const dataset = loadText('../datasets/wikitext/wiki.train.tokens')
       .map(text => processing.tokenize(tokenizer, text))
       .flatten()
-      .batch(config.blockSize + 1, 1)
+      .batch(config.contextLength + 1, 1)
 
     const preprocessedDataset = dataset
       .map((tokens) => [tokens.pop(), tokens.last()] as [List<number>, number])
--- a/cli/src/benchmark_gpt.ts
+++ b/cli/src/benchmark_gpt.ts
@@ -70,7 +70,6 @@ async function main(args: Required<CLIArguments>): Promise<void> {
       maxIter: iterationsPerEpoch,
       blockSize: contextLength,
       lr: 0.0001,
-      vocabSize: 50258 // default wikitext task uses the gpt2 tokenizer with vocabSize 50258
     }
 
     // Load the dataset after setting the Task batch size and max sequence length
--- a/discojs-node/src/loaders.spec.ts
+++ b/discojs-node/src/loaders.spec.ts
@@ -51,12 +51,13 @@ describe("image directory parser", () => {
 
 describe("text parser", () => {
   it("parses basic file", async () => {
+    const text = ["a", "b", "c"].join("\n")
     await withFile(async ({ path }) => {
-      await fs.writeFile(path, ["a", "b", "c"].join("\n"));
-
-      const parsed = loadText(path);
-
-      expect(await parsed.size()).to.equal(3);
+      await fs.writeFile(path, text);
+      
+      const sequences = await arrayFromAsync(loadText(path))
+      expect(sequences.length).to.equal(1);
+      expect(sequences[0]).to.equal(text);
     });
   });
 });
--- a/discojs-node/src/loaders/text.ts
+++ b/discojs-node/src/loaders/text.ts
@@ -1,14 +1,26 @@
-import * as fs from "node:fs/promises";
-import * as readline from "node:readline/promises";
-
+import createDebug from "debug";
+import { createReadStream } from 'node:fs';
 import { Dataset, Text } from "@epfml/discojs";
 
+const debug = createDebug("discojs-node:loaders:text");
+
+/**
+ * Returns chunks of text. Use `minChunkSize` to ensure that 
+ * each chunk is bigger than the expected sequence length.
+ * 
+ * @param path path to the text file to read
+ * @returns a dataset of tokenized input and label sequences
+ */
 export function load(path: string): Dataset<Text> {
   return new Dataset(async function* () {
-    const input = (await fs.open(path)).createReadStream({ encoding: "utf8" });
+    // Create a stream to read the text file chunk by chunk
+    const stream = createReadStream(path, { encoding: "utf8" });
+    for await (const chunk of stream) {
+      if (typeof chunk !== 'string')
+        throw new Error('Expected file stream to yield string')
 
-    // `readline` is a bit overkill but seems standard
-    // https://nodejs.org/api/readline.html#example-read-file-stream-line-by-line
-    yield* readline.createInterface({ input, crlfDelay: Infinity });
+      debug("yield chunk of length: %o", chunk.length);
+      yield chunk
+    }
   });
 }
--- a/discojs-web/src/loaders.spec.ts
+++ b/discojs-web/src/loaders.spec.ts
@@ -1,5 +1,4 @@
 import { describe, it, expect } from "vitest";
-
 import { loadCSV, loadText } from "./loaders/index.js";
 
 async function arrayFromAsync<T>(iter: AsyncIterable<T>): Promise<T[]> {
@@ -22,22 +21,16 @@ describe("csv parser", () => {
 });
 
 describe("text parser", () => {
-  it("loads", async () => {
+  it("loads a simple sequence", async () => {
+    const text = ["first", "second", "third"].join("\n")
+    
     // jsdom doesn't implement .text on File/Blob
     // trick from https://github.com/jsdom/jsdom/issues/2555
-    const text = await (
-      await fetch(
-        // data URL content need to be url-encoded
-        ["data:,first", "second", "third"].join("%0A"),
-      )
+    const file = await (
+      await fetch( "data:," + encodeURIComponent(text))
     ).blob();
-
-    const parsed = loadText(text);
-
-    expect(await arrayFromAsync(parsed)).to.have.ordered.members([
-      "first",
-      "second",
-      "third",
-    ]);
+    const parsed = loadText(file)
+    expect(await parsed.size()).to.equal(1);
+    expect((await arrayFromAsync(parsed))[0]).to.equal(text);
   });
 });
--- a/discojs-web/src/loaders/text.ts
+++ b/discojs-web/src/loaders/text.ts
@@ -1,35 +1,10 @@
 import { Dataset, Text } from "@epfml/discojs";
 
-class LineStream extends TransformStream<string, string> {
-  constructor() {
-    let current_line = "";
-
-    super({
-      transform: (chunk, controller) => {
-        const [head, ...lines] = chunk.split(/\r\n|\r|\n/);
-        const first_line = current_line + head;
-
-        if (lines.length === 0) {
-          current_line = first_line;
-          return;
-        }
-
-        controller.enqueue(first_line);
-        for (const line of lines.slice(0, -1)) controller.enqueue(line);
-
-        current_line = lines[lines.length - 1];
-      },
-      flush: (controller) => controller.enqueue(current_line),
-    });
-  }
-}
-
 export function load(file: Blob): Dataset<Text> {
   return new Dataset(async function* () {
     const reader = file
       .stream()
       .pipeThrough(new TextDecoderStream())
-      .pipeThrough(new LineStream())
       .getReader();
 
     while (true) {
--- a/discojs/src/dataset/dataset.ts
+++ b/discojs/src/dataset/dataset.ts
@@ -12,6 +12,22 @@ type DatasetLike<T> =
   | (() => AsyncIterator<T, void>)
   | (() => Iterator<T, void>);
 
+/** Convert a DatasetLike object to an async generator */
+async function* datasetLikeToGenerator<U>(content: DatasetLike<U>):
+  AsyncGenerator<U, void, undefined> {
+  let iter: AsyncIterator<U, void> | Iterator<U, void>;
+  if (typeof content === "function") iter = content();
+  else if (Symbol.asyncIterator in content)
+    iter = content[Symbol.asyncIterator]();
+  else iter = content[Symbol.iterator]();
+
+  while (true) {
+    const result = await iter.next();
+    if (result.done === true) break;
+    yield result.value;
+  }
+}
+
 /** Immutable series of data */
 export class Dataset<T> implements AsyncIterable<T> {
   readonly #content: () => AsyncIterator<T, void, undefined>;
@@ -23,18 +39,8 @@ export class Dataset<T> implements AsyncIterable<T> {
    */
   constructor(content: DatasetLike<T>) {
     this.#content = async function* () {
-      let iter: AsyncIterator<T, void> | Iterator<T, void>;
-      if (typeof content === "function") iter = content();
-      else if (Symbol.asyncIterator in content)
-        iter = content[Symbol.asyncIterator]();
-      else iter = content[Symbol.iterator]();
-
-      while (true) {
-        const result = await iter.next();
-        if (result.done === true) break;
-        yield result.value;
-      }
-    };
+      yield* datasetLikeToGenerator(content);
+    }
   }
 
   [Symbol.asyncIterator](): AsyncIterator<T> {
@@ -160,11 +166,13 @@ export class Dataset<T> implements AsyncIterable<T> {
     );
   }
 
-  /** Flatten chunks */
-  unbatch<U>(this: Dataset<Batched<U>>): Dataset<U> {
+  /** Flatten batches/arrays of elements */
+  flatten<U>(this: Dataset<DatasetLike<U>>): Dataset<U> {
     return new Dataset(
-      async function* (this: Dataset<Batched<U>>) {
-        for await (const batch of this) yield* batch;
+      async function* (this: Dataset<DatasetLike<U>>) {
+        for await (const batch of this) {
+          yield* datasetLikeToGenerator(batch);
+        }
       }.bind(this),
     );
   }
--- a/discojs/src/dataset/dataset.ts
+++ b/discojs/src/dataset/dataset.ts
@@ -112,22 +112,36 @@ export class Dataset<T> implements AsyncIterable<T> {
     ];
   }
 
-  /** Slice into chunks
-   *
-   * Last slice is smaller if dataset isn't perfectly divisible
+  /** Create batches of `size` elements with potential overlap.
+   * Last batch is smaller if dataset isn't perfectly divisible
+   * 
+   * If overlap is set to a positive integer, the last `overlap` elements of a batch 
+   * are the first `overlap` elements of the next batch.
+   * 
+   * This method is tailored to create text sequences where each token's label is the following token. 
+   * In order to have a label for the last token of the input sequence, we include the first token
+   * of the next sequence (i.e. with an overlap of 1).
    *
    * @param size count of element per chunk
+   * @param overlap number of elements overlapping between two consecutive batches
    */
-  batch(size: number): Dataset<Batched<T>> {
-    if (size <= 0 || !Number.isInteger(size)) throw new Error("invalid size");
+  batch(size: number, overlap = 0): Dataset<Batched<T>> {
+    if (size <= 0 || !Number.isInteger(size))
+      throw new Error("invalid size");
+    if (overlap >= size || !Number.isInteger(overlap))
+      throw new Error("invalid overlap");
 
     return new Dataset(
       async function* (this: Dataset<T>) {
         const iter = this[Symbol.asyncIterator]();
 
+        let overlapped = List<T>();
         for (;;) {
           const batch = List(
-            await Promise.all(Range(0, size).map(() => iter.next())),
+            // get the first elements of the next batch
+            await Promise.all(
+              Range(overlapped.size, size).map(() => iter.next())
+            )
           ).flatMap((res) => {
             if (res.done) return [];
             else return [res.value];
@@ -135,10 +149,12 @@ export class Dataset<T> implements AsyncIterable<T> {
 
           if (batch.isEmpty()) break;
 
-          yield batch;
+          // yield the current batch with the first elements of the next batch
+          yield overlapped.concat(batch);
+          overlapped = batch.takeLast(overlap);
 
           // iterator couldn't generate more
-          if (batch.size < size) break;
+          if (batch.size < size - overlap) break;
         }
       }.bind(this),
     );
@@ -172,6 +188,26 @@ export class Dataset<T> implements AsyncIterable<T> {
           if (l.done || r.done) return;
           yield [l.value, r.value] as [T, U];
         }
+      }.bind(this),
+    );
+  }
+
+  /**
+   * Repeat the dataset `times` times
+   * @param times number of times to repeat the dataset, if undefined, the dataset is repeated indefinitely
+   * @returns a dataset repeated `times` times
+   */
+  repeat(times?: number): Dataset<T> {
+    if (times !== undefined && (!Number.isInteger(times) || times < 1))
+      throw new Error("times needs to be a positive integer or undefined");
+
+    return new Dataset(
+      async function* (this: Dataset<T>) {
+        let loop = 0;
+        do {
+          yield* this;
+          loop++
+        } while (times === undefined || loop < times)
       }.bind(this),
     );
   }
--- a/discojs/src/models/gpt/config.ts
+++ b/discojs/src/models/gpt/config.ts
@@ -19,20 +19,17 @@ export interface GPTConfig {
   maxIter?: number
   weightDecay?: number
   verbose?: 0 | 1
-  bias?: boolean
   debug?: boolean
   dropout?: number
   residDrop?: number
   embdDrop?: number
-  tokEmb?: boolean
-  lmHead?: boolean
   nLayer?: number
   nHead?: number
   nEmbd?: number
 }
 // for a benchmark of performance, see https://github.com/epfml/disco/pull/659
 export const DEFAULT_CONFIG: Required<GPTConfig> = {
-  name: 'transformer',
+  name: 'transformer', // prefix for the model layer names
   lr: 0.001,
   weightDecay: 0,
   maxIter: 10,
@@ -43,13 +40,10 @@ export const DEFAULT_CONFIG: Required<GPTConfig> = {
   evaluateEvery: 100,
   blockSize: 128,
   vocabSize: 50257,
-  bias: true,
   debug: false,
   dropout: 0.2,
   residDrop: 0.2,
   embdDrop: 0.2,
-  tokEmb: true,
-  lmHead: true,
   nLayer: 3,
   nHead: 3,
   nEmbd: 48,
--- a/discojs/src/models/gpt/config.ts
+++ b/discojs/src/models/gpt/config.ts
@@ -26,6 +26,7 @@ export interface GPTConfig {
   nLayer?: number
   nHead?: number
   nEmbd?: number
+  seed?: number,
 }
 // for a benchmark of performance, see https://github.com/epfml/disco/pull/659
 export const DefaultGPTConfig: Required<GPTConfig> = {
@@ -47,6 +48,7 @@ export const DefaultGPTConfig: Required<GPTConfig> = {
   nLayer: 3,
   nHead: 3,
   nEmbd: 48,
+  seed: Math.random(),
 }
 
 export type ModelSize = {
@@ -91,6 +93,6 @@ export interface GenerationConfig {
 export const DefaultGenerationConfig: Required<GenerationConfig> = {
   temperature: 1.0,
   doSample: false,
-  seed: 42,
+  seed: Math.random(),
   topk: 50
 }
--- a/discojs/src/models/gpt/config.ts
+++ b/discojs/src/models/gpt/config.ts
@@ -9,7 +9,7 @@ type GPTModelType =
 
 export interface GPTConfig {
   lr: number
-  blockSize: number
+  contextLength: number
   vocabSize?: number
   modelType: GPTModelType
   name?: string,
@@ -39,7 +39,7 @@ export const DefaultGPTConfig: Required<GPTConfig> = {
   evaluate: true,
   maxEvalBatches: 12,
   evaluateEvery: 100,
-  blockSize: 128,
+  contextLength: 128,
   vocabSize: 50257,
   debug: false,
   dropout: 0.2,
--- a/discojs/src/models/gpt/config.ts
+++ b/discojs/src/models/gpt/config.ts
@@ -28,7 +28,7 @@ export interface GPTConfig {
   nEmbd?: number
 }
 // for a benchmark of performance, see https://github.com/epfml/disco/pull/659
-export const DEFAULT_CONFIG: Required<GPTConfig> = {
+export const DefaultGPTConfig: Required<GPTConfig> = {
   name: 'transformer', // prefix for the model layer names
   lr: 0.001,
   weightDecay: 0,
@@ -72,4 +72,25 @@ export function getModelSizes (modelType: GPTModelType): Required<ModelSize> {
     case 'gpt-nano':
       return { nLayer: 3, nHead: 3, nEmbd: 48 }
   }
+}
+
+export interface GenerationConfig {
+  // take random token weighted by its probability
+  // If false, predict the token with the highest probability.
+  doSample: boolean
+  // the generation temperature (higher means more randomness).
+  // Set to 0 for greedy decoding.
+  temperature: number
+  // only consider the topk most likely tokens for sampling. 
+  // used if doSample is true.
+  topk: number
+  // random seed for sampling.
+  seed: number
+}
+
+export const DefaultGenerationConfig: Required<GenerationConfig> = {
+  temperature: 1.0,
+  doSample: false,
+  seed: 42,
+  topk: 50
 }
--- a/discojs/src/models/gpt/config.ts
+++ b/discojs/src/models/gpt/config.ts
@@ -10,7 +10,7 @@ type GPTModelType =
 export interface GPTConfig {
   lr: number
   blockSize: number
-  vocabSize: number
+  vocabSize?: number
   modelType: GPTModelType
   name?: string,
   evaluate?: boolean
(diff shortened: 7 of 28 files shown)


discojs-node/src/loaders/text.ts
```diff
@@ -1,14 +1,98 @@
...
+          tokens.slice(currentPosition),
+          { skip_special_tokens: true }
+        )
+        debug("End of chunk, remaining text: '%s'", endOfPreviousChunk)
+      } else {
+        // Note that the difference between tokenizing and then concatenating
+        // vs concatenating and then tokenizing can happen if their is no
+        // remaining text. We consider this difference negligible
+        endOfPreviousChunk = "";
+      }
```
tharvik: duplicated in discojs-web/loaders/text. that hints to me that it shouldn't happen in the loader but applied after.
the issue at hand is that lines where outputted by the previous version. I think we can change it to output characters (single letter `string`) instead. that also would drop the `blockSize`, `batchSize` & `minChunkSize` argument which aren't really relevant for reading text ([separation of concerns](https://en.wikipedia.org/wiki/Separation_of_concerns) and all that)

in the newly merged processing PR (#781), it is much simpler to combine such transformation, I think that smth like
```ts
loadText($path).batch($blockSize).map((block) => tokenize(block, $tokenizer))
```
with tokenize updated to accept block/`List<string>` instead, and maybe drop the padding (but what would be the behavior at the end of the file?)

discojs-node/src/loaders/text.ts
```diff
@@ -1,14 +1,98 @@
...
+ * big enough to contain at least one batch.
+ * @param minChunkSize default to 16KiB, the minimum size of each chunk in bits
+ * @returns a dataset of tokenized input and label sequences
+ */
+export function load(path: string, tokenizer: PreTrainedTokenizer,
+  blockSize: number, batchSize: number = 1, minChunkSize = 16384): Dataset<Text> {
   return new Dataset(async function* () {
-    const input = (await fs.open(path)).createReadStream({ encoding: "utf8" });
+    if (batchSize < 1 || blockSize < 1 || minChunkSize < 1)
+      throw new Error("batchSize, blockSize and minChunkSize must be positive integers");
```
tharvik: positive checked but not that it is integral.

```suggestion
    if (batchSize < 1 || !Number.isInteger(batchSize) || blockSize < 1 || !Number.isInteger(blockSize) || minChunkSize < 1 || !Number.isInteger(minChunkSize))
      throw new Error("batchSize, blockSize and minChunkSize must be positive integers");
```

discojs-node/src/loaders/text.ts
```diff
@@ -1,14 +1,98 @@
...
+const debug = createDebug("discojs-node:loaders:text");
+
+/**
+ * Returns a Dataset that streams and tokenizes text to yield tokenized sequences
+ * one at a time. Each sequence has size `blockSize` + 1, where the first `blockSize`
+ * tokens are the input and the last token is the label. The following sequence
+ * starts with the last token of the previous sequence (so the previous label is now the 
+ * first input token).
+ * In other words, the dataset yields sequences of size `blockSize` + 1 but with an overlap
+ * of 1 token between each sequence.
```
tharvik: why is the overlap needed? I mean, it looks like a more complicated feature than a needed one. I guess that you wanted to use the dataset "in full" and not let the label of a given block only in the output.
based on the understanding I got from [the video you sent me a while back](https://www.youtube.com/watch?v=kCc8FmEb1nY&t=1020), if we want to actually use each token fairly, we need to "slide over" the token stream (such as [`slice.windows` in glorious Rust](https://doc.rust-lang.org/nightly/std/primitive.slice.html#method.windows)) (it still treat unfairly the end of the file but that should be fixed by padding if really wanted).
edit: that also helps with avoiding the "+ 1" everywhere.

cli/src/benchmark_gpt.ts
```diff
@@ -79,43 +79,24 @@ async function main(args: Required<CLIArguments>): Promise<void> {
       maxIter: iterationsPerEpoch,
       blockSize: contextLength,
       lr: 0.0001,
-      vocabSize: 50258 // default wikitext task uses the gpt2 tokenizer with vocabSize 50258
+      vocabSize: 50257 // default wikitext task uses the gpt2 tokenizer with vocabSize 50257
```
tharvik: can we have it as a default in `GPTConfig`? I saw it a few times and that's quite a magic value and I didn't found a way to extract it from the tokenizer either.

discojs-node/src/loaders.spec.ts
```diff
@@ -50,13 +52,83 @@ describe("image directory parser", () => {
...
+      expect(await parsed.size()).to.equal(1);
+      const next = await parsed[Symbol.asyncIterator]().next()
+      expect(next.done).to.be.false;
+      
+      const tokens = next.value as number[]
+      const expectedTokens = models.tokenize(tokenizer, text, {
+        padding: false,
+        truncation: false,
+        return_tensor: false
+      })
```
tharvik: yeah, better to test for the actual tokens values. it is not really an issue but it assumes that `models.tokenize` will always work as expected. also, I built test codebases with a bunch of logic testing like this and while it looks correct, it breaks quite easily.

discojs-node/src/loaders.spec.ts
```diff
@@ -50,13 +52,83 @@ describe("image directory parser", () => {
...
-      await fs.writeFile(path, ["a", "b", "c"].join("\n"));
+      const text = ["a", "b", "c"].join("\n")
+      await fs.writeFile(path, text);
+      // set block size to 4 to get 1 sequence of 4 tokens + 1 label token
+      const parsed = loadText(path, tokenizer, 4, 1);
+      expect(await parsed.size()).to.equal(1);
+      const next = await parsed[Symbol.asyncIterator]().next()
+      expect(next.done).to.be.false;
+      
+      const tokens = next.value as number[]
```
tharvik: better to check for the whole dataset content, with `Array.fromAsync` (as locally redone in discojs/src/dataset/dataset.spec.ts), that's more readable and ensure that it doesn't generate more values that expected

discojs-node/src/loaders.spec.ts
```diff
@@ -50,13 +52,83 @@ describe("image directory parser", () => {
 });
 
 describe("text parser", () => {
+  
   it("parses basic file", async () => {
+    const tokenizer = await AutoTokenizer.from_pretrained('Xenova/gpt2')
```
tharvik: we also test with "Xenova/gpt2", but our code should work with any tokenizer; can you switch it to another one?

discojs-node/src/loaders.spec.ts
```diff
@@ -50,13 +52,83 @@ describe("image directory parser", () => {
...
+      const blockSize = 4
+      const parsed = loadText(path, tokenizer, blockSize, 1);
+      // expect the number of sequences to be the total number of tokens divided by blockSize
+      // we use floor because the last incomplete sequence is dropped
+      expect(await parsed.size()).to.equal(Math.floor(expectedTokens.length / blockSize));
+      
+      let i = 0
+      for await (const tokens of parsed) {
+        // each sequence should have length blockSize + 1 (for the label)
+        expect(tokens).to.deep.equal(expectedTokens.slice(i, i + blockSize + 1));
```
tharvik: this tests more that advertisez, you can simplify it by only checking for the size of each block.

discojs-node/src/loaders.spec.ts
```diff
@@ -50,13 +52,83 @@ describe("image directory parser", () => {
...
+        // but the window should move by blockSize only
+        i += blockSize
+      }
+    })
+  });
+
+  it("reads multiple chunks", async () => {
+    const tokenizer = await AutoTokenizer.from_pretrained('Xenova/gpt2')
+    await withFile(async ({ path }) => {
+      const text = "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Donec sed risus maximus, ultricies ex sed, dictum elit. Curabitur faucibus egestas enim et auctor. Quisque vel dignissim turpis. Curabitur justo tellus, elementum sit amet erat eget, auctor ornare nisi. Nunc tortor odio, ultrices id leo vitae, euismod congue ex. Curabitur arcu leo, sagittis quis felis nec, imperdiet aliquet tellus. Integer a mollis nulla. Quisque pulvinar lectus eget nisi pharetra, non molestie magna ullamcorper. Sed porttitor diam non blandit molestie. Duis tristique arcu ut efficitur efficitur. Fusce et ullamcorper tortor. Pellentesque a accumsan lacus, nec mollis risus. Nunc quis eros a orci ultricies cursus. Maecenas sodales ipsum a magna malesuada efficitur. Maecenas at sapien blandit, egestas nisi eu, mollis elit."
```
tharvik: please split it into multiple lines using `[…].join(" ")`, that's more readable.

discojs-node/src/loaders.spec.ts
```diff
@@ -50,13 +52,83 @@ describe("image directory parser", () => {
...
+        expect(tokens).to.deep.equal(expectedTokens.slice(i, i + blockSize + 1));
+        // but the window should move by blockSize only
+        i += blockSize
+      }
+    })
+  });
+
+  it("reads multiple chunks", async () => {
+    const tokenizer = await AutoTokenizer.from_pretrained('Xenova/gpt2')
+    await withFile(async ({ path }) => {
```
tharvik: consider moving the definitions and the actions not neede the file outside of `withFile` (also for the previous ones). you can actually return a value out of `withFile` if wanted.

discojs-node/src/loaders.spec.ts
```diff
@@ -50,13 +52,83 @@ describe("image directory parser", () => {
...
+      // we use floor because the last incomplete sequence is dropped
+      expect(await parsed.size()).to.equal(Math.floor(expectedTokens.length / blockSize));
+    
+      let i = 0
+      for await (const tokens of parsed) {
+        // each sequence should have length blockSize + 1 (for the label)
+        expect(tokens).to.deep.equal(expectedTokens.slice(i, i + blockSize + 1));
+        // but the window should move by blockSize only
+        i += blockSize
+      }
```
tharvik: duplicated from previous test. consider a common function.

discojs-node/src/loaders/text.ts
```diff
@@ -1,14 +1,98 @@
...
+ * 
+ * @param path path to the text file to read
+ * @param tokenizer the tokenizer to use, should match the model that will be trained
+ * @param blockSize the context length, the maximum number of tokens of input sequences
+ * @param batchSize default to 1, the number of input sequences (of `blockSize` tokens) in each batch.
```
tharvik: seems to be kinda duplicated from blockSize

discojs-web/src/loaders.spec.ts
```diff
@@ -22,22 +24,51 @@ describe("csv parser", () => {
...
+
     // jsdom doesn't implement .text on File/Blob
     // trick from https://github.com/jsdom/jsdom/issues/2555
-    const text = await (
-      await fetch(
-        // data URL content need to be url-encoded
-        ["data:,first", "second", "third"].join("%0A"),
-      )
+    const file = await (
+      await fetch( "data:," + encodeURIComponent(text))
```
tharvik, line 30: hooo nice, well found!

discojs-web/src/loaders.spec.ts
tharvik, line 1: similar remarks as the one in discojs-node

discojs-web/src/loaders/text.ts
```diff
@@ -1,35 +1,76 @@
...
+        if (tokens.length < blockSize + 1) {
+          // throw if it happens on the 1st chunk
+          if (chunkNumber === 0)
+            throw new Error(`the chunk (${tokens.length} tokens) is too small ` +
+              `to get a sequence of length blockSize (${blockSize + 1} tokens). ` +
+              `Either the text file or the chunk size is too small.`);
+          // if this isn't the first iteration we simply skip
+          // as we expect the last chunk to be potentially smaller than the block size
+          debug("chunk smaller than block size, loading next chunk")
+          return
```
tharvik: isn't it the case where padding can be used instead? otherwise, we force a minimal text size (which is also hard to estimate, as various text length can generate the same number of token)

discojs-web/src/loaders/text.ts
```diff
@@ -1,35 +1,76 @@
...
+        }
+        // keep the last tokens for the next chunk
+        // if this was the last chunk the remaining tokens are discarded
+        if (currentPosition < tokens.length) {
+          endOfPreviousChunk = tokenizer.decode(
+            tokens.slice(currentPosition),
+            { skip_special_tokens: true }
+          )
+        }
+        else endOfPreviousChunk = "";
```
tharvik: using an index when it's a linear traversal is usually not needed and makes it harder to grasp.

```suggestion
        while (tokens.length >= blockSize + 1) {
          controller.enqueue(tokens.slice(0, blockSize + 1));
          tokens = tokens.slice(blockSize); // no +1 here
        }
        // keep the last tokens for the next chunk
        // if this was the last chunk the remaining tokens are discarded
        endOfPreviousChunk =
          tokens.length > 0
            ? tokenizer.decode(tokens, { skip_special_tokens: true })
            : "";
```

discojs-web/src/loaders/text.ts
```diff
@@ -1,35 +1,76 @@
...
+        chunkNumber++;
       },
-      flush: (controller) => controller.enqueue(current_line),
+      // No flush, discard the last tokens
```
tharvik: padding maybe? it's weird to have a loader of data not actually outputting all it can.

discojs/src/models/gpt/index.ts
```diff
@@ -133,16 +135,31 @@ export class GPT extends Model {
...
+   * @param tokenizer tokenizer object
+   * @param generationConfig config object for text generation
+   * @param generationConfig.maxNewTokens default to 10, the number of tokens to generate
+   * @param generationConfig.temperature default to 1.0, the generation temperature (higher means more randomness). 
+   * Set to 0 for greedy decoding.
+   * @param generationConfig.doSample default to false, if true, sample the next token according the model's probabilities
+   * If false, predict the token with the highest probability.
+   * @param generationConfig.topk default to 50, only consider the topk most likely tokens for sampling. 
+   * Only used if doSample is true.
+   * @param generationConfig.seed default to 42, random seed for sampling.
```
tharvik: better to push theses inside `GenerationConfig`: if we want at some point to publish the API (which I think we should if we want more devs to try out disco), it will be shown next to each field.

discojs/src/models/gpt/index.ts
```diff
@@ -133,16 +135,31 @@ export class GPT extends Model {
...
+   * @returns the input prompt concatenated with the generated tokens
+   */
+  async generate(input: string, tokenizer: PreTrainedTokenizer, generationConfig?: Partial<GenerationConfig>): Promise<string> {
+    const { input_ids: tokens } = await tokenizer(input, {
+      return_tensor: false,
+      padding: false,
+    }) as { input_ids: number[] }
```
tharvik: duplicated from `discojs/src/models/tokenizer.tokenize`

discojs/src/models/tokenizer.ts
tharvik, line 1: yeah, I kind of dislike this file: it's exposing ~internal methods which isn't used outside of discojs-*.
* `getTaskTokenizer` is a quirk to correctly use `Task<"text">`
  * we can simplify this by having a serialization of `Task`, which would take care of materializing `Task.trainingInformation.tokenizer`, apply `Task`'s defaults, parse some fields as URL, …
* `tokenize` will probably not be needed anymore when rebased, as it would be in a processing method

discojs/src/models/gpt/layers.ts
```diff
@@ -36,10 +42,17 @@ class LogLayer extends tf.layers.Layer {
...
-      }
+      if (Array.isArray(input)) input = input[0]
       this.invokeCallHook(input, kwargs)
+
+      const logs = {
+        // 'shape': input.shape,
+        // 'is_only_zero': !!input.equal(tf.tensor(0)).all().dataSync()[0],
+        // 'has_some_NaN': !!input.isNaN().any().dataSync()[0],
+        'min': +input.min().dataSync()[0].toPrecision(3),
+        'max': +input.max().dataSync()[0].toPrecision(3),
```
tharvik, line 59: is there a way to avoid `dataSync`? that hold the thread waiting for the data instead of allowing others to run

discojs/src/models/gpt/config.ts
```diff
@@ -42,17 +40,15 @@ export const DEFAULT_CONFIG: Required<GPTConfig> = {
...
   debug: false,
   dropout: 0.2,
   residDrop: 0.2,
   embdDrop: 0.2,
-  tokEmb: true,
-  lmHead: true,
   nLayer: 3,
   nHead: 3,
   nEmbd: 48,
+  seed: 42,
```
tharvik: that means that unless explicitly set, the model will always behave the same way, which might be unexpected. maybe we can `Math.random()` it if left undef before applying the default config?

discojs/src/models/gpt/layers.ts
```diff
@@ -78,26 +95,35 @@ class CausalSelfAttention extends tf.layers.Layer {
...
-      'c_proj/kernel',
+      'c_proj.kernel',
       [this.nEmbd, this.nEmbd],
       'float32',
-      tf.initializers.glorotNormal({})
+      // the input keeps accumulating through the residual stream so we
+      // scale the initialization with the nb of layers to keep a unit std
+      // Sources:
+      // https://github.com/karpathy/build-nanogpt/blob/6104ab1b53920f6e2159749676073ff7d815c1fa/train_gpt2.py#L103
+      // https://youtu.be/l8pRSuU81PU?si=5GcKfi_kPgLgvtg2&t=4640
```
tharvik, line 126: nope to youtube tracking id

```suggestion
      // https://youtu.be/l8pRSuU81PU?t=4640
```

discojs/src/models/gpt/layers.ts
```diff
@@ -121,77 +147,77 @@ class CausalSelfAttention extends tf.layers.Layer {
...
       }
       // Apply attention weights to inputs as one big matrix which is then split into the
       // query, key and value submatrices
+      // nHead is "number of heads", hs is "head size", and C (number of channels) = n_embd = nHead * hs
+      // e.g. in GPT-2 (124M), nHead = 12, hs = 64, so nHead * hs = C = 768 channels in the Transformer
       const cAttn = dense(input, this.cAttnKernel, this.cAttnBias)
-
       let [q, k, v] = tf.split(cAttn, 3, -1) as [tf.Tensor, tf.Tensor, tf.Tensor]
-      const [B, T, C] = k.shape
+      const [B, T, C] = k.shape // batch size, sequence length, embedding dimensionality (number of channels)
```
tharvik, line 175: instead of the comment, we can rename the variable? that helps when reading how it is used.
or at least, link the code/paper where theses names are used

discojs/src/models/gpt/layers.ts
```diff
@@ -121,77 +147,77 @@ class CausalSelfAttention extends tf.layers.Layer {
...
+      y = dense(y, this.cProjKernel, this.cProjBias) // output projection (B, T, C)
+      y = kwargs.training === true ? tf.dropout(y, this.dropout, undefined, this.seed) : y
       return y
     })
   }
 }
 tf.serialization.registerClass(CausalSelfAttention)
 
+/**
+ * GELU with tanh approximate
```
tharvik, line 226: here my non-knowledge of ML shows: is see that many layers in TFJS can be activated using various function, such as 'gelu' or 'gelu_new'. as this layer is used once, between some dense layer, isn't it possible to merge some and drop this custom GELU implementation?

discojs/src/models/gpt/layers.ts
```diff
@@ -210,81 +236,197 @@ class GELU extends tf.layers.Layer {
...
+      'wte', //use same name as GPT2
+      [this.vocabSize, this.nEmbd],
+      'float32',
+      tf.initializers.randomNormal({ mean:0, stddev:0.02, seed: this.seed })
+    )
+  }
+
+  override computeOutputShape(inputShape: tf.Shape | tf.Shape[]): tf.Shape | tf.Shape[] {
+    let shape: tf.Shape
+    if (Array.isArray(inputShape) && Array.isArray(inputShape[0])) shape = inputShape[0]
```
tharvik: it's only supporting a single `tf.Shape` there, consider throwing rather than silently dropping dimensions.

discojs/src/models/gpt/layers.ts
```diff
@@ -210,81 +236,197 @@ class GELU extends tf.layers.Layer {
...
+      // https://github.com/tensorflow/tfjs/blob/3daf152cb794f4da58fce5e21e09e8a4f89c8f80/tfjs-layers/src/layers/embeddings.ts#L155
+      // batch size and sequence length are undetermined 
+      // so the output shape is [null, null, nEmbd]
+      return [...shape, this.nEmbd]
+    }
+    // input shape for the language modeling head
+    // https://github.com/tensorflow/tfjs/blob/3daf152cb794f4da58fce5e21e09e8a4f89c8f80/tfjs-layers/src/layers/core.ts#L258
+    else if (shape.length === 3) {
+      shape[2] = this.vocabSize
+      return shape // [null, null, vocabSize]
```
tharvik: this reads more easily. (maybe you can also check before that `shape[0..1]` are indeed `null`.

```suggestion
      return [null, null, this.vocabSize]
```

discojs/src/models/gpt/layers.ts
```diff
@@ -210,81 +236,197 @@ class GELU extends tf.layers.Layer {
...
+  override computeOutputShape(inputShape: tf.Shape | tf.Shape[]): tf.Shape | tf.Shape[] {
+    let shape: tf.Shape
+    if (Array.isArray(inputShape) && Array.isArray(inputShape[0])) shape = inputShape[0]
+    else shape = inputShape as tf.Shape
+    // input shape for the token embedding
+    if (shape.length === 2) {
+      // https://github.com/tensorflow/tfjs/blob/3daf152cb794f4da58fce5e21e09e8a4f89c8f80/tfjs-layers/src/layers/embeddings.ts#L155
+      // batch size and sequence length are undetermined 
+      // so the output shape is [null, null, nEmbd]
+      return [...shape, this.nEmbd]
```
tharvik: better to simply return the values if we are sure of what it is.

```suggestion
      return [null, null, this.nEmbd]
```

discojs/src/models/gpt/layers.ts
```diff
@@ -210,81 +236,197 @@ class GELU extends tf.layers.Layer {
...
+      shape[2] = this.vocabSize
+      return shape // [null, null, vocabSize]
+    }
+    else throw new Error('unexpected input shape')
+  }
+
+  override call (input: tf.Tensor | tf.Tensor[], kwargs: Record<string, unknown>): tf.Tensor | tf.Tensor[] {
+    return tf.tidy(() => {
+      if (this.embeddings === undefined) throw new Error('not built')
+      if (Array.isArray(input)) input = input[0]
```
tharvik: drops instead of the safer throw

discojs/src/models/gpt/layers.ts
```diff
@@ -210,81 +236,197 @@ class GELU extends tf.layers.Layer {
...
+      // If the input is a 2D tensor, it is a batch of sequences of tokens
+      // so we translate the tokens into embeddings 
+      // using `this.embeddings` as a lookup table
+      if (input.shape.length === 2) {
+        // (batch_size, sequence_length) => (batch_size, sequence_length, nEmbd)
+        return tf.gather(this.embeddings.read(), tf.cast(input, 'int32'), 0)
+      }
+      // If the input is a 3D tensor, it is a sequence of embeddings
+      // so we apply a dense layer to project the embeddings back into the vocabulary space
+      else if (input.shape[2] === this.nEmbd) {
```
tharvik: more in sync with comment #nitpick (also consider using a switch, I find it sometimes more readable)

```suggestion
      else if (input.shape.length === 3) {
```

discojs/src/models/gpt/model.ts
```diff
@@ -194,44 +177,70 @@ function prepareIdx (idx: tf.TensorLike): tf.Tensor2D {
...
-  async generate (idxRaw: tf.TensorLike, conf: GenerateConfig): Promise<number[][]> {
-    const config = Object.assign({}, defaultGenerateConfig, conf)
-    let idx = prepareIdx(idxRaw)
+
+  /**
+   * 
+   * @param input 
+   * @param config 
+   * @returns 
+   */
```
tharvik: empty

```suggestion
```

discojs/src/models/gpt/model.ts
```diff
@@ -194,44 +177,70 @@ function prepareIdx (idx: tf.TensorLike): tf.Tensor2D {
...
+        .slice([0, sequenceBatch.shape[1] - 1, 0]) // take the logits at the last position (B, 1, vocab_size)
+        .reshape([logits.shape[0], logits.shape[2]]) // (B, vocab_size)
+        .div<tf.Tensor2D>(tf.scalar(config.temperature)) // the higher the temperature, the more random the output
+      const probs = logitsScaled.softmax(-1) // get the token probabilities (B, vocab_size) 
       if (config.doSample) {
-        return tf.multinomial(probs, 1) as tf.Tensor2D
+        // returns topk biggest values among the `vocab_size` probabilities and the corresponding tokens indices 
+        // both shapes are (B, config.topk)
+        const { values: topkProbs, indices: topkTokens } = tf.topk(probs, config.topk);
+        // sample an index from the top-k probabilities
```
tharvik: I don't quite understand why we filter the top-K values before sampling. is it to avoid unexpected (with a low probability) values? if yes, top-K would still allow unlikely values (if the distribution is very skewed toward a single token and not the others). maybe a threshold would be safer?

docs/CONTRIBUTING.md
```diff
@@ -104,6 +104,69 @@ npm -w webapp test
 
 The webapp tests rely on `cypress` and the test suite is located in the `webapp/cypress` folder.
 
+Note that you can also run test interactively in the browser of your choice. To do so, run
+```
+cd webapp
+VITE_SERVER_URL=http://server npx start-server-and-test start http://localhost:8081 'cypress open --e2e'
```
tharvik: you can use the same workspace argument as with `npm`

```suggestion
VITE_SERVER_URL=http://server npx -w webapp start-server-and-test start http://localhost:8081 'cypress open --e2e'
```

docs/CONTRIBUTING.md
```diff
@@ -104,6 +104,69 @@ npm -w webapp test
...
+It is possible to record the cypress tests ran in the Github Actions CI and visualize them in the [Cypress Cloud](cloud.cypress.io). It is currently used only when needed (because the free plan has a limited number of recordings). The [cypress documentation](https://docs.cypress.io/app/continuous-integration/github-actions) describes how to set up the recordings.
+
+1. A Disco project has been created in the Cypress Cloud and you need to be added to the project to be able to visualize the recordings.
+
+2. In `webapp/cypress.config.ts` you need to add the project ID, which is currently:
+```js
+  projectId: "aps8et"
+```
+
+3. Finally, we can set the Cypress parameters in the Github Actions workflow `.github/workflows/lint-test-build.yml`, for example:
```
tharvik: we can add a workflow especially for cypress, using a [`workflow_dispatch`](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#workflow_dispatch) event to only act when manually triggered

docs/CONTRIBUTING.md
```diff
@@ -104,6 +104,69 @@ npm -w webapp test
...
+test-webapp:
+    needs: [build-lib, build-lib-web, download-datasets]
+    runs-on: ubuntu-latest
+    # Runs tests in parallel with matrix strategy https://docs.cypress.io/guides/guides/parallelization
+    # https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs
+    # Also see warning here https://github.com/cypress-io/github-action#parallel
+    strategy:
+      fail-fast: false # https://github.com/cypress-io/github-action/issues/48
+      matrix:
+        containers: [1, 2] # Uses 2 parallel instances
```
tharvik: why parallelism is needed?

webapp/cypress.config.ts
```diff
@@ -9,6 +9,10 @@ export default defineConfig({
       on("task", {
         readdir: async (p: string) =>
           (await fs.readdir(p)).map((filename) => path.join(p, filename)),
+        log: (message) => {
+          console.log(message)
+          return null
+        },
```
tharvik: debug?

webapp/src/components/dataset_input/TextDatasetInput.vue
```diff
@@ -10,20 +10,27 @@
...
-import type { Text } from "@epfml/discojs";
-import { Dataset } from "@epfml/discojs";
+import type { Text, Task } from "@epfml/discojs";
+import { Dataset, models } from "@epfml/discojs";
 import { loadText } from "@epfml/discojs-web";
 
 import DatasetInput from "./DatasetInput.vue";
 import FileSelection from "./FileSelection.vue";
 
+const { task } = defineProps<{ task: Task }>()
```
tharvik: [destructuring props only works in vue >= 3.5](https://vuejs.org/guide/components/props.html#reactive-props-destructure) but we are using vue 3.4 here

tharvik: I tried pretty hard to avoid referencing `Task` (the Ugly Global as I call it in my sleep), hopefully, it'll won't be needed with the latest changes to `loadText`

webapp/src/components/testing/TestSteps.vue
```diff
@@ -39,12 +40,12 @@
         against a chosen dataset of yours. Below, once you assessed the model,
         you can compare the ground truth and the predicted values
         <div class="flex justify-center mt-4">
-          <CustomButton @click="startTest()"> test </CustomButton>
+          <CustomButton @click="startTest()" data-cy="start-test"> test </CustomButton>
```
tharvik: yeah, I saw that a few times in cypress' examples but I don't like this pattern. it mix test specific code into the rest of the codebase. especially here, it is possible to find a button with a "test" label.
same idea below.

tharvik (changes requested):
superb work, thanks for clearing the GPT's mud, every comments makes it more understandable!
yeah, sadly, as I forgot to merge the processing PR (#781) before you branched off, the whole processing pipeline changed a lot. sorry for the toes stepping (hopefully, it will simplify this PR).

btw, it seems that @xenova/transformer [has been recently updated to @huggingface/transformer](https://github.com/huggingface/transformers.js/pull/545). did you try it out? maybe it'll help with the tokenizer usage (doesn't look much changed to me but you know best)

discojs-node/src/loaders/text.ts
JulienVig: Yes the one-token overlap is to make sure that that we predict the next token of each token in the block. And what we are currently doing is fairly predicting the next token of every single token in the batch (the mess going on in the attention layer actually performs the "slide over" the block via matrix operations)
This video bypasses the tokenization complexity by predicting at the character-level (i.e. the next letter rather than the next token). In any case, the video ends up by implementing the batches [similarly to us](https://github.com/karpathy/ng-video-lecture/blob/52201428ed7b46804849dea0b3ccf0de9df1a5c3/gpt.py#L41) by yielding batches with one-token overlap.
Similarly a more recent video/repository addresses the tokenization and also yields batches with overlap https://github.com/karpathy/build-nanogpt/blob/631f7d6f080effc4342b5e2f2f41879642c76fd1/train_gpt2.py#L184
Is it clearer or more confusing?

discojs/src/models/gpt/layers.ts
JulienVig: I don't think so... At least this is only for debugging purposes.

discojs-node/src/loaders/text.ts
JulienVig: batchSize? not really it's another dimension of the model input, ideally we input a matrix [blockSize, batchSize] to the model so knowing both is useful to guess a good chunk size

discojs/src/models/gpt/model.ts
JulienVig: top k first filters out every value except the first k and then use softmax to get a probability distribution (so all the probability mass distributed over the tokens dropped is being reallocated on the first top k). I include top-k sampling because it's a very common strategy in language generation

discojs-node/src/loaders.spec.ts
JulienVig: Okay I tried with some Llama tokenizers but the tokenization yields inconsistent tokens with the playgrounds I agree (which agree with each other).
I created an [issue upstream](https://github.com/huggingface/transformers.js/issues/1019) but not sure if they maintain the pretrained tokenizers

docs/CONTRIBUTING.md
JulienVig: Oh nice to good to know! But here we always want to run this test. I'm duplicating the test to have one always run without recording and one with recording when manually triggered.

discojs-node/src/loaders.spec.ts
JulienVig: I'm not sure if you're asking for some changes or if you're saying it is good as it is?

discojs-node/src/loaders.spec.ts
JulienVig: Ah the comment wasn't describing what was being tested but explaining why we shift by blocksize + 1, porr choice of words with "should have" I'm rephrasing the comment

discojs-node/src/loaders.spec.ts
JulienVig: I moved out some things but I'm not sure that was all of what you were talking about

discojs-node/src/loaders.spec.ts
JulienVig: I'm adding a test case with the Llama 3 tokenizer by comparing with the python tokenizer rather than using the playgrounds.

discojs-node/src/loaders.spec.ts
tharvik: that's not super important but changing it to have a fixed list of token would be safer. so yeah, please do change it :)

discojs-node/src/loaders.spec.ts
tharvik: yep, that was I wanted: the shortest the code inside `withFile`, the more readable it is for me, as it makes it clear when the file is not needed anymore.

discojs-node/src/loaders/text.ts
tharvik: ho yeah, totally right, I got my paintbrushes mixed there

martinjaggi:
maybe rename block size to context len, that would be more specific

discojs-node/src/loaders/text.ts
JulienVig: Alright, the current implementation ended up being much simpler than my original tokenizing text loaders:
```js
loadText('../datasets/wikitext/wiki.train.tokens')
  .map(text => processing.tokenize(tokenizer, text)) // tokenize the whole chunk
  .flat() // (I renamed unbatch to flat)
  .batchWithOverlap(config.blockSize) // one token overlap between each batch
  .map((tokens) => [tokens.pop(), tokens.last()] as [List<number>, number])
  .batch(batchSize);
```

webapp/src/components/dataset_input/TextDatasetInput.vue
JulienVig: Yay no need for the Task anymore

discojs-node/src/loaders/text.ts
```diff
@@ -1,14 +1,34 @@
...
 
-export function load(path: string): Dataset<Text> {
+const debug = createDebug("discojs-node:loaders:text");
+
+/**
+ * Returns chunks of text. Use `minChunkSize` to ensure that 
+ * each chunk is bigger than the expected sequence length.
+ * 
+ * @param path path to the text file to read
+ * @param minChunkSize default to 16KiB, the minimum size of each chunk in bits
```
tharvik: I don't see it being used anywhere in code, we can drop the parameter altogether

discojs/src/dataset/dataset.ts
```diff
@@ -144,8 +144,48 @@ export class Dataset<T> implements AsyncIterable<T> {
...
+   * - The last batch is dropped as there are no next element to add.
+   *
+   * This method is tailored to create text sequences where each token's label is the following token. 
+   * In order to have a label for the last token of the input sequence, we include the first token
+   * of the next sequence.
+   * 
+   * @param size batch size excluding the overlapping element, at least 1
+   * @returns a dataset batch size `size + 1`
+   */
+  batchWithOverlap(size: number): Dataset<Batched<T>> {
```
tharvik: I managed to get a slightly modified version of `batch` encompassing both use cases (except the "drop last batch if smaller than expected"). I don't feel as it burden `batch` much (and I find it a bit more readable than `batchWithOverlap`)

```diff
-  batch(size: number): Dataset<Batched<T>> {
+  batch(size: number, overlap = 0): Dataset<Batched<T>> {
     if (size <= 0 || !Number.isInteger(size)) throw new Error("invalid size");
+    if (overlap >= size || !Number.isInteger(overlap))
+      throw new Error("invalid overlap");

     return new Dataset(
       async function* (this: Dataset<T>) {
         const iter = this[Symbol.asyncIterator]();

+        let overlapped = List<T>();
         for (;;) {
           const batch = List(
-            await Promise.all(Range(0, size).map(() => iter.next())),
+            await Promise.all(
+              Range(overlapped.size, size).map(() => iter.next()),
+            ),
           ).flatMap((res) => {
             if (res.done) return [];
             else return [res.value];
@@ -135,50 +141,51 @@ export class Dataset<T> implements AsyncIterable<T> {

           if (batch.isEmpty()) break;

-          yield batch;
+          yield overlapped.concat(batch);
+          overlapped = batch.takeLast(overlap);

           // iterator couldn't generate more
-          if (batch.size < size) break;
+          if (batch.size < size - overlap) break;
         }
       }.bind(this),
     );
   }
```

discojs/src/dataset/dataset.ts
```diff
@@ -176,6 +216,23 @@ export class Dataset<T> implements AsyncIterable<T> {
     );
   }
 
+  /**
+   * Repeat the dataset `times` times
+   * @param times number of times to repeat the dataset, if undefined, the dataset is repeated indefinitely
+   * @returns a dataset repeated `times` times
+   */
+  repeat(times?: number): Dataset<T> {
```
tharvik, line 208: `repeat(0)` is the same as `repeat(1)`, I would be in favor of requiring that times > 0 (and integral).

discojs/src/dataset/dataset.ts
```diff
@@ -176,6 +216,23 @@ export class Dataset<T> implements AsyncIterable<T> {
...
+    return new Dataset(
+      async function* (this: Dataset<T>) {
+        let loop = 0;
+        do {
+          for await (const e of this) yield e;
```
tharvik: you can use the `yield*` meaning to yield everything from the given iterator.

```suggestion
          yield* this;
```

discojs/src/processing/text.ts
```diff
@@ -1,47 +1,56 @@
...
+  config = { ...config }; // create a config if undefined
+  
+  if (config.padding || config.truncation) {
+    if (config.max_length === undefined) throw new Error("max_length needs to be specified to use padding or truncation");
+    if (!Number.isInteger(config.max_length))  throw new Error("max_length should be an integer");
+  }
+  
+  if (config.padding) {
+    // The padding side is set as an attribute, not in the config
+    tokenizer.padding_side = config.padding_side !== undefined ? config.padding_side : 'left'
```
tharvik:
```suggestion
    tokenizer.padding_side = config.padding_side ?? 'left'
```

discojs/src/processing/text.ts
```diff
@@ -1,47 +1,56 @@
...
+import { List } from "immutable";
 import { PreTrainedTokenizer } from "@xenova/transformers";
+import type { Text, TokenizedText } from '../index.js'
 
 function isArrayOfNumber(raw: unknown): raw is number[] {
   return Array.isArray(raw) && raw.every((e) => typeof e === "number");
 }
 
-type Token = number;
+interface TokenizingConfig {
```
tharvik, line 9: can we have a smaller config? many of theses fields are arcane to me.

discojs/src/dataset/dataset.ts
```diff
@@ -144,8 +144,48 @@ export class Dataset<T> implements AsyncIterable<T> {
...
+          // yield the current batch with the first element of the next batch
+          yield currentBatch.concat(nextFirstElement);
+          currentBatch = nextBatch
+        }
+      }.bind(this),
+    );
+  }
+
+  /** Flatten batches/arrays of elements */
+  flat<U>(this: Dataset<Batched<U>>): Dataset<U> {
```
tharvik: `flatten` is a [more](https://immutable-js.com/docs/latest@main/List/#flatten()) [standard](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.flatten) name. also, now that we changed the name, we can input `DatasetLike<U>` instead of `Batched<U>` to allow flattening `List`, arrays, other `Dataset`, …

docs/CONTRIBUTING.md
```diff
@@ -104,6 +104,60 @@ npm -w webapp test
...
+          wait-on: 'http://localhost:8081' # Waits for above
+          # Records to Cypress Cloud
+          # https://docs.cypress.io/guides/cloud/projects#Set-up-a-project-to-record
+          record: true
+        env:
+          VITE_SERVER_URL: http://server
+          CYPRESS_RECORD_KEY: ${{ secrets.CYPRESS_RECORD_KEY }}
+```
+
+Note that `CYPRESS_RECORD_KEY` is a repository secret which has already been added.
```
tharvik: as it is now a dedicated workflow, we can reference it here instead of having a codeblock.

tharvik (approved):
only cosmetic comments, thanks for the huge amount of work :tada:

discojs/src/processing/text.ts
JulienVig: I'm removing the non-essential ones but LLM generation is inherently complex (see the [HuggingFace docs](https://huggingface.co/docs/transformers/en/generation_strategies) for an overview of generation strategies)

discojs/src/dataset/dataset.ts
JulienVig: Because of code duplication with the constructor I created a common function to convert DatasetLike to AsyncGenerator, let me know if there was a simpler way to do it!

discojs/src/dataset/dataset.ts
JulienVig: Nice! Note that I had to change how we were calling batch with an overlap from `.batchSithOverlap(contextLength, 1)` to `.batch(contextLength + 1, 1)` as `batch` always returns batches of `size` elements.

JulienVig:
I'll wait for #840 and #825 to be merged before this one.

discojs/src/dataset/dataset.ts
tharvik: ho, good idea, I was reusing the constructor but a standalone function is more readable :ok_hand:

martinjaggi:
well done!

PR #825, opened 2024-11-05
Toggle graphs in the training board olena

0L3N4:
Client-side

Additions & Modifications:

1. Enhanced DropdownCard Component:
   - The DropdownCard component was refactored to allow dynamic control over its state (collapsed or expanded) based on a passed parameter. 
   - By default, the component remains open.

2. Refinements in the Training Information Section:
   - Graphs within the Training Information section were encapsulated within the newly refactored DropdownCard component. This provides users the option to expand or collapse the graphical data, improving the clarity and organization of training details.

3. Removed Training Logs from Training Information:
   - The Training Logs information was removed from the Training Information section, streamlining the display and focusing user attention on more important information.


webapp/src/components/containers/DropdownCard.vue
```diff
@@ -24,14 +24,26 @@
...
 import DownArrow from "@/assets/svg/DownArrow.vue";
 
 import IconCardHeader from "./IconCardHeader.vue";
 
+const props = defineProps({
+  initiallyOpen: {
+    type: String,
+    default: "true", //open by default
+  },
+});
```
tharvik: I personally find it clearer to use [type-only props declarations](https://vuejs.org/api/sfc-script-setup.html#type-only-props-emit-declarations). also, you can have a preciser type (only two value are possible here, true or false) and avoid `JSON.parse` at all.

```suggestion
const props = withDefaults(
  defineProps<{
    initiallyOpen?: boolean;
  }>(),
  { initiallyOpen: true },
);
```

webapp/src/components/containers/DropdownCard.vue
```diff
@@ -24,14 +24,26 @@
...
+
 const opened = ref(true);
+
+onMounted(() => {
+  opened.value = JSON.parse(props.initiallyOpen); 
+});
```
tharvik: `onMounted` [is quite special](https://vuejs.org/guide/essentials/lifecycle.html), it doesn't get triggered when data change for example. better to watch for prop to change instead, so that if someone change the prop from the outside, it will behave as expected.

```suggestion
watch(props, ({ initiallyOpen }) => {
  opened.value = initiallyOpen;
});
```

webapp/src/components/training/TrainingInformation.vue
```diff
@@ -186,6 +166,9 @@ import ModelExchangeIcon from "@/assets/svg/ModelExchangeIcon.vue";
 import ModelUpdateIcon from "@/assets/svg/ModelUpdateIcon.vue";
 import PeopleIcon from "@/assets/svg/PeopleIcon.vue";
 import Contact from "@/assets/svg/Contact.vue";
+import DropdownCard from "../containers/DropdownCard.vue";
+
+const currentOpen = ref(localStorage.getItem('initiallyOpen') || 'false');
```
tharvik: now that it is a boolean, parsing as to be updated, as close as the loading point you can. it helps with debugging because you know where the value is created. an easy parser is shown below.
also, the "or" op is quite unsafe (as many values can be falsy), better to use [a nullish coaslescing](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing) which only check for `null` and `undefined`.

```suggestion
const currentOpen = ref(localStorage.getItem("initiallyOpen") === "true" ? true : false);
```

one can even make it safer by checking for the full valid range, which would be the nicest (as it fails on unexpected values, which is always nice for finding logical bugs)
```ts
function loadCurrentOpen(): boolean {
  const raw = localStorage.getItem("initiallyOpen");
  switch (raw) {
    case "true":
      return true;
    case "false":
    case null:
      return false;
    default:
      throw new Error(`unexpected loadCurrent stored value: ${raw}`);
  }
}
const currentOpen = ref(loadCurrentOpen());
```

webapp/src/components/training/TrainingInformation.vue
```diff
@@ -51,8 +51,9 @@
       </IconCardSmall>
     </div>
 
-    <!-- Training and validation loss charts -->
-    <div
+    <DropdownCard :initiallyOpen="currentOpen" @click="toggleAdvancedInfo">
```
tharvik: consider renaming `currentOpen` to `initiallyOpen`, then you can use the short version of the attribute, which is a bit nicer

```suggestion
    <DropdownCard :initiallyOpen @click="toggleAdvancedInfo">
```

tharvik (changes requested):
thanks for the work! a few comments on how to improve it, please also fix the tests/lint found by the CI

webapp/src/components/training/__tests__/TrainingInformation.spec.ts
```diff
@@ -21,9 +21,10 @@ it("shows messages", async () => {
...
       hasValidationData: false,
       isTraining: true,
-      isTrainingAlone: false
+      isTrainingAlone: false,
     },
   });
 
-  expect(wrapper.findAll("#mapHeader li")).toHaveLength(3);
+  const listItems = wrapper.findAll("#mapHeader li");
+  expect(listItems.length === 0 || listItems.length === 3).toBe(true);
```
JulienVig: Only one length should be expected (is the advanced information open or closed by default?).  

Can you break this down into two assertions: one when the length is expected to be 0 and another when the length should be 3? You can use cypress' [.click()](https://docs.cypress.io/api/commands/click) to toggle the advanced information.

JulienVig (changes requested):
Great! The TrainingInformation's test can still be improved a bit :)

Can you resolve tharvik's comments if addressed them?

webapp/src/components/training/__tests__/TrainingInformation.spec.ts
0L3N4: This test is designed to validate the Training Logs section, which has been removed. Therefore, the test should only pass when the length is 0. However, does it make sense to keep this test at all, considering there is essentially nothing left for it to verify?

JulienVig: Indeed you can remove it altogether!

JulienVig (approved):

tharvik (approved):

JulienVig:
Hey @iamsherlocked1891, could you wait until Dec 5th to merge this PR? There a DISCO demo happening tomorrow so it would be safer to wait until it's past :)

PR #826, opened 2024-11-07
Make mnist use mean aggregation rather than secure

JulienVig:
@rabbanitw noticed that training on mnist locally was failing.
Indeed the console shows: `webapp Error: secure aggregation is currently supported for decentralized only`

Reverting the aggregation strategy to `mean` until we support `secure` aggregation with federated learning.


martinjaggi:
i edited decentralized to federated in your last sentence. is that right?

JulienVig:
Ah yes thanks!

PR #833, opened 2024-11-24
Emmanuel confusion matrix

tomasoignons:
This pull request add a confusion matrix to the testing page

image
image

The confusion matrix has been made so that it works with multiple class classfication, and not only binary classification.

This pull request modify the Validator class in the discoJS library, in order to retrieve
- If the element was correctly predicted
- The prediction
- the original label of the element

### DIFF
--- a/discojs/src/validator.ts
+++ b/discojs/src/validator.ts
@@ -15,21 +15,23 @@ export class Validator<D extends DataType> {
   async *test(
     dataset: Dataset<DataFormat.Raw[D]>,
   ): AsyncGenerator<{ result: boolean; predicted: DataFormat.Inferred[D]; truth : number }, void> {
-    const results = (await processing.preprocess(this.task, dataset))
-      .batch(this.task.trainingInformation.batchSize)
+    const preprocessed = await processing.preprocess(this.task, dataset);
+    const batched = preprocessed.batch(this.task.trainingInformation.batchSize);
+
+    const initialResults = batched
       .map(async (batch) =>
         (await this.#model.predict(batch.map(([inputs, _]) => inputs)))
-      .zip(batch.map(([_, outputs]) => outputs))
-      .map(([inferred, truth]) => ({ result: inferred === truth, predicted: inferred, truth : truth })),
-    )
-    .flatten();
+          .zip(batch.map(([_, outputs]) => outputs))
+          .map(([inferred, truth]) => ({ result: inferred === truth, predicted: inferred, truth })),
+      )
+      .flatten();
 
     const predictions = await processing.postprocess(
       this.task,
-      results.map(({ predicted }) => predicted),
+      initialResults.map(({ predicted }) => predicted),
     );
 
-    const finalResults = results.zip(predictions).map(([result, predicted]) => ({
+    const finalResults = initialResults.zip(predictions).map(([result, predicted]) => ({
       ...result,
       predicted,
     }));
--- a/discojs/src/validator.ts
+++ b/discojs/src/validator.ts
@@ -14,13 +14,13 @@ export class Validator<D extends DataType> {
   /** infer every line of the dataset and check that it is as labelled */
   async *test(
     dataset: Dataset<DataFormat.Raw[D]>,
-  ): AsyncGenerator<boolean, void> {
+  ): AsyncGenerator<{ result: boolean; predicted: DataFormat.ModelEncoded[D][1] }, void> {
     const results = (await processing.preprocess(this.task, dataset))
       .batch(this.task.trainingInformation.batchSize)
       .map(async (batch) =>
         (await this.#model.predict(batch.map(([inputs, _]) => inputs)))
           .zip(batch.map(([_, outputs]) => outputs))
-          .map(([inferred, truth]) => inferred === truth),
+          .map(([inferred, truth]) => ({ result: inferred === truth, predicted: inferred })),
       )
       .flatten();
 
--- a/discojs/src/validator.ts
+++ b/discojs/src/validator.ts
@@ -14,17 +14,27 @@ export class Validator<D extends DataType> {
   /** infer every line of the dataset and check that it is as labelled */
   async *test(
     dataset: Dataset<DataFormat.Raw[D]>,
-  ): AsyncGenerator<{ result: boolean; predicted: DataFormat.ModelEncoded[D][1]; truth : DataFormat.ModelEncoded[D][1] }, void> {
+  ): AsyncGenerator<{ result: boolean; predicted: DataFormat.Inferred[D]; truth : number }, void> {
     const results = (await processing.preprocess(this.task, dataset))
       .batch(this.task.trainingInformation.batchSize)
       .map(async (batch) =>
         (await this.#model.predict(batch.map(([inputs, _]) => inputs)))
-          .zip(batch.map(([_, outputs]) => outputs))
-          .map(([inferred, truth]) => ({ result: inferred === truth, predicted: inferred, truth : truth })),
-      )
-      .flatten();
+      .zip(batch.map(([_, outputs]) => outputs))
+      .map(([inferred, truth]) => ({ result: inferred === truth, predicted: inferred, truth : truth })),
+    )
+    .flatten();
 
-    for await (const e of results) yield e;
+    const predictions = await processing.postprocess(
+      this.task,
+      results.map(({ predicted }) => predicted),
+    );
+
+    const finalResults = results.zip(predictions).map(([result, predicted]) => ({
+      ...result,
+      predicted,
+    }));
+
+    for await (const e of finalResults) yield e;
   }
 
   /** use the model to predict every line of the dataset */
--- a/discojs/src/validator.ts
+++ b/discojs/src/validator.ts
@@ -1,4 +1,4 @@
-import type { Dataset, DataFormat, DataType, Model, Task } from "./index.js";
+import { Dataset, DataFormat, DataType, Model, Task } from "./index.js";
 import { processing } from "./index.js";
 
 export class Validator<D extends DataType> {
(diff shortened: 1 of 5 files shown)


JulienVig:
Hi @tomasoignons! Thanks for the PR! I'm super busy these days so I can maybe review it next week. 
Could you add a screenshot of how it looks like for MNIST? (when there are more categories)

tomasoignons:
I had to train my own model, because we don't have the MNIST on the models available as a simple demo, and it looks really good in my opinion : 

image

And in light mode : 
image

webapp/src/components/testing/TestSteps.vue
```diff
@@ -177,9 +205,60 @@ const visitedSamples = computed<number>(() => {
...
+  switch (props.task.trainingInformation.dataType) {
+    case "image":
+        (tested.value as Tested["image"]).map(
+          ( {output} ) => matrix[output.predicted][output.label] = matrix[output.predicted][output.label] + 1,
+        );
+        break;
+    //case "text":
+    //  return undefined;
+    case "tabular":
+      return undefined;
```
tharvik: no matrix for tabular? :cry:

webapp/src/components/testing/TestSteps.vue
```diff
@@ -177,9 +205,60 @@ const visitedSamples = computed<number>(() => {
     }
   }
 });
+
+const confusionMatrix = computed<{labels : Map<number, string>, matrix : number[][]} | undefined>(() => {
+  if (tested.value === undefined) return undefined;
+  const labels = new Set<number>();
+  const mapLabels = new Map<number, string>();
```
tharvik: better to use the `Map` and `Set` implementation from [`immutable`](https://immutable-js.com/). avoiding mutability quite simplify reasoning and avoid a lot of data update error (no out-of-sync, no object magically changing from wherever, …).

webapp/src/components/testing/TestSteps.vue
```diff
@@ -264,7 +343,9 @@ async function startImageTest(
         },
         output: {
           truth: label,
-          correct,
+          correct: result,
+          predicted: predicted,
+          label : truth,
```
tharvik: this made my brain exploded

```ts
{
    truth: label,
    […]
    label: truth,
}
```

discojs/src/validator.ts
```diff
@@ -14,14 +14,14 @@ export class Validator<D extends DataType> {
   /** infer every line of the dataset and check that it is as labelled */
   async *test(
     dataset: Dataset<DataFormat.Raw[D]>,
-  ): AsyncGenerator<boolean, void> {
+  ): AsyncGenerator<{ result: boolean; predicted: DataFormat.ModelEncoded[D][1]; truth : DataFormat.ModelEncoded[D][1] }, void> {
```
tharvik: `DataFormat.ModelEncoded` is meant to stay internal to discojs as it only makes sense in conjunction with a task and a model. you can pretty much obtain the same fonctionality by `zip`ing the dataset with the output of `Validator.predict` (as you can find in [`PredictSteps`]( https://github.com/epfml/disco/blob/1b361bb14e272d085527846e31a724129357407f/webapp/src/components/testing/PredictSteps.vue#L246))

webapp/src/components/testing/TestSteps.vue
```diff
@@ -142,24 +170,24 @@ const props = defineProps<{
...
   tabular: {
     labels: {
       input: List<string>;
-      output: { truth: string; correct: string };
+      output: { truth: string; correct: string, label : string };
     };
     results: List<{
       input: List<string>;
-      output: { truth: string; correct: boolean };
+      output: { truth: string; correct: boolean; predicted : number, label : number };
```
tharvik: I don't really understand why `label` was added to every field, especially the `tabular.labels.output.label`. it kinda duplicate the already existing labels (but keeping theses in the model-space rather than the human-readable-space)

webapp/src/components/testing/TestSteps.vue
```diff
@@ -177,9 +205,60 @@ const visitedSamples = computed<number>(() => {
...
+        labels.add(output.predicted);
+        mapLabels.set(output.label, output.truth);
+      });
+      break;
+    default: {
+      const _: never = props.task.trainingInformation;
+      throw new Error("should never happen");
+    }
+  }
+  const size = Math.max(labels.size, Math.max(...Array.from(labels)));
```
tharvik: the rest expansion operator (`...`) can actually work with any iterable (which `Set` is).

```suggestion
  const size = Math.max(labels.size, ...labels);
```

that being said, better to read the list of existing labels from `Task.LABEL_LIST` as to always have the same size for a given task.

webapp/src/components/testing/TestSteps.vue
```diff
@@ -58,6 +58,34 @@
...
 
+    <div v-if="confusionMatrix && confusionMatrix.matrix.length > 0" class="p-4 mx-auto lg:w-1/2 h-full bg-white dark:bg-slate-950 rounded-md">
+      <h4 class="p-4 text-lg font-semibold text-slate-500 dark:text-slate-300">
+        Confusion Matrix
+      </h4>
+      <table class="min-w-full divide-y divide-slate-600 dark:divide-slate-400 text-center">
+        <thead>
+          <tr>
+            <th class="pl-6 py-3 text-xs font-medium text-gray-800 dark:text-gray-200 uppercase tracking-wider text-center border-r-gray-600 dark:border-r-gray-400 border-r-2 diagonal-header">                  
+              <span class="">Label \ Prediction</span>
```
tharvik: no style applied and already a standalone element
```suggestion
              Label \ Prediction
```

webapp/src/components/testing/TestSteps.vue
```diff
@@ -58,6 +58,34 @@
...
+            </th>              
+            <th v-for="(label, index) in confusionMatrix.matrix[0]" :key="'header-' + index" class="px-6 py-3 text-xs font-medium text-gray-800 dark:text-gray-200 uppercase tracking-wider">
```
tharvik: I feel as the cells are too big, too much space around the various element. consider removing much of the `px-6 py-3` classes on the various `th` & `td` to make it more tightly packed which would be more readable and scales much better on smaller screen.

webapp/src/components/testing/TestSteps.vue
```diff
@@ -177,9 +205,60 @@ const visitedSamples = computed<number>(() => {
...
+      break;
+    case "text":
+      return undefined;
+    case "tabular":
+      (tested.value as Tested["tabular"]).results.forEach(({ output }) => {
+        labels.add(output.label);
+        labels.add(output.predicted);
+        mapLabels.set(output.label, output.truth);
+      });
+      break;
```
tharvik: no need to compute for tabular if we don't support it in the end

tharvik (changes requested):
thanks for the nicer test visualisation, love a table filling up of data!
I've got a few comments on how it can be tidier but the overall feature is great!

I also got a bunch of errors toaster at the start of the test:`TypeError: matrix[output.predicted] is undefined`

webapp/src/components/testing/TestSteps.vue
tomasoignons: It is added (I tested it with the titanic Discollaborative)

webapp/src/components/testing/TestSteps.vue
tomasoignons: I used another way of computing the labels, because in fact they were given in the task, I just didn't know where to search
Even when i did my weird thing, I had the feeling it was somewhere else. And I think this also causes the `TypeError : matrix[output.predicted] is undefined`.

webapp/src/components/testing/TestSteps.vue
tomasoignons: Yeah, it was really weird, I redefined it correctly.
I did it at the beginning to make things work without breaking anything, but I didn't realize I didn't changed it...

webapp/src/components/testing/TestSteps.vue
tomasoignons: I think I added them in case someone else wants to perform further analysis, it is always usefull to have the string of the category the element was initially, in my opinion. But I can remove it

webapp/src/components/testing/TestSteps.vue
tomasoignons: I added the tabular matrix now, but the big chunk of code was to have a complete list of label, which is not used anymore, because there is a better way to do so

webapp/src/components/testing/TestSteps.vue
tomasoignons: I did what I could in order to make it more compact. I don't think it is perfect, personnaly I prefer it with space, but that is for sure more readable

discojs/src/validator.ts
tomasoignons: Didn't expect it to be such a big refractor, because it makes the logic really different about how to compute the matrix (and at this point I am happy to have left the labels, because now I use them and not the numbers anymore)

tomasoignons:
I have no clue why I don't pass the timeout here, maybe my code is suboptimal, or the postprocess takes longer and I am just above, if @JulienVig or @tharvik could check this point (not review the whole merge request, just check why the test fails, because I try random things, but the test seems to work fine until the last bit) if would be perfect.

tharvik:
> I have no clue why I don't pass the timeout here

ho, yeah, some integration tests are a bit flaky, fell free to retry the CI/CD workflow if it doesn't look like it's coming from your code.
in this specific case, I think that it comes from the postprocess pass which wasn't needed in the previous implementation, so it's not the test flakiness. you can double the timeout of the validator/"can read and predict randomly on simple_face" test (we only care that it finishes, not that it does it fast, we would need benchmark for that)

tharvik (approved):
oupsi, I kinda forgot about reviewing this PR; to make up for it, I added a few changes instead of asking for more changes.

anyway, thank for the work, that's way more useful than simply having a growing percentage!

PR #837, opened 2024-11-28
FedProx implementation

### LINKED ISSUE #802: Implement FedProx to improve robustness to heterogeneity
Following @rabbanitw's comment, extending our federated averaging methods to support [FedProx](https://arxiv.org/pdf/1812.06127) would greatly improve Disco's robustness to client heterogeneity (data and system).

In short, FedProx adds a regularizing term (called "proximal term") to the local objective function. 
From the paper:
> The proximal term is beneficial in two aspects: 
(1) It addresses the issue of statistical heterogeneity by restricting the local updates to be closer to the initial (global) model without any need to manually set the number of local epochs.
(2) It allows for safely incorporating variable amounts of local work resulting from systems heterogeneity

The algorithm is summarized here. 
Screenshot 2024-10-08 at 17 30 05
Disco currently doesn't select a subset of users and relies either on all users or on the first ones that reply with local updates.

JulienVig:
Closes #802


JulienVig:
@tomasoignons I rebased the branch onto the latest develop commit. Here is a breakdown of the current commits:
1. tfjs allows training via multiple functions: fit, fitDataset, trainOnBatch. We were using fitDataset on each batch so I switched to using trainOnBatch to be consistent and simplify things
2. We used to compute logits twice for each update, once for the loss and once for the accuracy. This is very inefficient but not super practical to fix. I implemented one but feel free to suggest a better refactoring
3. A temporary commit, I override the training loss by implementing it by hand to then implement the fedprox loss. This commit saves the codebase state where the custom training loss behaves the same as the default loss (make sure this is the case)
4. Fedprox loss building on the previous commit, I didn't have the time to test it in depth so this is up to you

As we were discussing, the fedprox sketch has been hard-coded and doesn't allow users to choose which algorithm they want between fedavg and fedprox, and potential future algorithms like Mime.

tharvik:
added a few commits of cleanup:
* cli was using an old version of `immutable` (4 vs 5), removing it to use the global one
* there was some cli local "node_modules" which shouldn't exist
I suspect that it happened during a merge commit

it is still failing but rebasing/merging upon #903 (or "develop" when merged) should allow it to pass.

JulienVig:
Not planned anymore

PR #840, opened 2024-12-02
GDHF demo

JulienVig:
Implements a DISCO demo for GDHF with the help of @Mastel22 and @tharvik. The demo consists in binary dog cuteness classification

### Running the demo locally
Download the [dataset zip](https://storage.googleapis.com/deai-313515.appspot.com/tinder_dog.zip)
Setup the project
```
git clone git@github.com:epfml/disco.git
cd disco
git switch gdhf # go to the gdhf branch
npm ci && npm -ws run build # install and build the project
```
3. In the meantime, unzip the dataset into the `datasets/` folder
4. When the project is built, start the server, the webapp and the CLI in different tabs
```
npm -w server start # 1st terminal

npm -w webapp start # 2nd terminal

npm -w cli start # 3rd terminal
```
5. Go to the webapp in your browser: `http://localhost:8081/`
6. Go to the GDHF task
7. Connect the images in the 0th split: `datasets/tinder_dog/0` by CSV (not by group)
8. Click on Train
9. When the training is done click on next to test the model
10. You can use the data in the `test` folder

### Running the demo on discolab.ai
At step 4, add --host option when starting the CLI:
```
npm -w cli start -- --host https://disco-zbkj3i466a-oa.a.run.app/
```
At step 5, go to [discolab.ai](https://discolab.ai)

https://github.com/user-attachments/assets/871cd645-3f6b-493a-8237-1de01867b30f


cli/src/args.ts
```diff
@@ -22,12 +23,19 @@ const argExample = 'e.g. npm start -- -u 2 -e 3 # runs 2 users for 3 epochs'
...
+    numberOfUsers: { type: Number, alias: 'u', description: 'Number of users', defaultValue: 2 },
     epochs: { type: Number, alias: 'e', description: 'Number of epochs', defaultValue: 10 },
     roundDuration: { type: Number, alias: 'r', description: 'Round duration (in epochs)', defaultValue: 2 },
     batchSize: { type: Number, alias: 'b', description: 'Training batch size', defaultValue: 10 },
     save: { type: Boolean, alias: 's', description: 'Save logs of benchmark', defaultValue: false },
+    host: {
+      type: (raw: string) => new URL(raw),
+      typeLabel: "URL",
+      description: "Host to connect to",
+      defaultValue: new URL("http://localhost:8081"),
```
tharvik: oupsi, the default value targets the webapp and not the server, I got theses mixed again :sweat_smile: 

```suggestion
      defaultValue: new URL("http://localhost:8080"),
```

cli/src/data.ts
```diff
@@ -36,10 +37,50 @@ async function loadLusCovidData(): Promise<Dataset<DataFormat.Raw["image"]>> {
...
+    const label = imgToLabel.get(img.slice(0, -4)) // remove the file extension
+    if (label === undefined) throw Error(`Image ${img} not found in CSV`)
+    return label.toString()
+  })
+  const imgPaths = imagesFile.map(imgName => path.join(folder, imgName))
+  console.log(`Found ${imgPaths.length} in split ${split}`)
+  const images = await Promise.all(imgPaths.map(imgPath => loadImage(imgPath)))
+  
+  return new Dataset(images).zip(labels)
+}
```
tharvik: we can use `loadCSV` and some processing to reduce the memory footprint (no need to load images before creating the dataset) and make it somewhat more readable IMO. it also drops the `csv-parse` deps from the cli.

```suggestion
function loadTinderDogData(split: number): Dataset<DataFormat.Raw["image"]> {
  const folder = path.join("..", "datasets", "tinder_dog", `${split + 1}`);

  return loadCSV(path.join(folder, "labels.csv"))
    .map(
      (row) =>
        [
          processing.extractColumn(row, "filename"),
          processing.extractColumn(row, "label"),
        ] as const,
    )
    .map(async ([filename, label]) => {
      try {
        const image = await Promise.any(
          ["png", "jpg", "jpeg"].map((ext) =>
            loadImage(path.join(folder, `${filename}.${ext}`)),
          ),
        );
        return [image, label];
      } catch {
        throw Error(`${filename} not found in ${folder}`);
      }
    });
}
```

datasets/.gitignore
```diff
@@ -17,3 +17,6 @@
 
 # LUS Covid
 /lus_covid/
+
+# GDHF demo
+/tinder_dog/
```
tharvik, line 22: is there a way to download the dataset via `populate`? I see that in the task definition, there is a [link](https://storage.googleapis.com/deai-313515.appspot.com/tinder_dog.zip) to do so and seems to work.

discojs/src/default_tasks/tinder_dog.ts
```diff
@@ -0,0 +1,84 @@
...
+        aggregationStrategy: 'mean',
+        minNbOfParticipants: 3,
+        tensorBackend: 'tfjs'
+      }
+    }
+  },
+
+
+  async getModel(): Promise<Model<'image'>> {
+    const seed = 42
```
tharvik: maybe add a comment on this "cheat" so that we remove it later

webapp/src/components/testing/TestSteps.vue
```diff
@@ -78,7 +78,7 @@
               class="font-bold"
               :class="result.output.correct ? 'text-green-500' : 'text-red-700'"
             >
-              {{ result.output.truth }}
+              {{ result.output.truth.toUpperCase() }}
```
tharvik: okay, maximum nitpicking mode; CSS is in charge of styling the elements organized by HTML, here uppercasing is a way to format data, not change its meaning, so it's better to use [the `uppercase` class of tailwindcss](https://tailwindcss.com/docs/text-transform)

webapp/src/components/testing/PredictSteps.vue
```diff
@@ -73,7 +73,7 @@
           :image="result.input.image"
         >
           <template #title>
-            <span class="font-bold">{{ result.output }}</span>
+            <span class="font-bold text-slate-600">{{ result.output.toUpperCase() }}</span>
```
tharvik: see uppercase nitpicking (the idea behind it is also that is reads more easily with assistive technologies, but yeah, a screenreader on disco would probably intensely fail)

discojs/src/default_tasks/tinder_dog.ts
```diff
@@ -0,0 +1,84 @@
...
+        dataFormatInformation: 'Accepted image formats are .png .jpg and .jpeg.',
+        dataExampleText: '',
+        dataExampleImage: 'https://storage.googleapis.com/deai-313515.appspot.com/tinder_dog_preview.png',
+        sampleDatasetLink: 'https://storage.googleapis.com/deai-313515.appspot.com/tinder_dog.zip',
+        sampleDatasetInstructions: 'Opening the link should start downloading a zip file which you can unzip. To connect the data, pick one of the data splits (the folder 0 for example) and use the CSV option below to select the file named "labels.csv". You can now connect the images located in the same folder.'
+      },
+      trainingInformation: {
+        epochs: 10,
+        roundDuration: 2,
+        validationSplit: 0,
```
tharvik: no validation? temporary for a nicer demo? worth a comment

webapp/src/components/dataset_input/LabeledImageDatasetInput/ByGroup.vue
```diff
@@ -67,7 +67,13 @@ function refreshWatcher() {
...
           files.value?.map((f) => [label, f] as const)?.toArray() ?? [],
       );
-
+      // shuffle the filenames o.w. they are ordered by labels
+      for (let i = 0; i < expanded.length; i++) {
+        const j = Math.floor(Math.random() * i)
+        const swap = expanded[i]
+        expanded[i] = expanded[j]
+        expanded[j] = swap
+      }
```
tharvik, line 77: that shouldn't influence the training, it's for a nicer output in the prediction yes?
if that's the case, consider moving the shuffling there

tharvik (approved):
all ready for the demo, let's go!
very few non-blocking changes

webapp/src/components/dataset_input/LabeledImageDatasetInput/ByGroup.vue
JulienVig: No it actually makes the training fail if the data stays ordered by labels. Batches are created sequentially, and because we don't implement any dataset-level shuffling (only batch-level), most of the batches contain a single class which makes the model updates too wild to learn anything. 
I implemented this file shuffling as a temporary placeholder until we have some kind of dataset shuffling, I'll add a comment

discojs/src/default_tasks/tinder_dog.ts
JulienVig: Ah yes it's indeed for a nicer demo, I'll add a comment

webapp/src/components/testing/TestSteps.vue
JulienVig: Haha this is a fair remark

datasets/.gitignore
JulienVig: Oh yes I forgot, added

cli/src/data.ts
JulienVig: Ah yes using `loadCSV` is much better

PR #845, opened 2024-12-09
Redesign of homepage

0L3N4:
Rework the home page to include more information of Disco.

#### Before

Screenshot 2025-02-17 at 12 49 17
Screenshot 2025-02-17 at 12 49 17

#### After

https://github.com/user-attachments/assets/a3433425-a38c-454b-955e-49635a937d8b

### Features

After discussing with @MalloryHenriet, here are some ideas to consider:
- [x] disco font for headings
- [x] only use 2-3 font size and use it consistently
	* A good body size is "Build and train"
- [x] center the website to make reading vertical rather than horizontal
	* "Data is never shared" should go under the heading
- [x] the white text on the orange button doesn't have enough contrast
	* look at color.review
- [x] round buttons in Disco and use Disco font
- [x] round the Task list tags and use slack icon colors
- [x] try left-aligning logo in home page
- [x] align "3 steps" heading with the following elements
- [x] make logo smallers
- [x] Use "Lab" and "University" headers between logos
- [ ] change the background to dark blue for the logo section
- [x] Add a header and more spacing for the CTA button at the bottom of the page
- [x] Add outlines to the federated learning animations
- [x] Put give us feedback in footer
        * removing feedback form for now since it hasn't been used and we provide contact details
- [x] Move dark theme toggle at the bottom of the sidebar or the footer
- [x] Display names to the sidebar icons
- [x] Hover on card can only show outline, no zoom
- [x] make main container narrower and potentially show information in a column on the right
- [ ] make drag and drop zone green to show it is secure and make it blue on hover

### Other Changes:

* The "Info Section" icon has been removed from the sidebar
* Deleted unused icons and renamed icons to multi-word names to comply with Vuejs convention
* Renamed multiple component files to multi-word names
  * Notably `Testing.vue` has been renamed to `ModelLibrary.vue` and `Trainer` to `TrainerDashboard`
* Define webapp-wide default font size and color in App.vue
* Define theme colors in tailwind.config.ts


webapp/src/assets/svg/DiscoLogoBoule.vue
tharvik, line 1: "Boule" is very français, using various languages in a codebase makes it harder for people not using every language used in it to work with. rename it to smth like "DiscoLogo" (as it is not only the ball).

webapp/src/assets/svg/MLOLogo.vue
tharvik, line 1: in dark mode, we can't read the subtitle, only a kinda omnious "optimization" shows up.

webapp/src/assets/svg/LightLogo.vue
```diff
@@ -0,0 +1,64 @@
...
+      </g>
+    </g>
+  </svg>
+</template>
+
+<script lang="ts">
+export default {
+  name: "LightLogo",
+};
+</script>
```
tharvik: the name isn't different from [the one derived from the file](https://vuejs.org/api/options-misc.html#name).
```suggestion
```

webapp/src/assets/svg/DiscoLogoBoule.vue
```diff
@@ -0,0 +1,75 @@
...
+          </g>
+        </g>
+      </g>
+    </g>
+  </svg>
+</template>
+
+<script lang="ts">
+export default {
+  name: "DiscoLogoBoule",
```
tharvik: name derived from filename

```suggestion
```

webapp/src/assets/svg/DiscoLogoBoule.vue
```diff
@@ -0,0 +1,75 @@
...
+  props: {
+    customClass: {
+      default: "bi bi-info-circle w-7 h-7",
+      type: String,
+    },
+    viewBox: { default: "-1 -1 18 18", type: String },
```
tharvik: both seems unused, rm (and with the previous comment, it allows to remove the whole script block)

webapp/src/assets/svg/MLOLogo.vue
```diff
@@ -0,0 +1,225 @@
...
+          style="fill-rule: nonzero"
+        />
+      </g>
+    </g>
+  </svg>
+</template>
+
+<script lang="ts">
+export default {
+  name: "MLOLogo",
```
tharvik: no different from filename

webapp/src/assets/svg/MLOLogo.vue
```diff
@@ -0,0 +1,225 @@
...
+  props: {
+    color: {
+      type: String,
+    },
+  },
+};
+</script>
```
tharvik: we prefer to use the [composition API](https://vuejs.org/guide/introduction.html#api-styles) which has [numerous advantages when used with single file components (SFC)](https://vuejs.org/api/sfc-script-setup.html). this block can be rewriten to
```suggestion
<script lang="ts" setup>
defineProps<{
  color: string;
}>();
</script>
```

webapp/src/assets/svg/RealTimeCollaboration.vue
```diff
@@ -0,0 +1,375 @@
...
+<script lang="ts">
+export default {
+  name: "RealTimeCollaboration",
+  props: {
+    color: {
+      type: String,
+    },
+  },
+};
+</script>
```
tharvik: name from filename and unused "color"

```suggestion
```

webapp/src/components/home/Home.vue
```diff
@@ -1,45 +1,189 @@
...
         <span class="font-disco text-disco-blue font-semibold">CO</span
         >llaborative Learning
       </span>
     </div>
-    <div class="flex h-full flex-col justify-center">
-      <Landing v-show="!started" @click="started = true" />
-      <GetStarted v-show="started" />
+    <div class="px-4 pt-20 sm:pt-28 text-justify lg:pb-14 lg:pt-0">
+      <p
+        class="text-lg sm:text-2xl lg:text-4xl text-slate-600 dark:text-slate-300"
```
tharvik: it's weird that depending on the screen size, the title changes size but not the rest of the text. it looks "less important" when viewed on smaller screen. I'm not sure of to better fix that, but keeping it coherence between the parts would be nice. while at it, the bottom part ("Why use disco") hasn't the same font size either.

ho, btw, you can also merge the `div` into `p`, it doesn't change the layout

webapp/src/components/home/Home.vue
```diff
@@ -1,45 +1,189 @@
...
+        Enables collaborative and privacy-preserving training of machine
+        learning models
+      </p>
+    </div>
+  </div>
+
+  <!-- Information -->
+
+  <div
+    class="flex flex-col space-y-10 mt-8 text-justify py-3 sm:text-2xl md:text-2xl lg:text-2xl xlg:text-3xl"
```
tharvik: texts in here are rendered as black, better to use the `text-slate-*` classes to use the same color in the same context.

note: there is a bunch of incistencies in the UI, font color/size, gaps, margin, … a nicer way to ensure consistency would be to share CSS classes/var via webapp/src/assets/css/styles.css and add `<style scoped>` block to components which need it (and drop tailwindcss..). that would be much simpler for newcomers to help with styling; but yeah, much to do and not too much time, so I'm commenting where I can :shrug:

tharvik: consider using `grid` & `grid-cols-1 xl:grid-cols-2` instead of `flex`. I find that makes it nicer on very large screen.

webapp/src/components/home/Home.vue
```diff
@@ -1,45 +1,189 @@
...
+          class="text-xl text-justify py-3 sm:text-2xl md:text-2xl lg:text-2xl xlg:text-3xl"
+        >
+          Building upon the same principles as in federated learning,
+          decentralized learning achieved allows collaboration and data privacy
+          without the need for a central coordinator. Updates are shared purely
+          via peer2peer communication. Disco puts users in control of the entire
+          collaborative training process, without a central point of failure.
+        </p>
+      </div>
+    </div>
```
tharvik: a bunch of theses uses custom padding, margin, gap-x and other "by hand layout" classes which I think can be greatly simplified by some well though mix of the top level div with `gap-10` and consistent margins on every childs.
I recommend reading some documentation on [flexbox](https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/CSS_layout/Flexbox) and [grid](https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/CSS_layout/Grids), that helped me understand a lot about layout in CSS
on firefox, you can also inspect the page and show the flex & grid constraints directly on the page (by clicking on the "flex" button next to the element), which is very useful to see why some elements are missbehaving.

webapp/src/components/home/Home.vue
```diff
@@ -1,45 +1,189 @@
...
+
+    <div
+      class="flex flex-col justify-center mx-2 lg:my-8 text-xl text-justify py-3 sm:text-2xl md:text-2xl lg:text-2xl xlg:text-3xl"
+    >
+      <h6
+        class="font-bold mx-auto mb-8 text-justify leading-none text-xl tracking-wider sm:text-2xl md:text-3xl lg:text-3xl xlg:text-4xl"
+      >
+        Why use Disco?
```
tharvik: this title is describing a category, but it's the first time we see one looking like this (centered, without a body) (apart from the disco logo). consider either adding a title to the previous part, or make it "flow" more with the previous part.

webapp/src/components/home/Home.vue
```diff
@@ -1,45 +1,189 @@
...
+          <span class="text-lg">
+            Disco runs in the browser, from any device equipped with a modern
+            browser. For developers, we also offer a standalone node.js version,
+            not needing any browser.
+          </span>
+        </div>
+      </div>
+    </div>
+    <div class="flex flex-col justify-center">
+      <!-- <Landing v-show="!started" @click="started = true" /> -->
```
tharvik: dead comment
```suggestion
```

webapp/src/assets/svg/MLOLogo.vue
tharvik: ho right, I see that you added the fill color in code but it doesn't work when switching the theme after the mounting the component. that's why people are not reading the localStorage directly but are [using a store to do that](https://pinia.vuejs.org/). we have a few around, for the tasks, the models, look into webapp/src/store and feel free to create your own for the user settings (which would only contain the theme for now).

tharvik (commented):
thanks for the work, that makes a way more pro homepage, love it!

a few comments, mostly on style, aside from the text consistency on the homepage nothing blocking

while at it, you can also remove `src/assets/{gif/DiscoGIF,svg/Disco,AboutUsIcon}.vue` now that it is unused

webapp/src/assets/logos/ONNXLogo.vue
```diff
@@ -0,0 +1,162 @@
...
+</script>
+
+
+<style scoped>
+  .st0{fill:v-bind('color');}
+  .st1{fill:#545554;}
+  .st2{fill:#3E3E3D;}
+  .st3{fill:#4D4C4D;}
+  .st4{fill:#5D5D5D;}
+  .st5{fill:#717272;}
```
tharvik: there are all overriden by `style` attribute

```suggestion
```

webapp/src/assets/logos/getThemeColor.ts
tharvik, line 1: hoo, good idea to have a clear way to select color based on theme!

IMO, we can move it directly into the store, it might be useful for any consumer component of the theme

edit: see related store comments

webapp/src/components/pages/AboutUs.vue
```diff
@@ -1,21 +1,25 @@
...
           <a
-          class="text-blue-500 hover:underline"
-          href="https://www.epfl.ch/labs/mlo/"
-          target="_blank"
+            class="text-blue-500 hover:underline"
+            href="https://www.epfl.ch/labs/mlo/"
+            target="_blank"
           >
-            MLO
+            <span>MLO</span>
```
tharvik: why the span? it is a single word and no specific style is applied

```suggestion
            MLO
```

webapp/src/components/pages/AboutUs.vue
```diff
@@ -26,19 +30,19 @@
         <li>
           <a
             class="text-blue-500 hover:underline"
-            href="https://www.yale-light.org/"
+            href="https://www.light-laboratory.org/"
             target="_blank"
           >
-          LiGHT
+            <span>LiGHT</span>
```
tharvik: same span argument

```suggestion
            LiGHT
```

webapp/src/store/theme.ts
```diff
@@ -0,0 +1,17 @@
...
+  function toggleTheme() {
+    const newTheme = currentTheme.value === "light" ? "dark" : "light";
+    currentTheme.value = newTheme;
+  }
+  return { currentTheme, toggleTheme }
+}, {
+  persistedState: {
+    persist: true
+  }
+});
```
tharvik: small rewrite of store
 * mv getColor direclty here
   * return computed ref instead of using `watch`
 * `currentTheme` -> `current`
   * avoid the double "theme" when `useThemeStore().currentTheme`
 * inline `toogleTheme` as used only once
 
```suggestion
import { defineStore } from "pinia";
import type { Ref } from "vue";
import { computed, ref } from "vue";

export const useThemeStore = defineStore(
	"theme",
	() => {
		const current = ref<"light" | "dark">("light");

		function getColor(lightColor: string, darkColor: string): Ref<string> {
			return computed(() =>
				current.value === "light" ? lightColor : darkColor,
			);
		}

		return { current, getColor };
	},
	{ persistedState: { persist: true } },
);
```

webapp/src/components/sidebar/SideBar.vue
```diff
@@ -28,52 +30,47 @@
...
 const toggleDarkMode = () => {
-  const newTheme = currentTheme.value === 'light' ? 'dark' : 'light';
-  document.documentElement.classList.toggle('dark', newTheme === 'dark');
-  localStorage.setItem('theme', newTheme);
-  currentTheme.value = newTheme;
+  themeStore.toggleTheme();
+  document.documentElement.classList.toggle("dark", themeStore.currentTheme === "dark");
 };
-
-
```
tharvik: this is not the correct place to actually add the "dark" class, it would make more sense to have it in `App`. smth like

```vue
<template>
  <div :class="[themeClass]">
```
```ts
const themeStore = useThemeStore();
const themeClass = computed(() =>
  themeStore.current === "dark" ? "dark" : undefined,
);
```

webapp/src/components/pages/Home.vue
```diff
@@ -0,0 +1,271 @@
...
+            class="
+            mt-4 px-6 py-3 min-w-[8rem]
+            text-xl text-white tracking-wide
+            bg-orange-400 rounded-xl duration-200
+            hover:scale-105 transform
+            drop-shadow-[0px_0px_5px_rgba(251,146,60,0.5)]
+            hover:drop-shadow-[0px_0px_10px_rgba(251,146,60,0.5)]"
+          >
+            Start training
+          </button>
```
tharvik: better to use a `RouterLink`

```suggestion
          <RouterLink
            to="/list"
            class="
            mt-4 px-6 py-3 min-w-[8rem]
            text-xl text-white tracking-wide
            bg-orange-400 rounded-xl duration-200
            hover:scale-105 transform
            drop-shadow-[0px_0px_5px_rgba(251,146,60,0.5)]
            hover:drop-shadow-[0px_0px_10px_rgba(251,146,60,0.5)]"
          >
            Start training
          </RouterLink>
```

webapp/src/components/pages/Home.vue
```diff
@@ -0,0 +1,271 @@
...
+            type="button"
+            @click="goToTaskList()"
+            class="
+            mt-4 px-6 py-3 min-w-[8rem]
+            text-xl text-white tracking-wide
+            bg-orange-400 rounded-xl duration-200
+            hover:drop-shadow-[0px_0px_10px_rgba(251,146,60,0.5)]"
+          >
+            Get started
+          </button>
```
tharvik:
```suggestion
          <RouterLink
            to="/list"
            class="
            mt-4 px-6 py-3 min-w-[8rem]
            text-xl text-white tracking-wide
            bg-orange-400 rounded-xl duration-200
            hover:drop-shadow-[0px_0px_10px_rgba(251,146,60,0.5)]"
          >
            Get started
          </RouterLink>
```

webapp/src/components/pages/Home.vue
```diff
@@ -0,0 +1,271 @@
...
+          <a
+            href="mailto:discolab@listes.epfl.ch"
```
tharvik: here there is the mailto (which can be machine read) but not in the AboutUs page

webapp/src/components/pages/Home.vue
```diff
@@ -0,0 +1,271 @@
...
+            @click="goToTaskList()"
+            class="
+            mt-4 px-6 py-3 min-w-[8rem]
+            text-xl text-white tracking-wide
+            bg-orange-400 rounded-xl duration-200
+            hover:scale-105 transform
+            drop-shadow-[0px_0px_5px_rgba(251,146,60,0.5)]
+            hover:drop-shadow-[0px_0px_10px_rgba(251,146,60,0.5)]"
+          >
+            Start training
```
tharvik: the buttons in this page not of the same style as the others place in the app (blue, uppercase, less rounded, …). is it wanted? shall we add an issue on keeping a button style coherence?

webapp/src/components/pages/Home.vue
```diff
@@ -0,0 +1,271 @@
...
+        <span
+          class="text-xl sm:text-2xl md:text-3xl lg:text-4xl text-slate-600 dark:text-slate-300"
+        >
+          <span class="font-disco text-disco-cyan font-semibold">DIS</span
+          >tributed
+          <span class="font-disco text-disco-blue font-semibold">CO</span
+          >llaborative Learning
+        </span>
+      </div>
+      <div class="text-center md:text-left mt-24">
```
tharvik: while the rest of the page ~grow upon large screen, the header only does so when reaching medium screen, which is a bit weird

```suggestion
      <div class="text-center lg:text-left mt-24">
```

tharvik (approved):
superbe! this look super pro and very straightforward! :star_struck: thanks to you both!

some nitpicking comments (I'm getting lost into reviews theses days)

webapp/src/components/pages/AboutUs.vue
JulienVig: Otherwise the line breaks introduce whitespace around the word (" MLO "), another fix is to remove the line break but I find it less legible: 
```
<a
href>"..."
>MLO</a>
```
Do you have a preference?

webapp/src/store/theme.ts
JulienVig: Very nice!

webapp/src/components/sidebar/SideBar.vue
JulienVig: Yea I guess @tomasoignons did it here because it's where the toggle theme button is located, I'll move it to App

webapp/src/components/pages/Home.vue
JulienVig: Yes it was intentional to make them standout, these buttons are mostly for the look so I wanted them flashy while the others are more functional. I'll at least make the two orange buttons on this page identical. The Contact us button is outlined to show a clear importance hierarchy between the two.
I think the home page is often standing out (= slightly different) from the rest of a website so I'm not sure it's worth creating components out of those buttons. Though I'll make the specific orange shade a tailwind class

webapp/src/assets/logos/ONNXLogo.vue
JulienVig: This was in the original svg from wikipedia 🤨

webapp/src/components/pages/AboutUs.vue
JulienVig: On second thoughts I'll go with removing the line breaks, EPFL and AriadneLabs are already this way and it's not that bad

webapp/src/components/pages/Home.vue
JulienVig: to be clear, the concrete changes are:
- create CTAButton component (call-to-action button) for the two orange buttons in the landing page
- create tailwind class disco-orange to replace orange-400

webapp/src/components/containers/IconCard.vue
tharvik, line 1: ha yes, I also [changed that in my PR](https://github.com/epfml/disco/blob/690-fix-creation-form-tharvik/webapp/src/components/containers/IconCard.vue), even got rid of the `fillSpace`/`halfWidth`. maybe that's something that works here?
usually, parent element should be responsible for correctly sizing their children. the latter should "only" take care of showing their content. if the parent element want to make it bigger, it can show it in a `flex` to do so

webapp/.eslintrc.cjs
tharvik, line 1: hoo, great, I never thought that this would go away :tada:

webapp/src/components/simple/CTAButton.vue
```diff
@@ -0,0 +1,15 @@
...
+<template>
+  <RouterLink
+    to="/list"
+    class="
+      px-6 py-3 min-w-[8rem]
+      text-xl text-white tracking-wide
+      bg-disco-orange rounded-xl duration-200
+      hover:scale-105 transform
+      drop-shadow-[0px_0px_5px_rgba(251,146,60,0.5)]
+      hover:drop-shadow-[0px_0px_10px_rgba(251,146,60,0.5)]"
```
tharvik: hoo woaw, glowyy :sparkles:

webapp/src/router/router.ts
```diff
@@ -2,67 +2,42 @@ import createDebug from "debug";
...
 import AboutUs from '@/components/pages/AboutUs.vue'
 
 const debug = createDebug("webapp:router");
 
 const router = createRouter({
   history: createWebHashHistory(),
+  scrollBehavior(_to, _from, _savedPosition) {
+    // If the route is the same, do nothing
+    // this happen when toggling between dark and light mode
+    if (_to.path === _from.path) return;
```
tharvik: ho, yeah that's weird: the toggling shouldn't be a RouterLink. made a small rewrite of SidebarButton

```vue
<template>
  <div
    class="p-2 text-body-secondary-light dark:text-body-secondary-dark hover:cursor-pointer"
    v-tippy="{ content: text, placement: 'right', delay: 0, animation: 0 }"
  >
    <span class="sr-only"> {{ text }} </span>

    <slot />
  </div>
</template>

<script setup lang="ts">
defineProps<{ text: string }>();
</script>
```

now the component consumers can choose to add or not a RouterLink

webapp/tailwind.config.ts
tharvik, line 1: hoo yeah, superbe, a coherent theme :raised_hands:

webapp/src/components/App.vue
```diff
@@ -57,14 +58,17 @@
...
 const debug = createDebug("webapp:App");
 
 const route = useRoute()
-const tasksStore = useTasksStore()
 
+const tasksStore = useTasksStore()
 tasksStore.initTasks().catch((e) => debug("while init tasks: %o", e));
+
+const themeStore = useThemeStore();
+const themeClass = themeStore.getColor("", "dark")
```
tharvik: hum, it's is not really a color we have here, but a class; should we change `getColor` to smth like `selectOneOf`?

tharvik (approved):
thanks for the cleanup & renamings :ok_hand: 

the more I review, the more I nitpick: nothing really blocking, some nice to fix

I found some readability issues in the TrainingDashboard with the dark theme
 * the banner indicating that the training is done on a new model: very-light grey over pink
 * the labels of the graphs: black over deep blue

also, when training locally, it shows "Status: Waiting for more participants" but there isn't anyone to wait for (it not blocking as the training advances anyway)

webapp/src/components/containers/IconCard.vue
JulienVig: Yea that makes sense I'll remove the attribute.

webapp/src/components/App.vue
JulienVig: Haha I knew you'd say something about it, I'm changing it to `selectByTheme`

webapp/src/router/router.ts
JulienVig: Ah nice I hadn't found why toggling the theme was reloading the page thanks!

Issue #677, opened 2024-05-27
Implement an actual error handling mechanism


JulienVig:
Currently, errors are handled individually on the direct catch. We need a proper mechanism to communicate errors between the backend and the frontend to enable good user feedback when something fails or when users are not using DISCO in the intended way. 
For example, rather than parsing the error message to identify the error, we can subtype errors to make identifiable more reliably.

JulienVig:
This issue is getting more and more relevant to provide user feedback when an error occurs during training. We need a reliable framework to pass an error that occurs in discojs or the server to the webapp and convert it to a user-friendly error message.

Currently this is done by parsing the error message which is far from ideal:
```js
if (error.message.includes("provided in columnConfigs does not match any of the column names")) {
      // missing field is specified between two "quotes"
      const missingFields: String = error.message.split('"')[1].split('"')[0];
      toaster.error(`The input data is missing the field "${missingFields}"`);
}
```

An idea would be to give specific errors identifiers such as error numbers and associate each error number with a user-friendly error message. However this requires sharing the error number table between discojs, the webapp and the server and increase interdepency.

JulienVig:
@tharvik do you have any ideas or opinions?

tharvik:
[error handling is hard](https://www.fpcomplete.com/blog/error-handling-is-hard/).
there is no real standard, that I know of, to safely handle error in JS/TS (as in nearly all exception based languages). there is [some ideas](https://github.com/microsoft/TypeScript/issues/13219#issuecomment-1515037604) with a JS proposal to pattern match on `catch` that could help a bit in a far future.

exception based:
the best we have now is documenting the various functions with [`@throws`](https://tsdoc.org/pages/tags/throws) (but provides no compile-time checks, maybe via some eslint plugins). as you wrote, we can easily subclass `Error` with our own types, documenting when we're throwing theses subclasses, and `try...catch` at callsite.

values based:
there are some libraries, such as [`neverthrow`](https://github.com/supermacro/neverthrow), that provides safer values to force users to check that returned values are safe. this could be useful in some places where failure are excepted (opening files, read CSV, use user input, ...). as there is not language support, it's more cumbersome than in :sparkles:Rust:sparkles: but might worth a look.

so that doesn't really answers on what we can actually do in our case. I'm proposing the following
 * function likely to fail: use `neverthrow.Result` to force users to check it. at least to give some context and rethrow it.
[...]

PR #850, opened 2024-12-19
webapp: fix creation form

### LINKED ISSUE #690: fix task creation form
as we don't have test for it, task creation form driffted off.

* submit button only check fields, not actually submit
  * nothing is shown in logs, as it it isn't connected at all
* reset button & help request doesn't do anything
  * consider moving to feedback form for requesting help
* it seems that the vee-form isn't linked to any of our code
* from what I tried, we must pretty much rewrite all the components to use vee-form correctly and be able trigger events

tharvik:
see #690 for context

* webapp
  * redo form
    * split into more categories
    * make it nicer (no more loooong input field, grid layout on bigger screen)
    * greatly simplify fields implementation (no more type specific container, no dynamic form)
  * rework the internal layout of many pages to better separate concerns
    * avoid some forced sizing
    * have common `cards-gap` CSS class instead of inconsistent gap
  * nicer data example layout
  * testing
    * fix unconsistent temporary server URL
    * spin test server on another port to avoid having to kill dev-server to launch tests
* discojs 
  * mv `Task.TrainingInformation.dataType` to `Task.dataType`
    * allows `Task.DisplayInformation` to be type dependant
  * redo most of `Task.DisplayInformation`
  * define `Task` and related via zod schema, to facilitate serialization
    *  inline `getTokenizer`
  * `TaskProvider.getTask()` is async to allow `Tokenizer` to be loaded
  * add `Tokenizer` wrapping the one from :hugs: to allow to serialize it
* github: add cypress cache and misc cleanup


server/src/routes/training_router.ts
```diff
@@ -27,19 +24,10 @@ export class TrainingRouter {
...
-    * this.onNewTask is registered as a listener to taskSet, which has 2 consequences:
-    * - this.onNewTask is executed on all the default tasks (which are already loaded in taskSet)
-    * - Every time a new task and model are added to taskSet, this.onNewTask is executed on them.
-    * For every task and model, this.onNewTask creates a path /taskID and routes it to this.handle.
-    */
-    process.nextTick(() => {
-      taskSet.on('newTask',
-        async ({ task, encodedModel }) => { await this.onNewTask(task, encodedModel) }
-      )
-    })
```
JulienVig, line 42: No more `process.nextTick` 🎉

JulienVig (changes requested):
Awesome work thank you for tackling the infamous creation form!
Okay I have a LOT of ideas, comments and two issues which I'll start with. Let's clearly choose what's reasonable to do in this PR and create an issue with future work :)

0. As mentioned in slack, the inference seems to be ~4x slower now with slight pauses every 10 images.
1. When submitting the form successfully (according to the toaster), I get the error `Uncaught SyntaxError: JSON.parse: unexpected character at line 1 column 1 of the JSON data`. We should show a toaster error when something fails unexpectedly. Here are the values I used:
Screenshot 2025-02-12 at 12 07 57 
Screenshot 2025-02-12 at 12 08 06
2. Note how in the previous screenshot the width and height fields are not fully responsive
3. I would like to add some explanation for each field, maybe when hovering over an "i" or "?" icon as @tomasoignons had implemented, I'm happy to help writing some descriptions if you want.
4. I think we need to add more information for the model.json field, for example explain what is "TFJS", why we are using tensorflow.js, linking tfjs docs, where do download pre-trained models etc
5. I suggest changing "Category of data" to "Data type"
6. I think there should be a learning rate field in the model section, this is quite important and depends on the data, loss and model architecture
7. Are we validating that the loss and optimizer input by the user exist in tfjs?
8. "Number of pass over the data" ⇾ "Epochs" (and "Number of **passes** over the data" on hover). So that we use conventional names
9. "Number of dataset element to process at once" ⇾ "Batch size" (with "Number of dataset elements to process at once" on hover)
10. "Kind of network to run on" ⇾ "Collaborative algorithm"?
11. "Number of epoch before exchanging with the network" ⇾ "Number of **epochs** before exchanging with the network" 
12. Can you add asterisks or another marker on questions that are required? (and potentially writing "optional" for questions that are not) Right now we need to test each field individually to see which ones are required
13. Can we make error messages more user-friendly like "Please choose an image type"? Would it be possible to make the page scroll to the error when clicking submit? If too complicated can we print an error above/below the submit button? Otherwise there is no UI feedback when the error is outside the screen
Screenshot 2025-02-12 at 10 48 37
14. Should we create the unique identifier ourselves instead? Are we even checking that it is unique when a user submits a new task?
15. Regarding Differential Privacy it's not clear that leaving fields blank doesn't use DP, I think we should first ask whether to use DP and if yes ask for values (and make the fields required)
16. I don't think anyone can pick a good value for "Maximum weights difference between each round"... Does it need a unit or naming how the difference is computed (e.g. "per Euclidean distance")?
17. What is the "Size of the noise added to the weights"?
18. "Secure" aggregation is currently only available for Decentralized learning
19. Idem for "Secure Aggregation". Additionally it should only be available when choosing Decentralized Learning and "secure" aggregation type
20. Should "Maximum value of a single weight" be max absolute value or maximum magnitude? It's not clear to me if weight is a scalar, a vector or a tensor, isn't DP clipping the gradient rather than single weight values?
21. I notice small indents after "Differential Privacy" and "Secure Aggregations" but I don't think it's clear enough that these are headings, consider making the headings' font size bigger rather than adding an indent
22. "Expected data format" has placeholder "Images with [...]" even if the current data type is "tabular" or "text". Anyway I think this fields is more about accepted file format (e.g. for images we accept png jpeg and jpg, tabular only csv, etc) than the data content, isn't it? Though we can add both accepted file extensions and what data should look like.
23. "How to join after ?" I think having to explain this denotes that we're doing something wrong that isn't intuitive enough. But if we keep it I'd write it "Joining this task". For example (and probably future work) clicking submit successfully could go to a new page with "Your tasks has been created successfully", and two buttons: "Share this task" with a permalink and "Go to the task".
24. When clicking submit and form validation succeeds, rather than a toaster we should show a loading indicator with "Creating the task..."
25. Refreshing the page empties out the fields which can be annoying. Can we either persist them or have the browser confirm that we want to leave this page?
26. For future work, after successfully filling the form we could have the user connect data and train a model locally for a few epochs to make sure that labels, image dimensions, and would let the user see if the learning rate is okay, if the loss is adequate or diverges etc (so let them go back and forth between the form and the sample training until final validation and uploading the task)
27. For future work, it would be useful to have a preview of how the task would look like as the user fills the task title, description, sample data url etc before final validation.
28. For future work I was thinking of making the form a succession of pages rather than having everything in one page, it would make the experience smoother and less daunting and would hide how choosing the data type changes subsequent questions. I'm attaching a random example that I think looks nice:

https://github.com/user-attachments/assets/37d1eb17-1c17-411b-acd1-95a6da662ac1

tharvik:
that's finally it :tada: most comments handled, the remaining is sitting at #921 to do another time.
this is a **huge** PR, sorry about that, this hopefully cleanups up a lot of misc issues/improvements that has been staying in the back of my mind. a cleaner setup for a new semester (and a new release before that :rocket:)

webapp/src/components/task_creation_form/TaskCreationForm.vue
```diff
@@ -1,38 +1,831 @@
...
+                  as="input"
+                  type="number"
+                  min="0"
+                  step="0.01"
+                />
+              </FormLabel>
+
+              <FormLabel label="TFJS model loss" type="required">
+                <FormField name="model.loss" placeholder="hinge" as="input" />
+              </FormLabel>
```
JulienVig, line 249: Same here, a dropdown menu would make more sense

webapp/src/components/task_creation_form/TaskCreationForm.vue
```diff
@@ -1,38 +1,831 @@
...
+                  name="model.optimizer.name"
+                  placeholder="sgd"
+                  as="input"
+                />
+              </FormLabel>
+
+              <FormLabel
+                label="TFJS model optimizer learning rate"
+                type="required"
+              >
```
JulienVig, line 210: Can we make this a dropdown menu with only the available optimizers as options? This avoids typos and having to lookup the TFJS documentation.

webapp/src/components/task_creation_form/TaskCreationForm.vue
```diff
@@ -1,38 +1,831 @@
...
+    topology: z.unknown().transform((file, ctx) => {
+      // TODO why is it returning a single file instead of a set?
+      if (file === undefined) {
+        ctx.addIssue({
+          code: z.ZodIssueCode.custom,
+          message: "Missing JSON file",
+        });
+        return z.NEVER;
+      }
+      if (!(file instanceof File))
```
JulienVig: 1. I get an error here when trying to drag and drop titanic's model.json (no error if I click on the button to choose the file). If I print what `file` is when drag and dropping, here is what I get:
```js
Object { size: 1, _map: {…}, __ownerID: undefined }​
__ownerID: undefined​
_map: Object { size: 1, _root: {…}, __altered: false, … }​​
__altered: false​​
__hash: undefined​​
__ownerID: undefined​​
_root: Object { ownerID: {}, entries: (1) […] }​​​
entries: Array [ (2) […] ]​​​​
0: Array [ File, File ]​​​​​
0: File { name: "model.json", lastModified: 1755078810962, size: 45623, … }​​​​​​
lastModified: 1755078810962​​​​​​
name: "model.json"​​​​​​
size: 45623​​​​​​
type: "application/json"​​​​​​
webkitRelativePath: ""​​​​​​
<prototype>: FilePrototype { name: Getter, lastModified: Getter, webkitRelativePath: Getter, … }​​​​​
1: File { name: "model.json", lastModified: 1755078810962, size: 45623, … }​​​​​
length: 2​​​​​
<prototype>: Array []​​​​
length: 1​​​​
<prototype>: Array []​​​
ownerID: Object { }
<prototype>: Object { get: get2(shift, keyHash, key, notSetValue), update: update2(ownerID, shift, keyHash, key, value, didChangeSize, didAlter), iterate: iterate(fn, reverse3), … }
size: 1
<prototype>: Object { constructor: Map2(value), toString: toString5(), get: get11(k, notSetValue), … }
size: 1
<prototype>: Object { constructor: Set2(value), toString: toString5(), has: has5(value), … }
```

2. when I upload the file by clicking on the upload button and choose the file I don't get an error here but one later on when I submit (´Model file isn't valid`), cf comment l.771

JulienVig: Should we also add a toaster error here?

webapp/src/components/task_creation_form/TaskCreationForm.vue
```diff
@@ -1,38 +1,831 @@
...
+    throw new Error("zod validated model info aren't valid");
+  switch (optimizer.name) {
+    case modelOptimizerNames[0]:
+    case modelOptimizerNames[1]:
+    case modelOptimizerNames[2]:
+    case modelOptimizerNames[3]:
+    case modelOptimizerNames[4]:
+    case modelOptimizerNames[5]:
+      // @ts-expect-error whole array is valid
+      modelOptimizerNames[6];
```
JulienVig, line 836: There're only 6 values in this array, why silence the typescript error on a 7th element? I see that the momentum optimizer was commented out, maybe there was a confusion on the total number of elements at some point

webapp/src/components/task_creation_form/TaskCreationForm.vue
```diff
@@ -1,38 +1,831 @@
...
+              size="20"
+            />
+          </FormLabel>
+
+          <FormLabel label="Data type" type="required">
+            <FormField name="dataType" as="select" v-model="dataType">
+              <option value="image">Image</option>
+              <option value="tabular">Tabular</option>
+              <option value="text">Text</option>
+            </FormField>
```
JulienVig: When I select a data type, "Missing dependant fields" immediately shows up underneath. I need to enter all required values after that to make this error disappear.
1. The type of aggregation is not a required field but an error if I don't choose it, I guess we can either make "mean" the default value or make the field required.
2. Since required fields already show an error when missing can we remove the "missing dependant fields" error? (A consequence of having two errors for one missing field is that clicking submit scrolls to the first error which is the data type with "Missing dependant field" rather than the actual missing field and I find this confusing)

webapp/src/components/task_creation_form/TaskCreationForm.vue
```diff
@@ -1,38 +1,831 @@
...
+      throw new Error("zod validated model info aren't valid");
+  }
+
+  // TODO TFJS' browserFiles hangs when parsing invalid JSON tfjs#8517
+  try {
+    JSON.parse(await topology.text());
+  } catch {
+    toaster.error("Model file isn't valid");
+    return;
+  }
```
JulienVig, line 848: When I upload any model.json among titanic, COVID or LLM task, I get the error: `SyntaxError: JSON.parse: unexpected character at line 1 column 1 of the JSON data`. Did you manage to successfully upload a model?

JulienVig (changes requested):
I haven't looked at everything yet but it already looks much better! 
I've encountered a few errors while trying to create tasks so I'm reporting them right away before I start going through the code.

webapp/src/components/dataset_input/FileSelection.vue
```diff
@@ -96,13 +97,18 @@ import CustomButton from "@/components/simple/CustomButton.vue";
 
 const props = withDefaults(
   defineProps<{
-    type: "tabular" | "image" | "text";
+    type: "image" | "json" | "tabular" | "text";
     multiple?: boolean; // accept one or multiple files
+    private?: boolean; // file stays in browser
```
JulienVig: "file stays in browser" makes it sound like setting private=False will upload the data. Can we replace the prop by lockIcon?: boolean for example?

webapp/src/components/dataset_input/FileSelection.vue
```diff
@@ -96,13 +97,18 @@ import CustomButton from "@/components/simple/CustomButton.vue";
...
   }>(),
   {
     multiple: false,
+    private: false,
   },
 );
+const emit = defineEmits<{
+  blur: [];
+}>();
```
JulienVig, line 117: I guess from the commit title that vee-validate expects 'blur' events from custom components, could you add some comments with more info on this event and maybe some pointers to the vee documentation?

webapp/src/components/training/TrainerDashboard.vue
```diff
@@ -5,7 +5,10 @@
       <IconCard title-placement="center">
         <template #title> Control the Training Flow </template>
           <!-- If we are not currently training -->
-          <div v-if="!isTraining" class="flex flex-col gap-y-4">
+          <div
+            v-if="!isTraining"
+            class="flex flex-col gap-y-4 min-w-full lg:min-w-[42rem]"
```
JulienVig: I think this css change de-centered the toggle button
Screenshot 2025-08-13 at 17 14 04

webapp/src/components/task_creation_form/TaskCreationForm.vue
tharvik: good idea! I got used to it but that's clearly more user friendly

webapp/src/components/task_creation_form/TaskCreationForm.vue
tharvik: no, the user can't input smth else as this value is coming from our component so nothing they can do about it to fix it

webapp/src/components/task_creation_form/TaskCreationForm.vue
tharvik: yeah, that's a typescript trick I'm using to ensure that the switch is correctly checking for all the values inside the array. if a new value is added to the array, it will fail to compile until this switch is changed. it's setting a consistency trap for later if you will

webapp/src/components/task_creation_form/TaskCreationForm.vue
tharvik: arf, yeah, I'm unable to find the actual source of the bug, probably some weird behavior that vee-validate is doing. working around that

webapp/src/components/task_creation_form/TaskCreationForm.vue
tharvik: I think you are uploading the `model.json` we store on the server, no? its format is actually our own serialized version and not a TFJS model encoding. yeah, that's confusing..
you can get a "classic" model.json there https://storage.googleapis.com/tfjs-models/tfjs/iris_v1/model.json

webapp/src/components/task_creation_form/TaskCreationForm.vue
tharvik: > When I select a data type, "Missing dependant fields" immediately shows up underneath. I need to enter all required values after that to make this error disappear.
> Since required fields already show an error when missing can we remove the "missing dependant fields" error?

yeah, that's indeed bothering; moving to supress this error (and it's indeed nicer!)

> 1. The type of aggregation is not a required field but an error if I don't choose it, I guess we can either make "mean" the default value or make the field required.

making it both required and starting with "mean"

webapp/src/components/task_creation_form/TaskCreationForm.vue
```diff
@@ -180,12 +180,23 @@
...
               </FormLabel>
 
+              <FormLabel label="TFJS model optimizer learning rate">
+                <FormField
+                  name="model.optimizer.learningRate"
+                  placeholder="0.01"
+                  as="input"
+                  type="number"
+                  min="0"
+                  step="0.01"
```
JulienVig: Can we remove the step parameter? I don't think we need to constrain user input here. I sometimes use learn rates down to 1e-7.
We should exclude 0 too, so somthing like gt=0

webapp/src/components/task_creation_form/TaskCreationForm.vue
```diff
@@ -701,8 +701,10 @@ async function onSubmit(form: unknown): Promise<void> {
   try {
     await pushTask(CONFIG.serverUrl, task, model);
   } catch (e) {
-    toaster.error("An error occured server-side");
     debug("while pushing task to server: %o", e);
+    if (e instanceof Error && e.message.endsWith("HTTP status 409"))
+      toaster.error("This identifier is already taken");
+    else toaster.error("An error occured server-side");
```
JulienVig, line 879: One day we'll have a better framework for error handling #677 ...

discojs/src/task/task.ts
```diff
@@ -1,49 +1,70 @@
...
 import { TrainingInformation } from "./training_information.js";
 
 export namespace Task {
-  export type ID = string;
+	export type ID = string;
 }
 
-const baseSchema = z
-	.object({
+export namespace Task {
```
JulienVig: Maybe directly merge the two `Task` namespaces

webapp/src/components/task_creation_form/TaskCreationForm.vue
```diff
@@ -1,38 +1,886 @@
...
+                  name="trainingInformation.maxShareValue"
+                  placeholder="100"
+                  as="input"
+                  type="number"
+                />
+              </FormLabel>
+            </FormLabel>
+
+            <FormLabel
+              label="Number of epochs before exchanging with the network"
```
JulienVig:
```suggestion
              label="Number of epochs before aggregating weights"
```
Since we're saying "Type of aggregation" in the previous field we might as well re-use the same language I think?

webapp/src/components/task_creation_form/TaskCreationForm.vue
```diff
@@ -1,38 +1,886 @@
...
+                v-model="scheme"
+                supress-error
+              >
+                <option value="federated">Federated</option>
+                <option value="decentralized">Decentralized</option>
+                <option value="local">Local</option>
+              </FormField>
+            </FormLabel>
+
+            <FormLabel label="Type of aggregation to perform" type="required">
```
JulienVig:
```suggestion
            <FormLabel label="Type of aggregation" type="required">
```

webapp/src/components/task_creation_form/TaskCreationForm.vue
```diff
@@ -1,38 +1,886 @@
...
+                      placeholder="label"
+                      as="input"
+                    />
+
+                    <CustomButton @click="remove(i)">
+                      <i class="fa-solid fa-xmark"></i>
+                    </CustomButton>
+                  </div>
+
+                  <CustomButton @click="push('')"> add label </CustomButton>
```
JulienVig, line 93: The Add Label button's actual size doesn't match its visual

https://github.com/user-attachments/assets/bea1665f-36f7-4ec0-bfa7-42b10e171fbc

webapp/src/components/task_creation_form/TaskCreationForm.vue
```diff
@@ -1,38 +1,886 @@
...
+                      placeholder="field_name"
+                      as="input"
+                    />
+
+                    <CustomButton @click="remove(i)">
+                      <i class="fa-solid fa-xmark"></i>
+                    </CustomButton>
+                  </div>
+
+                  <CustomButton @click="push('')"> add column </CustomButton>
```
JulienVig, line 146: Same here, the actual clickable zone is bigger than the button's visual

webapp/src/components/task_creation_form/TaskCreationForm.vue
```diff
@@ -1,38 +1,886 @@
...
+              <FormField
+                name="trainingInformation.aggregationStrategy"
+                as="select"
+                v-model="aggregationStrategy"
+              >
+                <option value="mean">Mean</option>
+                <option v-if="scheme === 'decentralized'" value="secure">
+                  Secure
+                </option>
+              </FormField>
```
JulienVig: If it's easy it would make sense I think to omit this field when there's only a single option (federated only allows mean aggregation). I think having a field with a single option is a bit weird/confusing but it's not a big deal if that's time-consuming to implement

JulienVig (approved):
The form is working great!! I left some minor comments but nothing blocking.

1. If it's easy it would be great if successfully submitting a task takes you to the task list (and resets the creation form), right now it just shows a toaster without anything else which made me wonder if it actually worked
2. Once the PR is merged, created tasks will get deleted at every server reboot right? Did you have something in mind about this?
3. If we implement something to keep tasks permanently, we'll probably have to also have a way to remove some of them.

webapp/src/components/training/TrainerDashboard.vue
tharvik: haa yes, well spotted!

webapp/src/components/task_creation_form/TaskCreationForm.vue
tharvik: yeah, that's a bit weird indeed. I looked into it a bit and I found the rational behind it: as I'm using `<label>` to wrap every field, and when interacting with it it behave as if you interacted with the field directly. the rational from the W3C is for accessibility https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/label#usage_notes

now, to get the best of both world, we can start with a empty label already created, and so having two buttons, delete and add, so it won't move to one on hover (and is also more explicit that it needs at least one label)

webapp/src/components/task_creation_form/TaskCreationForm.vue
tharvik: removing the field proved more complicated then expected. disabling the field instead.

tharvik:
> 1. If it's easy it would be great if successfully submitting a task takes you to the task list (and resets the creation form), right now it just shows a toaster without anything else which made me wonder if it actually worked

great idea! added

> 2. Once the PR is merged, created tasks will get deleted at every server reboot right? Did you have something in mind about this?

right, for now, I'm telling myself that it's okay as it's only a demo (and it's already the case). persistence can be added and would prove useful for sure!

> 3. If we implement something to keep tasks permanently, we'll probably have to also have a way to remove some of them.

yeah, that'd make sense, but allowing anyone to remove a task (as we don't have an user identification system) seems wrong. I'm maybe over thinking this and we might just allow that, and regularly restore it to a know good state (i.e. our default task list)

anyway, thanks for the multiple reviews, that helped to have another look at it! I'm currently rebasing everything and checking that each commit works but soon, very soon, we'll have it merged :tada: finally :mirror_ball:

PR #875, opened 2025-02-17
*: bump deps

tharvik:
node-datachannel strangly fails in recent builds. ~hopefully bumping deps will fix that.~ it did :rocket: the central issue is that `prebuild-install@7.1.2` (used by node-datachannel) had [a broken NAPI detection](https://github.com/prebuild/prebuild-install/issues/203)

while at it: bump major immutable, msgpack (not stable because [still broken in browser](https://github.com/msgpack/msgpack-javascript/issues/260)) and a bunch of misc pkgs


cli/src/benchmark_gpt.ts
```diff
@@ -119,7 +119,8 @@ async function main(args: Required<CLIArguments>): Promise<void> {
     for (let i = 0; i < iterations; i++) {
       const timeStart = performance.now()
       for (let n = 0; n < maxNewTokens; n++) {
-        const next: number = (await model.predict(List.of(tokens))).first();
+				const next = (await model.predict(List.of(tokens))).first();
+				if (next === undefined) throw new Error("empty prediction");
```
JulienVig, line 123: Is there a lot of whitespace here or is it github freaking out?

cli/src/train_gpt.ts
```diff
@@ -36,9 +36,8 @@ async function main(): Promise<void> {
 
   const maxNewTokens = 14
   for (let n = 0; n < maxNewTokens; n++) {
-    const next: number = (await model.predict(
-      List.of(tokens), { seed })
-    ).first();
+		const next = (await model.predict(List.of(tokens), { seed })).first();
+		if (next === undefined) throw new Error("empty prediction");
```
JulienVig, line 40: Same here regatding whitespace

JulienVig (approved):
All good thanks for the quick fix and the updates!

I see a lot of unnecessary indents could you check if you're adding whitespace?

cli/src/benchmark_gpt.ts
tharvik: ha yes, I'm using tabs for indentation now via [biome](https://biomejs.dev/) which is shown as two spaces in my editor (but not github it seems)

shall we [commit to a formatter](https://github.com/epfml/disco/issues/631) one of theses days?

JulienVig: Yes it might be time

PR #876, opened 2025-02-17
docker: fix missing dep in image

tharvik:
after #875, look like it [failed to deploy the server](https://github.com/epfml/disco/actions/runs/13374731961/job/37351408626), saying that it misses a dep. fixing it by removing code :1st_place_medal:
also adding a CI step to try to avoid that.


JulienVig (approved):
Looks good to me! Why move with_server's content into lint-test-build?

tharvik:
> Why move with_server's content into lint-test-build?

it has been unused for a while and the new test in the CI is pretty much doing the same thing

PR #890, opened 2025-03-07
2 more tokenization tests

christinakopi:


discojs/src/processing/decode.spec.ts
```diff
@@ -0,0 +1,23 @@
...
+    // Load the GPT2 tokenizer
+    const tokenizer = await AutoTokenizer.from_pretrained("Xenova/gpt2");
+    const originalText = "Hello, world! This is a test for encode-decode tokenization.";
+    
+    // Perform round-trip tokenization
+    const decodedText = encodeDecode(tokenizer, originalText);
+    
+    // Normalize both strings to account for minor tokenization differences
+    const normalizedOriginal = originalText.trim().toLowerCase();
+    const normalizedDecoded = decodedText.trim().toLowerCase();
```
JulienVig: Ideally we want the tokenization to be exact, were there inconsistencies?

discojs/src/processing/decode.ts
```diff
@@ -0,0 +1,21 @@
...
+ * @param tokenizer - An instance of a PreTrainedTokenizer
+ * @param text - The text to process
+ * @returns The decoded text obtained after encoding and then decoding
+ */
+export function encodeDecode(tokenizer: PreTrainedTokenizer, text: string): string {
+  // Encode the text using the tokenizer.
+  const encoding = tokenizer(text, { return_tensor: false }) as TokenizerOutput;
+  // Decode the token IDs back into text while skipping special tokens.
+  return tokenizer.decode(encoding.input_ids, { skip_special_tokens: true });
+}
```
JulienVig: `encodeDecode` is only used for testing purposes. Rather than creating a new file and exporting the function (as we do for helper function), you can just create a private function in the spec file.

discojs/src/processing/tokenize.ts
```diff
@@ -0,0 +1,31 @@
...
+  config: Record<string, unknown> = {}
+): List<number> {
+  const tokenizerResult = tokenizer(text, { ...config, return_tensor: false }) as TokenizerOutput;
+
+  if (!tokenizerResult || !Array.isArray(tokenizerResult.input_ids)) {
+    throw new Error("Tokenizer returned unexpected output");
+  }
+  
+  return List(tokenizerResult.input_ids);
+}
```
JulienVig: This function has already been implemented in `processing/text.ts` :)

discojs/src/processing/tokenize.spec.ts
```diff
@@ -0,0 +1,32 @@
+import { expect } from "chai";
+import { AutoTokenizer } from "@xenova/transformers";
+import { tokenize } from "./tokenize.js";
+
+describe("Multi-Tokenizer Tests", function () {
+  this.timeout(200000);
```
JulienVig: 200s is really long, can the timeout be tighter?

discojs/src/processing/tokenize.spec.ts
JulienVig, line 1: The test is good! Though rather than creating a new tokenize.spec.ts file you can put this test in text.spec.ts instead.

discojs/src/processing/decode.ts
```diff
@@ -0,0 +1,21 @@
+import { PreTrainedTokenizer } from "@xenova/transformers";
+
+
+interface TokenizerOutput {
+  input_ids: number[];
+}
```
JulienVig: If you need to define the same interface twice, then you should define it once and import it in the other file!

JulienVig (changes requested):
Very clean work! The code is neat and well documented. I just have some file organization comments:
* `tokenize.ts` is duplicated as already defined in `text.ts`
* move all the tests in `text.spec.ts` rather than create two new spec files
* `decode.ts` shouldn't be a new file, just a private function in `text.spec.ts`

You can request a review again once you resolved the comments!

JulienVig (approved):

PR #895, opened 2025-03-13
Fix number of participant inconsistency

### LINKED ISSUE #894: Number of participant is inconsistent during training
As identified by @christinakopi, during collaborative training the number of participant displayed is inconsistent.
Upon start, the training dashboard displays 1 participant, which is expected as model sharing hasn't been performed yet.
However, the number of participants remains at 1 after the first weight sharing round (which happens only when more than 1 participant is training).
The correct nb of participant is displayed after the 2nd weight sharing round.
Tested with GPT2, training with two browsers, or MNIST with 3 participants.

JulienVig:
Fixes #894 identified by @christinakopi.

We used to update the number of participants only during weight sharing. I made a few changes to update the participants more frequently (doesn't cover every case yet):
* `Client` emits a `participants` event (which is forwarded by `Disco`) that the webapp can watch
* The server now sends the number of participants at connection, in `WaitForMoreParticipants` messages and in `EnoughParticipants` messages
* With tests©
* The number of participants is **not** dynamically updated when clients are waiting for more participants, it stays fixed until enough participants joined and the training can start (we can implement a new message type to update clients when a new client joins).


discojs/src/client/client.ts
```diff
@@ -40,7 +43,10 @@ export abstract class Client extends EventEmitter<{'status': RoundStatus}>{
    * we were doing before waiting (training locally or updating our model).
    * We use this attribute to store the status to rollback to when we stop waiting
    */
-  private previousStatus: RoundStatus | undefined;
+  #previousStatus: RoundStatus | undefined;
+
+  // Current number of participants in the training session
```
tharvik: I'm never sure if that includes the current client or not, consider adding a small comment to help my brain :)

server/tests/e2e/utils.ts
```diff
@@ -29,3 +37,30 @@ export class Queue<T> {
     }
   }
 }
+
+export async function setupLusCOVID(scheme: "federated" | "decentralized") {
```
tharvik: I guess you can also use it tests/e2e/federated.lusCovidUser and validator

server/tests/e2e/decentralized.spec.ts
```diff
@@ -262,40 +251,65 @@ describe('end-to-end decentralized', function () {
...
 
+    /* USER 3 JOINS */
+
     // Create User 3
     const discoUser3 = new Disco(lusCovidTask, url, { preprocessOnce: true });
     const statusUser3 = new Queue<RoundStatus>();
+    const nbParticipantsUser3 = new Queue<number>();
     discoUser3.on("status", status => { statusUser3.put(status) })
+    discoUser3.on("participants", (participants) => {
+      console.log("user 3: ", participants)
```
tharvik: debug
```suggestion
```

tharvik (approved):
haa, works flawlessly, niceux!

PR #899, opened 2025-03-18
test GPT layers

christinakopi:


discojs/src/models/gpt/layers.spec.ts
```diff
@@ -0,0 +1,308 @@
...
+
+    afterEach(() => {
+      // Dispose of variables to avoid name collisions in subsequent tests.
+      tf.disposeVariables();
+    });
+
+    it('should compute GELU activation correctly for known inputs', async function () {
+      const geluLayer = new GELU();
+
+      const input: tf.Tensor1D = tf.tensor1d([0, 1, -1, 2, -2]);
```
tharvik: `tf.tensor1d` already returns a `tf.Tensor1D`

```suggestion
      const input = tf.tensor1d([0, 1, -1, 2, -2]);
```

discojs/src/models/gpt/layers.spec.ts
```diff
@@ -0,0 +1,308 @@
...
+
+      const output = geluLayer.apply(input) as tf.Tensor;
+      const outputData: Float32Array = await output.data() as Float32Array;
```
tharvik: no need to cast as only used as an array of number

```suggestion
      const outputData = await output.data();
```

discojs/src/models/gpt/layers.spec.ts
```diff
@@ -0,0 +1,308 @@
...
+
+    it('should return token embeddings with shape [batch_size, sequence_length, nEmbd] for 2D input', function () {
+      const vocabSize = 100;
+      const nEmbd = 16;
+      const seed = 42;
+
+      const lmEmbedding = new LMEmbedding(vocabSize, nEmbd, seed);
+      
+      // dummy 2D input representing token indices: shape [batch_size, sequence_length]
+      const tokenIndices = tf.tensor2d([[1, 2, 3, 4, 5], [6, 7, 8, 9, 10]], [2, 5], 'int32');
```
tharvik: it's more readable to use `tf.randomUniformInt` as you do below to clearly state that it is a random tensor

```suggestion
      const tokenIndices = tf.randomUniformInt([2, 5], 0, 1);
```

discojs/src/models/gpt/layers.spec.ts
```diff
@@ -0,0 +1,308 @@
...
+      const lmEmbedding = new LMEmbedding(vocabSize, nEmbd, seed);
+
+      // dummy 3D input representing a batch of embeddings: shape [batch_size, sequence_length, nEmbd]
+      const embeddingsInput = tf.randomUniform([2, 5, nEmbd]);
+
+      const output = lmEmbedding.apply(embeddingsInput) as tf.Tensor;
+
+      // expected output shape for 3D input: [2, 5, vocabSize]
+      expect(output.shape).to.deep.equal([2, 5, vocabSize]);
+    });
```
tharvik: hmm, I would merge the two tests (w/ "should return token embeddings [...]") as their content are quite similar.

it's always hard to gauge when to have a new testcase rather than adding to one. in here, it is also that the setup is very similar (I even thought that it was duplicated), so if you do not want to merge theses, consider extracting the setup in a dedicated function to let the reader focus on what is different btwn the two

smth like
```ts
it("should work for 2D & 3D inputs", () => {
	const vocabSize = 100;
	const nEmbd = 16;
	const seed = 42;

	const lmEmbedding = new LMEmbedding(vocabSize, nEmbd, seed);

	const tokenIndices = tf.randomUniformInt([2, 5], 0, 1);
	const embeddingsInput = tf.randomUniform([2, 5, nEmbd]);

	const outputForToken = lmEmbedding.apply(tokenIndices) as tf.Tensor;
	const outputForEmbedding = lmEmbedding.apply(embeddingsInput) as tf.Tensor;

	expect(outputForToken.shape).to.deep.equal([2, 5, nEmbd]);
	expect(outputForEmbedding.shape).to.deep.equal([2, 5, vocabSize]);
});
```

discojs/src/models/gpt/layers.spec.ts
```diff
@@ -0,0 +1,308 @@
...
+
+    it('should throw an error if input is an array with more than one tensor', function () {
+      const vocabSize = 100;
+      const nEmbd = 16;
+      const seed = 42;
+      const lmEmbedding = new LMEmbedding(vocabSize, nEmbd, seed);
+      const input1 = tf.tensor2d([[1, 2, 3]], [1, 3], 'int32');
+      const input2 = tf.tensor2d([[4, 5, 6]], [1, 3], 'int32');
+      expect(() => lmEmbedding.apply([input1, input2])).to.throw('expected exactly one tensor');
+    });
```
tharvik: maybe also merge theses two. their purpose are to check that bad values get rejected which is quite close also.

btw, testing for bad values is more cumbersome to write than really useful. unit tests are mostly to find out if a piece of code is doing the right thing, not that it does in every case (which might be computionally prohibitive in turing complete languages). such checks also rarely break.

discojs/src/models/gpt/layers.ts
```diff
@@ -125,110 +125,289 @@ class CausalSelfAttention extends tf.layers.Layer {
...
+//       let y = tf.matMul(att, v) // (B, nHead, T, T) x (B, nHead, T, hs) -> (B, nHead, T, hs)
+//       y = tf.transpose(y, [0, 2, 1, 3]) // (B, T, nHead, hs)
+//       y = tf.reshape(y, [B, T, C]) // (B, T, C = nHead * hs)
+//       y = dense(y, this.cProjKernel, this.cProjBias) // output projection (B, T, C)
+//       y = kwargs.training === true ? tf.dropout(y, this.dropout, undefined, this.seed) : y
+//       return y
+//     })
+//   }
+// }
+// tf.serialization.registerClass(CausalSelfAttention)
```
tharvik: do not keep commented code around, either use it or get rid of it

discojs/src/models/gpt/layers.spec.ts
```diff
@@ -0,0 +1,308 @@
...
+    //     ) as tf.layers.LayerVariable;
+  
+    //     const output = csa._projectOutput(x, projKernel, projBias);
+    //     const data = await output.data();
+    //     // Calculation:
+    //     // [1,2,3] dot kernel = [1*1+2*0+3*1, 1*0+2*1+3*(-1)] = [4, -1]
+    //     // Then add bias [0.5, 0.5] = [4.5, -0.5]
+    //     expect(Array.from(data)).to.deep.equal([4.5, -0.5]);
+    //   });
+    // });
```
tharvik: rm commented code

discojs/src/models/gpt/layers.spec.ts
```diff
@@ -0,0 +1,308 @@
...
+    //       tf.initializers.constant({ value: [0.5, -0.5] })
+    //     ) as tf.layers.LayerVariable;
+  
+    //     const output = csa._dense(x, kernel, bias);
+    //     const outData = await output.data();
+    //     // Expected calculation:
+    //     // [1,2] dot [[1,0],[0,1]] = [1,2] and then add bias [0.5, -0.5] gives [1.5, 1.5]
+    //     expect(Array.from(outData)).to.deep.equal([1.5, 1.5]);
+    //   });
+    // });
```
tharvik: rm or use

discojs/src/processing/text.spec.ts
```diff
@@ -1,84 +1,153 @@
...
+
+  it("should return text close to the original after encode-decode tokenization using GPT2 tokenizer", async function () {
+    // Load the GPT2 tokenizer
+    const tokenizer = await AutoTokenizer.from_pretrained("Xenova/gpt2");
+    const originalText = "Hello, world! This is a test for encode-decode tokenization.";
+    
+    // Perform round-trip tokenization
+    const decodedText = encodeDecode(tokenizer, originalText);
+    
+    // Check that the decoded text is almost equal to the original text
```
tharvik: "almost equal" (and "close" in the testcase description)? that's not what the check below does

discojs/src/processing/text.spec.ts
```diff
@@ -1,84 +1,153 @@
...
+
+/**
+ * Encodes the text into token IDs and then decodes them back to text
+ * Special tokens are skipped during decoding
+ *
+ * @param tokenizer - An instance of a PreTrainedTokenizer
+ * @param text - The text to process
+ * @returns The decoded text obtained after encoding and then decoding
+ */
+export function encodeDecode(tokenizer: PreTrainedTokenizer, text: string): string {
```
tharvik: no need to export it as it is only used here. also, you can inline it as it used only once.

discojs/src/models/gpt/layers.spec.ts
```diff
@@ -0,0 +1,308 @@
...
+      const output = rangeLayer.apply(dummyInput) as tf.Tensor;
+  
+      // We expect the output to have shape [1, T] i.e. [1, 10]
+      expect(output.shape).to.deep.equal([1, 10]);
+  
+      // verify the content: the layer should output a range [0, 1, ..., T-1]
+      const outputData = await output.data();
+      for (let i = 0; i < 10; i++) {
+        expect(outputData[i]).to.equal(i);
+      }
```
tharvik: for small arrays, I find that checking directly is more readable

```suggestion
      expect(await output.data()).to.deep.equal(
        Int32Array.of(0, 1, 2, 3, 4, 5, 6, 7, 8, 9),
      );
```

discojs/src/models/gpt/layers.spec.ts
```diff
@@ -0,0 +1,308 @@
...
+      const expected: number[] = [0, 0.8412, -0.1588, 1.955, -0.045];
+
+      for (let i = 0; i < expected.length; i++) {
+        expect(outputData[i]).to.be.closeTo(expected[i], 0.05);
+      }
+    });
+  });
+
+  // LMEmbedding Layer tests
+  describe('LMEmbedding Layer', function () {
```
tharvik, line 33: you only check for shapes of the tensors for this layer, is there a way to also check for the outputted data?

discojs/src/models/gpt/layers.spec.ts
```diff
@@ -0,0 +1,308 @@
...
+  
+      // get predictions from both models
+      const output1 = model1.predict(input) as tf.Tensor;
+      const output2 = model2.predict(input) as tf.Tensor;
+  
+      const arr1 = await output1.data();
+      const arr2 = await output2.data();
+  
+      // check lengths are equal
+      expect(arr1.length).to.equal(arr2.length);
```
tharvik: the "deep" check below already does the job
```suggestion
```

discojs/src/models/gpt/layers.spec.ts
```diff
@@ -0,0 +1,308 @@
...
+  
+      // check that the models produce the same output
+      expect(arr1).to.deep.equal(arr2);
+
+      // Check that there are no NaN values in the outputs.
+      for (let i = 0; i < arr1.length; i++) {
+        expect(isNaN(arr1[i])).to.be.false;
```
tharvik: #nitpick (also a bit safer as [`isNan` is doing type cohersion](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isNaN#description))

```suggestion
        expect(arr1[i]).to.not.be.NaN;
```

discojs/src/models/gpt/layers.spec.ts
```diff
@@ -0,0 +1,308 @@
...
+      const arr2 = await output2.data();
+  
+      // check lengths are equal
+      expect(arr1.length).to.equal(arr2.length);
+  
+      // check that the models produce the same output
+      expect(arr1).to.deep.equal(arr2);
+
+      // Check that there are no NaN values in the outputs.
+      for (let i = 0; i < arr1.length; i++) {
```
tharvik: better to use `for … of` to loop over an array
```suggestion
      for (const v of arr1) {
```

discojs/src/models/gpt/layers.spec.ts
```diff
@@ -0,0 +1,308 @@
...
+        expect(isNaN(arr1[i])).to.be.false;
+      }
+      for (let i = 0; i < arr2.length; i++) {
+        expect(isNaN(arr2[i])).to.be.false;
+      }
```
tharvik: same `for … of` & and `expect(…).to.not.be.NaN`

discojs/src/models/gpt/layers.spec.ts
```diff
@@ -0,0 +1,308 @@
...
+    //   });
+    // });
+  
+    describe('splitHeads', function () {
+      it('should reshape and transpose the input correctly', function () {
+        const B = 2;
+        const T = 6;
+        const totalChannels = config.nEmbd; // 8 channels
+        // input tensor with shape [B, T, totalChannels]
+        const input = tf.tensor3d(new Array(B * T * totalChannels).fill(1), [B, T, totalChannels]);
```
tharvik: [`tf.ones` creates a tensor filled with ones](https://js.tensorflow.org/api/4.22.0/#ones)

```suggestion
        const input = tf.ones([B, T, totalChannels]);
```

discojs/src/models/gpt/layers.spec.ts
```diff
@@ -0,0 +1,308 @@
...
+        const att = tf.zeros([1, 1, T, T], 'float32');
+        const masked = csa.applyCausalMask(att, T);
+        const data = await masked.data();
+        // for each position (i,j): if j > i expect -1e9 else 0
+        const expected: number[] = [];
+        for (let i = 0; i < T; i++) {
+          for (let j = 0; j < T; j++) {
+            expected.push(j > i ? -1e9 : 0);
+          }
+        }
```
tharvik: yeah, tests should be very readable as to spot errors more rapidly. what do you think of have the excepted array directly in there?

```suggestion
        const expected = [
          [0, 1, 1, 1, 1],
          [0, 0, 1, 1, 1],
          [0, 0, 0, 1, 1],
          [0, 0, 0, 0, 1],
          [0, 0, 0, 0, 0],
        ]
          .flat()
          .map((v) => (v === 0 ? 0 : -1e9));
```

discojs/src/models/gpt/layers.spec.ts
```diff
@@ -0,0 +1,308 @@
...
+          for (let h = 0; h < nHead; h++) {
+            for (let i = 0; i < T; i++) {
+              // calculate the starting index for the i-th row in the flattened tensor
+              const rowStart = b * nHead * T * T + h * T * T + i * T;
+              const row = attArray.slice(rowStart, rowStart + T);
+              const rowSum = row.reduce((sum, val) => sum + val, 0);
+              expect(rowSum).to.be.closeTo(1, 1e-3);
+            }
+          }
+        }
```
tharvik: you can [sum over the last dimension](https://js.tensorflow.org/api/4.22.0/#sum) ("-1" mean the last one if tfjs)

```suggestion
        for (const rowSum of await att.sum(-1).data())
          expect(rowSum).to.be.closeTo(1, 1e-3);
```

discojs/src/models/gpt/layers.ts
```diff
@@ -125,110 +125,289 @@ class CausalSelfAttention extends tf.layers.Layer {
...
-      y = dense(y, this.cProjKernel, this.cProjBias) // output projection (B, T, C)
+      y = this.dense(y, this.cProjKernel, this.cProjBias) // output projection (B, T, C)
       y = kwargs.training === true ? tf.dropout(y, this.dropout, undefined, this.seed) : y
-      return y
-    })
+      return y;
+    });
+  }
+
+  // --- Helper Methods ---
```
tharvik, line 195: no need to have the helper methods as `public`: it is not used elsewhere

also, consider moving these outside of the class (into flat function in the same file maybe), as from a quick look, these do not need access to `this`.

tharvik (approved):
haa thanks for shedding some light on the dark corners of GPT! some style comments but nothing blocking.
note that I wasn't able to check for the layers implementations as it is still magical to me but as long as the "gpt-tfjs can overfit one sentence" test is green, sounds good!

also, I rebased your branch to update it with the latest `develop` changes (which also fixes the CI error)

discojs/src/models/gpt/layers.ts
christinakopi: Should I export them in order to use them in the test file instead of defining them as public?

tharvik: ho oupsi, I overlooked that I was used in the tests! nvm, keep theses `public` then

PR #903, opened 2025-03-31
webapp: bump cypress

tharvik:
[cypress GitHub action broke with newer node](https://github.com/cypress-io/github-action/issues/1408).
waited for upstream to solve it and bumped it.


JulienVig (approved):
Yay 🎉

PR #906, opened 2025-04-02
Convert homepage webm animations to gif

JulienVig:
Converting videos to GIFs for performance reasons


tharvik (approved):
aren't the animation the same in both themes? if so, consider adding an alpha channel to use the same image in both cases

JulienVig:
The software @MalloryHenriet is using doesn't support transparent backgrounds for GIFs :(

PR #913, opened 2025-04-21
Byzantine robust aggregator - initial implementation

mina5rovic:
Initial implementation of Byzantine robust aggregator with momentums and Center-Clipping aggregation.


discojs/src/aggregator/byzantine.ts
```diff
@@ -0,0 +1,185 @@
...
+
+const debug = createDebug("discojs:aggregator:mean");
+
+type ThresholdType = 'relative' | 'absolute'
+
+/** 
+ * Mean aggregator whose aggregation step consists in computing the mean of the received weights. 
+ * 
+ */
+export class ByzantineRobustAggregator extends Aggregator {
```
tharvik: there are some duplicated logic from `MeanAggregator`, consider extending from a common class (smth like `MultiRoundAggregator`?) to simplify

discojs/src/aggregator/byzantine.ts
```diff
@@ -0,0 +1,185 @@
...
+     * set `threshold = 1` and `thresholdType = 'absolute'`
+     * @param thresholdType 'relative' or 'absolute', defaults to 'relative'. Is only used to clarify the case when threshold = 1, 
+     * If `threshold != 1` then the specified thresholdType is ignored and overwritten
+     * If `thresholdType = 'absolute'` then `threshold = 1` means waiting for 1 contribution
+     * if `thresholdType = 'relative'` then `threshold = 1`` means 100% of this.nodes' contributions, 
+     * @param roundCutoff - from how many past rounds do we still accept contributions. 
+     * If 0 then only accept contributions from the current round, 
+     * if 1 then the current round and the previous one, etc.
+     */
+    constructor(roundCutoff = 0, threshold = 1, thresholdType?: ThresholdType, clippingRadius: number = 1.0, maxIterations: number = 10) {
```
tharvik: type trivially infered

```suggestion
    constructor(roundCutoff = 0, threshold = 1, thresholdType?: ThresholdType, clippingRadius = 1.0, maxIterations = 10) {
```

discojs/src/aggregator/byzantine.ts
```diff
@@ -0,0 +1,185 @@
...
+        this.log(
+            this.contributions.hasIn([0, nodeId]) ? AggregationStep.UPDATE : AggregationStep.ADD,
+            nodeId,
+        );
+
+        const beta: number = 0.9; // Momentum parameter
+
+        // Apply worker-side momentum
+
+        const prevMomentum: WeightsContainer = this.contributions.getIn([0, nodeId]) as WeightsContainer;
```
tharvik: if it isn't set (such when a node fails to reply in time), it will actually returned `undefined`, it's safer to show it in type

```suggestion
        const prevMomentum = this.contributions.getIn([0, nodeId]) as WeightsContainer | undefined;
```

discojs/src/aggregator/byzantine.ts
```diff
@@ -0,0 +1,185 @@
...
+    }
+
+    override _add(nodeId: client.NodeID, contribution: WeightsContainer): void {
+
+        this.log(
+            this.contributions.hasIn([0, nodeId]) ? AggregationStep.UPDATE : AggregationStep.ADD,
+            nodeId,
+        );
+
+        const beta: number = 0.9; // Momentum parameter
```
tharvik: is it model/task dependent or is it in most cases the "right" value? add a comment to explain a bit more the tradeoff about it

discojs/src/aggregator/byzantine.ts
```diff
@@ -0,0 +1,185 @@
...
+        // Apply worker-side momentum
+
+        const prevMomentum: WeightsContainer = this.contributions.getIn([0, nodeId]) as WeightsContainer;
+        let momentum: WeightsContainer;
+
+        if (prevMomentum) {
+            // m_t = (1 - beta) * grad + beta * m_{t-1}
+            momentum = contribution.map((g, i) =>
+                g.mul(1 - beta).add(prevMomentum.weights[i].mul(beta))
+            );
```
tharvik: you can use `mapWith` to map over the two `WeightsContainer` element-wise (zip and map in other languages)

```suggestion
            momentum = contribution.mapWith(prevMomentum, (g, prev) =>
                g.mul(1 - beta).add(prev.mul(beta))
            );
```

discojs/src/aggregator/byzantine.ts
```diff
@@ -0,0 +1,185 @@
...
+        } else {
+            momentum = contribution.map(g => g.mul(1 - beta));
+        }
```
tharvik: correct me if I'm wrong but isn't that the same as the `if` branch if `prevMomentum` is a null vector of the same shape? if yes, consider merging the two and upon `contributions.getIn`, use a default value instead.

discojs/src/aggregator/byzantine.ts
tharvik: can you add some comment on what theses new parameters do?

discojs/src/aggregator/byzantine.ts
tharvik: can you add some information on what it is based? (paper URL, assumption required to use it safely, tradeoffs, …)

discojs/src/aggregator/byzantine.ts
```diff
@@ -0,0 +1,185 @@
...
+    constructor(roundCutoff = 0, threshold = 1, thresholdType?: ThresholdType, clippingRadius: number = 1.0, maxIterations: number = 10) {
+
+        if (threshold <= 0) throw new Error("threshold must be strictly positive");
+        if (threshold > 1 && (!Number.isInteger(threshold)))
+            throw new Error("absolute thresholds must be integral");
+
+        super(roundCutoff, 1);
+        this.#threshold = threshold;
+        this.clippingRadius = clippingRadius
+        this.maxIterations = maxIterations
```
tharvik: add some values check to ensure that both are positive, that `maxIterations` is an integer, and so on

discojs/src/aggregator/byzantine.ts
```diff
@@ -0,0 +1,185 @@
...
+                return diff.mul(scale).add(v)
+            })
+
+            v = aggregation.avg(updated.values());
+        }
+
+        return v
+    }
+
+    private euclideanNorm(w: WeightsContainer): tf.Scalar {
```
tharvik: doesn't need access to `this`, better to extract it out of the class (in the same file)

tharvik (changes requested):
thanks for the great work! :tada:

a few things to move around/improve and we'll be ready to go!
can you also add a testcase to ensure that it behave well?

martinjaggi:
maybe later once your code is tested we could also do a small comparison to the older implementation here https://github.com/epfml/disco/commit/e53676c2f4bc9c00a34f61df925437d06326d24c  (that's a very old completely differently structured version of disco, but potentially achieving similar ML accuracy target under byzantine, which might be interesting to check if time remains)

discojs/src/aggregator/byzantine.ts
```diff
@@ -0,0 +1,125 @@
...
+ *     The update typically looks like: `v_t = beta * v_{t-1} + (1 - beta) * g_t`, where `g_t` is the current clipped average.
+ *   - A higher beta gives more weight to past rounds (more smoothing), while a lower beta makes the aggregator more responsive to new updates.
+ */
+
+
+  constructor(roundCutoff = 0, threshold = 1, thresholdType?: ThresholdType, clippingRadius = 1.0, maxIterations = 1, beta = 0.9) {
+    super(roundCutoff, threshold, thresholdType);
+    if (clippingRadius <= 0) throw new Error("Clipping radius needs to be positive number > 0.");
+    if (maxIterations < 1) throw new Error("There must be at least one iteration for clipping.");
+    if (!Number.isInteger(maxIterations)) throw new Error("Number of iterations must be intiger value.");
```
tharvik:
```suggestion
    if (!Number.isInteger(maxIterations)) throw new Error("Number of iterations must be an integer.");
```

discojs/src/aggregator/byzantine.ts
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    // Step 1: Initialize v to average of previous aggregations
+    let v: WeightsContainer;
+    if (this.prevAggregate) {
+      v = this.prevAggregate;
+    } else {
+      // Use shape of the first contribution to create zero vector
+      const sample = currentContributions.values().next().value as WeightsContainer;
+      v = sample.map((t: tf.Tensor) => tf.zerosLike(t));
```
tharvik: better not to cast

```suggestion
      const first = currentContributions.values().next();
      if (first.done) throw new Error("zero sized contribution")
      v = first.value.map((t: tf.Tensor) => tf.zerosLike(t));
```

discojs/src/aggregator/byzantine.ts
```diff
@@ -0,0 +1,125 @@
...
+    }
+    // Step 2: Iterative Centered ClippingF
```
tharvik:
```suggestion
    // Step 2: Iterative Centered Clipping
```

tharvik (changes requested):
incredible work, that's so great! in addition I forgot about the `MultiRoundAggregator` and that's waay clearer to use this way.

can you also add a way to create a task with the byzantine aggregator? in `TaskCreationForm`, we need to add a way to select it, and with parameters you think are understable by laypersons

here goes a small patch for the core of disco (feel free to change it, I choose of using the default values but you know best what can be interesting to be set in the `Task`), adding a e2e test in server using this aggregator would also ensure that it works with the whole system.

```patch
diff --git a/discojs/src/aggregator/get.ts b/discojs/src/aggregator/get.ts
index 23b3bc88..49550bb6 100644
--- a/discojs/src/aggregator/get.ts
+++ b/discojs/src/aggregator/get.ts
@@ -1,5 +1,6 @@
 import type { DataType, Network, Task } from '../index.js'
 import { aggregator } from '../index.js'
+import { ByzantineRobustAggregator } from './byzantine.js';
 
 type AggregatorOptions = Partial<{
   scheme: Task<DataType, Network>["trainingInformation"]["scheme"]; // if undefined, fallback on task.trainingInformation.scheme
@@ -33,6 +34,8 @@ export function getAggregator(
   const scheme = options.scheme ?? task.trainingInformation.scheme
   
   switch (task.trainingInformation.aggregationStrategy) {
+    case 'byzantine':
+      return new ByzantineRobustAggregator()
     case 'mean':
       if (scheme === 'decentralized') {
         // If options are not specified, we default to expecting a contribution from all peers, so we set the threshold to 100%
diff --git a/discojs/src/default_tasks/cifar10.ts b/discojs/src/default_tasks/cifar10.ts
index 56ca523d..ba3ace5d 100644
--- a/discojs/src/default_tasks/cifar10.ts
+++ b/discojs/src/default_tasks/cifar10.ts
@@ -35,7 +35,7 @@ export const cifar10: TaskProvider<"image", "decentralized"> = {
         IMAGE_W: 224,
         LABEL_LIST: ['airplane', 'automobile', 'bird', 'cat', 'deer', 'dog', 'frog', 'horse', 'ship', 'truck'],
         scheme: 'decentralized',
-        aggregationStrategy: 'mean',
+        aggregationStrategy: 'byzantine',
         privacy: { clippingRadius: 20, noiseScale: 1 },
         minNbOfParticipants: 3,
         maxShareValue: 100,
diff --git a/discojs/src/task/training_information.ts b/discojs/src/task/training_information.ts
index 145fb223..391e0039 100644
--- a/discojs/src/task/training_information.ts
+++ b/discojs/src/task/training_information.ts
@@ -72,7 +72,10 @@ export namespace TrainingInformation {
 			.and(
 				z.union([
 					z.object({
-						aggregationStrategy: z.literal("mean"),
+						aggregationStrategy: z.union([
+							z.literal("byzantine"),
+							z.literal("mean"),
+						]),
 					}),
 					z.object({
 						aggregationStrategy: z.literal("secure"),
@@ -85,7 +88,10 @@ export namespace TrainingInformation {
 		federated: z
 			.object({
 				scheme: z.literal("federated"),
-				aggregationStrategy: z.literal("mean"),
+				aggregationStrategy: z.union([
+					z.literal("mean"),
+					z.literal("byzantine"),
+				]),
 			})
 			.merge(nonLocalNetworkSchema),
 		local: z.object({
```

discojs/src/aggregator/get.ts
```diff
@@ -33,6 +34,42 @@ export function getAggregator(
...
+    case 'byzantine': {
+        const {
+          clippingRadius = 1.0,
+          maxIterations = 1,
+          beta = 0.9,
+        }:{
+            clippingRadius?: number;
+            maxIterations?: number;
+            beta?: number;
+          } = task.trainingInformation;
```
tharvik: no need to explicit the type, it is already inferred

```suggestion
        } = task.trainingInformation;
```

discojs/src/aggregator/get.ts
```diff
@@ -33,6 +34,42 @@ export function getAggregator(
...
+            ...options,
+          };
+        } else {
+          options = {
+            roundCutOff: undefined,
+            threshold: 1,
+            thresholdType: "absolute",
+            ...options,
+          };
+        }
```
tharvik: duplicated from the "mean" case, consider extracting it into a dedicated variable, smth like
```ts
const networkOptions: Required<AggregatorOptions> = {
	scheme,
	roundCutOff: 0,
	threshold: 1,
	thresholdType: scheme === "decentralized" ? "relative" : "absolute",
	...options
};
```

discojs/src/task/training_information.ts
```diff
@@ -82,11 +88,24 @@ export namespace TrainingInformation {
...
+				z.union([
+					z.object({
+						aggregationStrategy: z.literal("mean"),
+					}),
+					z.object({
+						aggregationStrategy: z.literal("byzantine"),
+						clippingRadius: z.number().positive().optional().default(1.0),
+						maxIterations: z.number().int().positive().optional().default(1),
+						beta: z.number().min(0).max(1).optional().default(0.9),
+					}),
```
tharvik: duplicated, consider moving it to a dedicated schema (otherwise, we might have theses out of sync if we modify the defaults at some point)

webapp/src/components/task_creation_form/TaskCreationForm.vue
```diff
@@ -318,6 +318,7 @@
               >
                 <option value="mean">Mean</option>
                 <option value="secure">Secure</option>
+                <option value="byzantine">Byzantine</option>
```
tharvik: this select is only enabled when selecting a "decentralized" scheme, it works also with federated & byzantine now. we can disable it iff the scheme is local, and `v-if` the compatible aggregator

webapp/src/components/task_creation_form/TaskCreationForm.vue
```diff
@@ -641,17 +665,30 @@ const trainingInformationNetworks = z.union([
         z.object({
           aggregationStrategy: z.literal("mean"),
         }),
+        z.object({
```
tharvik, line 714: you also need to add the various parameters of this aggregator to the schema, otherwise we don't have validation of the fields (for eg, there isn't feedback when entering a negative momentum coefficient, but there is when choosing the number of epochs before aggregation) and I actually think that the unknown values in a zod schema are stripped (you can log the submitted task and see if it is still present)

webapp/src/components/task_creation_form/TaskCreationForm.vue
```diff
@@ -333,6 +334,50 @@
               </FormLabel>
             </FormLabel>
 
+            <!-- Byzantine Robust Aggregator Parameters -->
+            <FormLabel
+              v-show="aggregationStrategy === 'byzantine'"
+              label="Clipping radius (λ) for Centered Clipping"
+              type="required"
```
tharvik: hum, I see that new params are marked as required, but is it the case? I mean, can't we use the same reasonable defaults as set in `TrainingInformation` so as to not force the user to select a value?

tharvik (changes requested):

discojs/src/aggregator/get.ts
```diff
@@ -31,29 +32,47 @@ export function getAggregator(
...
-          roundCutOff: undefined, threshold: 1, thresholdType: 'absolute',
-          ...options
-        }
+    case 'byzantine': {
+        const {clippingRadius = 1.0, maxIterations = 1, beta = 0.9,
+        } = task.trainingInformation;
+
+        return new ByzantineRobustAggregator(
+          networkOptions.roundCutOff ?? 0,
+          networkOptions.threshold ?? 1,
```
tharvik: both are now fully defined

```suggestion
          networkOptions.roundCutOff,
          networkOptions.threshold,
```

webapp/src/components/task_creation_form/TaskCreationForm.vue
```diff
@@ -376,26 +427,6 @@
...
-                >
-                  <FormField
-                    name="trainingInformation.privacy.clippingRadius"
-                    placeholder="40"
-                    as="input"
-                    type="number"
-                  />
-                </FormLabel>
-              </div>
-            </FormLabel>
```
tharvik, line 398: why removing this `clippingRadius`? there are now two types of clipping (yeah, a bit confusing, if you've a better name, go for it) no?

tharvik (changes requested):
only one or two things to iron out: I'm currently unable to submit the form if I don't set the clipping radius for center clipping, dunno why but I suspect that it has smth to do with using the same name for two fields :shrug:

webapp/src/components/task_creation_form/TaskCreationForm.vue
mina5rovic: I don't think we use other clipping anywhere. Not so sure, but I will double check.

tharvik (approved):

PR #917, opened 2025-05-05
Hellaswag GPT evaluation

christinakopi:
HellaSwag evaluation for pretrained GPT-2 of transformers.js as well as our tfjs GPT-2.

The times that we get for our tfjs implementation of GPT-2 and the pretrained ONNX Xenova/gpt2 are the following:

|                           | **TFJS GPT-2 (nano)** | **ONNX GPT-2** |
|---------------------------|----------------|----------------|
| **Tokens per second**     | 2327           | 146            |

Didn't manage to get rid of the annoying warnings:
2025-05-05 18:25:03.655349674 [W:onnxruntime:, graph.cc:3490 CleanUnusedInitializersAndNodeArgs] Removing initializer '/transformer/h.8/mlp/act/Constant_4_output_0'. It is not used by any node and should be removed from the model.
2025-05-05 18:25:03.655354008 [W:onnxruntime:, graph.cc:3490 CleanUnusedInitializersAndNodeArgs] Removing initializer '/transformer/h.5/mlp/act/Constant_2_output_0'. It is not used by any node and should be removed from the model.
2025-05-05 18:25:03.655356999 [W:onnxruntime:, graph.cc:3490 CleanUnusedInitializersAndNodeArgs] Removing initializer '/transformer/h.8/mlp/act/Constant_2_output_0'. It is not used by any node and should be removed from the model.

even though I tried what's written in the issue:
https://github.com/microsoft/onnxruntime/issues/17377#issuecomment-2224889784


discojs/src/models/gpt/hellaswag.ts
```diff
@@ -0,0 +1,177 @@
...
+
+  await new Promise<void>((resolve, reject) => {
+    res.body?.pipe(fileStream);
+    res.body?.on('error', reject);
+    fileStream.on('error', reject);
+    fileStream.on('finish', () => resolve());
+  });
+}
+
+interface HellaSwagExample {
```
JulienVig: Could you add some documentation for this interface (e.g. what are the fields)?

discojs/src/models/gpt/hellaswag.ts
```diff
@@ -0,0 +1,177 @@
...
+import { List } from 'immutable';
+import { ONNXModel } from '../onnx.js';
+
+
+const HELLASWAG_URL = 'https://raw.githubusercontent.com/rowanz/hellaswag/master/data/hellaswag_val.jsonl';
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = path.dirname(__filename);
+const LOCAL_FILE = path.resolve(__dirname, '../../../../datasets/hellaswag_val.jsonl');
+
+async function fileExists(path: string): Promise<boolean> {
```
JulienVig: I think every function except `evaluate` can be made private (and not exported)

discojs/src/models/gpt/hellaswag.ts
```diff
@@ -0,0 +1,177 @@
...
+  });
+}
+
+interface HellaSwagExample {
+  ctx: string;
+  endings: string[];
+  label: number;
+}
+
+export async function* loadExamples(limit = 100): AsyncGenerator<HellaSwagExample> {
```
JulienVig: How do you set `limit` to load all the examples? Can you document this?

discojs/src/models/gpt/hellaswag.ts
```diff
@@ -0,0 +1,177 @@
...
+      const data = JSON.parse(line.trim());
+      yield { ctx: data.ctx, endings: data.endings, label: data.label };
+    } catch (e) {
+      console.error(`Failed to parse line ${count}:`, line);
+      throw e;
+    }
+  }
+}
+
+async function computeLogLikelihood(gpt: GPT, inputIds: number[], ctxLength: number): Promise<number> {
```
JulienVig: It would be nice to have a docstring and some comments within the code explaining how the loss is computed

discojs/src/models/onnx.ts
```diff
@@ -0,0 +1,107 @@
...
+  static async init_pretrained(modelName = 'Xenova/gpt2'): Promise<ONNXModel> {
+    const model = await AutoModelForCausalLM.from_pretrained(modelName);
+    return new ONNXModel(model);
+  }
+
+  getConfig(): any {
+    return this.model.config;
+  } 
+
+  async predict(batch: List<List<number>>): Promise<List<number>> {
```
JulienVig: When is the `predict` method used? I only see `getLogits` in `computeONNXLogLikelihood`

I think this method should make the assumption that the tokens are already padded/truncated and in the right format, i.e. that `batch` is a list of constant-length lists (as does the GPT class in models/gpt/index.ts). 
The preprocessing should be handled beforehand by the caller of the ONNXModel class. For example for GPT-tfjs it's done in processing/index.ts L.62:
```js
return d.map(text => processing.tokenize(tokenizer, text))
        .flatten()
        .batch(contextLength + 1, 1)
        .map((tokens) => [tokens.pop(), tokens.last()]) as
          Dataset<DataFormat.ModelEncoded[D]>;
```

discojs/src/models/onnx.ts
```diff
@@ -0,0 +1,107 @@
...
+
+    const predictions: number[] = logits.map((sequence: number[][]) => {
+      const lastLogits = sequence[sequence.length - 1];
+      return lastLogits.indexOf(Math.max(...lastLogits));
+    });
+  
+    return List(predictions);
+  }
+
+  async getLogits(batch: List<List<number>>): Promise<Tensor> {
```
JulienVig, line 73: Same thing here, getLogits shouldn't handle the preprocessing, only getting the attention mask and running forward

discojs/src/models/gpt/hellaswag_test.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+import { PreTrainedTokenizer } from '@xenova/transformers';
+import { GPT } from '../index.js';
+import { ONNXModel } from '../onnx.js';
+
+describe('HellaSwag Evaluator', () => {
+  it('evaluates tfjs GPT model', async () => {
+    const tokenizer = await PreTrainedTokenizer.from_pretrained('Xenova/gpt2');
+    const gpt = new GPT({seed: 42,}); // seed for reproducibility
+
+    const accuracy = await evaluate(gpt, tokenizer);  // evaluate the model on HellaSwag dataset
```
JulienVig: You can evaluate it on 2-3 examples here

discojs/src/models/gpt/hellaswag_test.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+  }).timeout(20000);
+});
+
+describe('HellaSwag Evaluator with Xenova GPT-2', () => {
+  it('evaluates the pretrained GPT-2 model', async () => {
+    const tokenizer = await PreTrainedTokenizer.from_pretrained('Xenova/gpt2');
+    const model = await ONNXModel.init_pretrained('Xenova/gpt2');
+    // console.log(await model.getConfig());
+
+    const accuracy = await evaluate(model, tokenizer);
```
JulienVig: same 2-3 examples

discojs/src/models/gpt/hellaswag.ts
JulienVig: If this function (and computeONNXLogLikelihood) are adapted from Karparthy's code can you link his code as source?

JulienVig (commented):
Thanks for the great PR! Aside from formatting comments, I think that ONNXModel's predict method shouldn't handle the preprocessing so it can be made simpler. We can discuss it via slack if you want.

cli/src/hellaswag_gpt.ts
```diff
@@ -0,0 +1,55 @@
...
+const hellaswag_path = path.join('..', 'datasets', 'hellaswag_val.jsonl');
+const logFile = path.join('..', 'datasets', 'LogFile_hellaswag.txt');
+const logLines: string[] = [];
+
+function log(message: string) {
+    console.log(message);
+    logLines.push(message);
+}
+
+const hellaswagDataset: models.HellaSwagDataset = await loadHellaSwag(hellaswag_path)
```
JulienVig: ~~hellaswag should be downloaded here rather than assuming that it has already been downloaded~~ (or decide to not download it at all and let `loadHellaSwag` fetch it with `fs.createReadStream(HELLASWAG_URL)`)
Edit: Actually `loadHellaSwag` should be handling all the logic -> it shouldn't take any argument and take care of downloading the file if it doesn't exist or read it.
Edit 2: `loadHellaSwag` could handle the `limit` parameter instead of `evaluate` so that if we only want to evaluate on 2 examples, we don't have to load the whole dataset.

discojs-node/src/hellaswag.spec.ts
```diff
@@ -0,0 +1,22 @@
+import { expect } from 'chai';
+import { load as loadHellaSwag } from './hellaswag.js';
+import fs from 'fs';
+
+describe('HellaSwag parser', () => {
+  it('should load all examples and return them as an array', async () => {
+    const filename: string = "../datasets/hellaswag_val.jsonl"
+    expect(fs.existsSync(filename)).to.be.true;
+
```
JulienVig: loadHellaswag should handle this logic

discojs/src/models/hellaswag.ts
```diff
@@ -0,0 +1,222 @@
...
+  const fileStream = fs.createWriteStream(path_);
+
+  await new Promise<void>((resolve, reject) => {
+    res.body?.pipe(fileStream);
+    res.body?.on('error', reject);
+    fileStream.on('error', reject);
+    fileStream.on('finish', () => resolve());
+  });
+}
+
```
JulienVig: All this is specific to Node.js so it should only be in discojs-node

discojs/src/models/hellaswag.ts
```diff
@@ -0,0 +1,222 @@
...
+ * @param dataset - An array of HellaSwagExample to evaluate on
+ * @param limit - Number of examples to evaluate (default: all)
+ * @param print - Whether to print results (default: true)
+ * @returns The accuracy of the model on the dataset
+ */
+export async function evaluate(
+  model: ModelType,
+  tokenizer: Tokenizer,
+  dataset: HellaSwagExample[],
+  limit = -1,
```
JulienVig: I suggest that we move the `limit` argument to loadHellaswag to allow only loading a subset. Then evaluate always loops over all the array it's provided with

discojs/src/models/hellaswag.ts
```diff
@@ -0,0 +1,222 @@
...
+  print = true
+): Promise<number> {
+  await downloadHellaSwag(HELLASWAG_URL);
```
JulienVig: this line can be removed since the dataset is provided as argument

discojs-web/src/hellaswag.ts
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * Loads the HellaSwag dataset from a .jsonl file as a Blob
+ *
+ * @param file - A Blob representing the HellaSwag dataset (.jsonl format)
+ * @returns A fully loaded HellaSwagDataset
+ */
+export async function load(file: Blob): Promise<models.HellaSwagDataset> {
+  // Read the file as text
+  const text = await file.text();
```
JulienVig: Same here, load should handle fetching hellaswag, (HELLASWAG_URL can be exported by discojs and imported here)

JulienVig (approved):
All good for me thanks a lot for the hard work!

PR #927, opened 2025-06-26
ONNX to Tensorflow.js conversion of GPT-2

christinakopi:
Roadmap:
- [x] ensure bijection between onnx and tfjs weights
- [ ] add progress indicator to hellaswag benchmark
- [x] move filesystem logic to discojs-node and cli
- [x] automate downloading onnx model from hub
- [x] convert python script to nodejs and web-compatible logic
- [ ] experiment with webgpu for hellaswag
- [ ] experiment with converting tfjs back to onnx

The table below shows the accuracy along with the time of the evaluation on the whole HellaSwag for each model tested: 
| Model                             | Accuracy | Eval Time (s) |
|----------------------------------|----------|---------------|
| TFJS GPT (gpt-nano)              | 24.67%   | 1390.25       |
| Xenova GPT-2 (ONNX)              | 29.03%   | 22767.03      |
| Loaded TFJS GPT (from ONNX)      | 28.41%   | 12523.59      |


martinjaggi (commented):
amazing work, thanks a lot!
later i'd be very curious to check if also the conversion in the reverse direction (to onnx) works fine.

cli/src/hellaswag_gpt.ts
```diff
@@ -1,50 +1,102 @@
...
+import { fileURLToPath } from 'url';
+import { parse } from 'ts-command-line-args'
+
 import '@tensorflow/tfjs-node';
 import fs from 'node:fs';
 import path from 'node:path';
-import { Tokenizer, models } from '@epfml/discojs';
+import { models, serialization, Tokenizer } from '@epfml/discojs';
 import { loadHellaSwag } from '@epfml/discojs-node';
+// import { AutoTokenizer } from '@xenova/transformers';
```
tharvik: commented

cli/src/hellaswag_gpt.ts
```diff
@@ -1,50 +1,102 @@
+// import fs from 'fs';
```
tharvik: commented

cli/src/hellaswag_gpt.ts
tharvik: commented

cli/src/hellaswag_gpt.ts
```diff
@@ -1,50 +1,102 @@
...
 
-const logFile = path.join('..', 'datasets', 'LogFile_hellaswag.txt');
-const logLines: string[] = [];
+const __dirname = dirname(fileURLToPath(import.meta.url));
 
+const logLines: string[] = [];
 function log(message: string) {
     console.log(message);
     logLines.push(message);
 }
```
tharvik: we don't need a log system for the CLI, we can simply output to the console, no?

cli/src/hellaswag_gpt.ts
```diff
@@ -1,50 +1,102 @@
...
+    switch (args.model) {
+        case 'onnx':
+            log("Using ONNX pretrained model Xenova/gpt2")
+            model = await models.ONNXModel.init_pretrained('Xenova/gpt2');
+            break;
+            case 'gpt-tfjs-random':
+            log("Using GPT-TFJS with random initialization")
+            model = new models.GPT({ seed: 42 });
+            break;
+            case 'gpt-tfjs-pretrained':
```
tharvik: that was confusing
```suggestion
        case 'gpt-tfjs-random':
            log("Using GPT-TFJS with random initialization")
            model = new models.GPT({ seed: 42 });
            break;
        case 'gpt-tfjs-pretrained':
```

cli/src/hellaswag_gpt.ts
```diff
@@ -1,50 +1,102 @@
...
+    pretrainedModelPath: string
+    help?: boolean
 }
 
 async function main(): Promise<void> {
-    fs.writeFileSync(logFile, '', 'utf-8'); // Clear old log file
+    const defaultPretrainedModelPath = path.join(__dirname, "..", "..", "onnx-converter", "assets", "model.json")
+    const args = parse<HellaSwagArgs>({
+        model: {
+            type: (raw: string) => raw as ModelType,
```
tharvik: casting isn't nice, especially for user interfaces, better to implement so value checking with a `switch`

cli/src/hellaswag_gpt.ts
```diff
@@ -1,50 +1,102 @@
...
+            if (args.pretrainedModelPath === undefined) {
+                throw new Error("If choosing gpt-tfjs-pretrained, provide the relative path to the TF.js pretrained model `pretrainedModelPath")
+            }
+            const encodedModel = await fsPromise.readFile(args.pretrainedModelPath);
+            model = await serialization.model.decode(encodedModel) as models.GPT;
+            break;
+    } 
+    await evaluateModel(model, args.numDataPoints);
 
     fs.writeFileSync(logFile, logLines.join('\n'), 'utf-8');
```
tharvik: no sync write, we are in an async function so better to use `fsPromise`

datasets/.gitignore
tharvik, line 1: I don't think this is needed as the dataset is fetched everytime (I would argue against doing that)

discojs/src/default_tasks/titanic.ts
```diff
@@ -12,10 +12,10 @@ export const titanic: TaskProvider<"tabular", "federated"> = {
         title: 'Titanic Prediction',
         summary: {
           preview: "The Titanic classification task is one of the main entrypoints into machine learning. Using passenger data (name, age, gender, socio-economic class, etc), the goal is to identify who was more likely to survive the infamous shipwreck.",
-          overview: "The original competition can be found on <a  target='_blank' class='underline text-blue-400' href='https://www.kaggle.com/c/titanic'>Kaggle</a> and a link to the training set can be found here <a target='_blank' class='underline text-blue-400' href='https://storage.googleapis.com/deai-313515.appspot.com/example_training_data/titanic_train.csv'>here</a>."
+          overview: ""
```
tharvik: no text?

discojs/src/models/gpt/layers.ts
```diff
@@ -289,31 +293,32 @@ export class GELU extends tf.layers.Layer {
 tf.serialization.registerClass(GELU)
 
 export type MLPConfig = ConstructorParameters<typeof tf.layers.Layer>[0] &
-  Required<ModelSize> & Record<'contextLength' | 'residDrop' | 'nLayer' | 'seed', number>
+  Required<ModelSize> & Record<'contextLength' | 'residDrop' | 'nLayer' | 'seed', number> & { name: string }
```
tharvik: don't think that's needed, I remove it and it still compiles
```suggestion
  Required<ModelSize> & Record<'contextLength' | 'residDrop' | 'nLayer' | 'seed', number>
```

onnx-converter/.gitignore
tharvik, line 1: to be quite pedantic, you can prepend ignore lines with `/` to note that it is starting from the root of this gitignore and not any file matching the line. that helps finding out when `node_modules` are generated in weird places which will confuse tools.

onnx-converter/README.md
```diff
@@ -0,0 +1,36 @@
...
+export default onnxModule;
+```
+4. Create `onnx-proto.d.ts` with the matching TypeScript definition:
+```ts
+export { onnx } from './onnx.js';
+declare const onnxModule: {
+  onnx: typeof import('./onnx.js').onnx;
+};
+export default onnxModule;
+```
```
tharvik: I don't think we need to duplicate code in markdown, we can rather link the files in the repo

onnx-converter/.gitignore
```diff
@@ -0,0 +1,3 @@
+node_modules
```
tharvik: as it is part of the workspace, everything should get hoisted to the root node_modules

onnx-converter/package.json
```diff
@@ -0,0 +1,21 @@
...
+    "convert_onnx": "npm run build && node dist/convert_onnx.js",
+    "build": "tsc && cp -r src/protobuf dist",
+    "lint": "npx eslint .",
+    "test": ": nothing"
+  },
+  "author": "",
+  "license": "ISC",
+  "dependencies": {
+    "@epfml/discojs-node": "*"
+  },
```
tharvik: do we need the -node version? only `@epfml/discojs` is imported

onnx-converter/package.json
```diff
@@ -0,0 +1,21 @@
...
+  "devDependencies": {
+    "nodemon": "3",
+    "ts-command-line-args": "2"
+  }
```
tharvik, line 17: unused

onnx-converter/package.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "onnx-converter",
+  "private": true,
+  "type": "module",
+  "main": "dist/gpt2_from_onnx.js",
+  "scripts": {
+    "convert_onnx": "npm run build && node dist/convert_onnx.js",
+    "build": "tsc && cp -r src/protobuf dist",
```
tharvik: with the new typescript project layout, we need to explicit use `--build` otherwise it doesn't nothing

```suggestion
    "build": "tsc --build && cp -r src/protobuf dist",
```

onnx-converter/src/convert_onnx.ts
```diff
@@ -0,0 +1,148 @@
...
+            tensor.rawData.byteOffset + tensor.rawData.byteLength
+        );
+        if (tensor.dataType != onnx.TensorProto.DataType.FLOAT) {
+            throw new Error("found protobuf data type different from expected float 32.")
+        }
+        return new Float32Array(buffer);
+    }
+    // Fallback to specific field arrays if rawData is empty
+    console.log("WARNING: protobuf raw data is empty, falling back on specific data fields.")
+    if (tensor.floatData && tensor.floatData.length > 0) return new Float32Array(tensor.floatData);
```
tharvik: that look to me as more relevant field than using `rawData`, why not using it first (and drop the warning)?

onnx-converter/src/convert_onnx.ts
```diff
@@ -0,0 +1,148 @@
...
+
+  console.log("Initializing a new TFJS GPT-2 model...")
+  if (preTrainedWeights.size !== onnxTfjsMapping.size)
+      throw new Error(`Expected to load ${onnxTfjsMapping.size} weights, but loaded ${preTrainedWeights.size}.`);
+  
+  // Overwrite the GPT-TF.js model weights with the ONNX weights
+  if (gptLayersModel.weights.length !== onnxTfjsMapping.size) 
+    throw new Error(`Mismatch between TFJS and ONNX weight mapping weights.`);
+    
+  const finalWeights = gptLayersModel.weights.map((weight, _i) => {
```
tharvik:
```suggestion
  const finalWeights = gptLayersModel.weights.map((weight) => {
```

onnx-converter/src/convert_onnx.ts
```diff
@@ -0,0 +1,148 @@
...
+import { dirname } from 'path';
+import path from 'node:path';
+import fsPromise from 'node:fs/promises';
+import * as tf from '@tensorflow/tfjs-node';
+
+import { models, serialization } from "@epfml/discojs";
+
+const __dirname = dirname(fileURLToPath(import.meta.url));
+const ASSET_FOLDER = path.join(__dirname, "..", "assets");
+const OUTPUT_FILENAME = path.join(ASSET_FOLDER, "model.json");
```
tharvik: can we more simply store it in the current directory and so letting choose the user where to have it?

onnx-converter/tsconfig.lib.json
```diff
@@ -0,0 +1,6 @@
+{
+	"extends": "../tsconfig.base.json",
+	"compilerOptions": { "outDir": "dist" },
+	"include": ["src"],
+	"exclude": ["**/*.spec.ts"]
```
tharvik: we don't have tests anyway
```suggestion
```

cli/src/hellaswag_gpt.ts
```diff
@@ -1,50 +1,102 @@
...
+            defaultValue: -1
+        },
+        logFile: {
+            type: String,
+            description: 'Relative path to the log file, default to ./hellaswag.log', defaultValue: 'hellaswag.log'
+        },
+        pretrainedModelPath: {
+            type: String,
+            description: 'If specifying gpt-tfjs-pretrained, provide the relative path to the TF.js pretrained model',
+            defaultValue: defaultPretrainedModelPath
```
tharvik: single use, we can inline it IMO

tharvik (approved):
haa, that's great work, well done in hacking around the onnx protobuf!
nothing blocking (except the `tsc --build` thing), only a few nitpicks and questions here and there

onnx-converter/src/convert_onnx.ts
JulienVig: Because it's empty in the default scenario, only rawData has the data and this is just a backup option for some unforeseen scenario. We could potentially drop this if case since I don't know when this would come into play.

tharvik: hum, if we don't really know when it'd be useful, better to wait for an example/model causing issue (this is [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it))

datasets/.gitignore
JulienVig: I changed the discojs-node load method to cache it locally!

discojs/src/default_tasks/titanic.ts
JulienVig: oops I think I got distracted, thanks!

PR #944, opened 2025-08-13
webapp/sidebar: rm empty legacy file

JulienVig:
An empty `ModelLibrary.vue` file was left behind in [this commit](https://github.com/epfml/disco/commit/f4a24f56777ab36a651233ca076481b4853f31ec). The current `ModelLibrary.vue` is under `webapp/src/components/testing`.


tharvik (approved):
love cleanup, the less code, the less bug

PR #961, opened 2025-10-16
bump deps

tharvik:
* update all dependencies!
* replace apexchart by graph.js because of some weird SVG generation bug I encountered
* commonize all the lint into latest eslint (no need anymore to run it per package, simply do it at root w/ `npm run lint`)
* update tsconfig to use [project references](https://www.typescriptlang.org/docs/handbook/project-references.html) which makes it faster and allow to jump to the correct code when using go to ref!


webapp/src/components/training/TrainingInformation.vue
```diff
@@ -232,85 +250,72 @@ const lossSeries = computed(() =>
...
-      show: true,
-      colors: darkMode ? "#ffffff" : "#000000",
-    },
-    axisTicks: {
-      show: true,
-      color: darkMode ? "#ffffff" : "#000000",
+
+const chartsLabels = computed<string[]>(() =>
+  Range(1, allEpochs.value.size + 1)
+    .map((e) => `Epoch ${e}`)
```
JulienVig:
```suggestion
    .map((e) => `${e}`)
```

Can we remove the "Epoch" prefix? I find the axis too crowded when there are many epochs

webapp/src/components/testing/TestSteps.vue
```diff
@@ -228,59 +226,59 @@ const visitedSamples = computed<number>(() => {
   }
 });
 
-const confusionMatrix = computed<{ [key: string]: { [key: string]: number } } | undefined>(() => {
-  if (tested.value === undefined) return undefined;
+const confusionMatrix = computed<Map<string, Map<string, number>> | undefined>(
```
JulienVig, line 229: confusion_matrix
The confusion matrix is not being displayed correctly

webapp/src/assets/css/tailwind.css
JulienVig, line 1: The borders are now drawing in a darker gray than before:
borders
This is how it looks like on discolab currently:
Screenshot 2025-10-23 at 11 57 10
Could we keep the lighter borders?

JulienVig (approved):
A few UI comments but aside from that it looks great! Great to have typescript project references

webapp/src/assets/css/tailwind.css
tharvik: ha yes, there was a [default border color in the previous version of tailwindcss](https://tailwindcss.com/docs/upgrade-guide#default-border-color). adding it where needed

PR #983, opened 2025-11-14
Installation issues

mina5rovic:
1. Normalize LF line endings for shell scripts -> This change ensures that all shell scripts in the repository use consistent LF line endings, which are required for proper execution when being cloned using Windows

2. Add firewall issue to FAQ -> solution is to turn it off or delete rules that block npm


datasets/populate
tharvik, line 1: please keep it executable, it has a [shebang](https://en.wikipedia.org/wiki/Shebang_(Unix)) to instruct the kernel on how to start it

DEV.md
```diff
@@ -49,8 +49,10 @@ flowchart LR
 The following instructions will install the required dependencies, build Disco.js and launch a DISCO server and a web client. If you run into any sort of trouble check our [FAQ](./docs/FAQ.md); otherwise please create a new issue.
 
 **1.** Clone the repository
+(if using WSL or you are cloning the repository from Windows, make sure there is no file ending conversion to CRLF by running first of the following commands)
```
tharvik: I think we can avoid this extra step by adding a line in `.gitattributes` setting a default text handling. that's always nicer for users
https://docs.github.com/en/get-started/git-basics/configuring-git-to-handle-line-endings#per-repository-settings

tharvik (changes requested):
thanks for the work! a few tweaks and we're good to go!

tharvik (approved):
perfect, thanks!

PR #997, opened 2025-12-04
Optimized differential privacy implementation

ahzero7d1:
1. Gaussian noise calibration implementation
     - Added Gaussian noise calibration for differential privacy (DP) based on privacy budget (epsilon), failure possibility (delta) and weight clipping radius

2. Adaptive clipping mechanism
     - Implemented adaptive clipping radius using historical moving average of weight updates
3. cli testing tool modification

4. Webapp interface updates
     - Added UI fields for the new DP parameters

### DIFF
--- a/cli/src/args.ts
+++ b/cli/src/args.ts
@@ -10,6 +10,10 @@ interface BenchmarkArguments {
   epochs: number
   roundDuration: number
   batchSize: number
+  validationSplit: number
+  epsilon?: number
+  delta?: number
+  dpDefaultClippingRadius?: number
   save: boolean
   host: URL
 }
@@ -28,6 +32,10 @@ const unsafeArgs = parse<BenchmarkUnsafeArguments>(
     epochs: { type: Number, alias: 'e', description: 'Number of epochs', defaultValue: 10 },
     roundDuration: { type: Number, alias: 'r', description: 'Round duration (in epochs)', defaultValue: 2 },
     batchSize: { type: Number, alias: 'b', description: 'Training batch size', defaultValue: 10 },
+    validationSplit : { type: Number, alias: 'v', description: 'Validation dataset ratio', defaultValue: 0.2 },
+    epsilon: { type: Number, alias: 'n', description: 'Privacy budget', optional: true, defaultValue: undefined},
+    delta: { type: Number, alias: 'd', description: 'Probability of failure, slack parameter', optional: true, defaultValue: undefined},
+    dpDefaultClippingRadius: {type: Number, alias: 'f', description: 'Default clipping radius for DP', optional: true, defaultValue: undefined},
     save: { type: Boolean, alias: 's', description: 'Save logs of benchmark', defaultValue: false },
     host: {
       type: (raw: string) => new URL(raw),
@@ -52,6 +60,7 @@ const supportedTasks = Map(
       defaultTasks.simpleFace,
       defaultTasks.titanic,
       defaultTasks.tinderDog,
+      defaultTasks.mnist,
     ).map(
       async (t) =>
         [(await t.getTask()).id, t] as [
@@ -77,10 +86,29 @@ export const args: BenchmarkArguments = {
       task.trainingInformation.batchSize = unsafeArgs.batchSize;
       task.trainingInformation.roundDuration = unsafeArgs.roundDuration;
       task.trainingInformation.epochs = unsafeArgs.epochs;
+      task.trainingInformation.validationSplit = unsafeArgs.validationSplit;
 
       // For DP
-      // TASK.trainingInformation.clippingRadius = 10000000
-      // TASK.trainingInformation.noiseScale = 0
+      const {dpDefaultClippingRadius, epsilon, delta} = unsafeArgs;
+
+      if (
+        // dpDefaultClippingRadius !== undefined &&
+        epsilon !== undefined &&
+        delta !== undefined
+      ){
+        if (task.trainingInformation.scheme === "local")
+          throw new Error("Can't have differential privacy for local training");
+
+        const defaultRadius = dpDefaultClippingRadius ? dpDefaultClippingRadius : 1;
+          
+        // for the case where privacy parameters are not defined in the default tasks
+        task.trainingInformation.privacy ??= {}
+        task.trainingInformation.privacy.differentialPrivacy = {
+          clippingRadius: defaultRadius,
+          epsilon: epsilon,
+          delta: delta,
+        };
+      }
 
       return task;
     },
--- a/cli/src/data.ts
+++ b/cli/src/data.ts
@@ -1,6 +1,7 @@
 import path from "node:path";
-import { Dataset, processing } from "@epfml/discojs";
-import type {
+import { promises as fs } from "fs";
+import { Dataset, processing, defaultTasks } from "@epfml/discojs";
+import {
   DataFormat,
   DataType,
   Image,
@@ -9,7 +10,7 @@ import type {
 import { loadCSV, loadImage, loadImagesInDir } from "@epfml/discojs-node";
 import { Repeat } from "immutable";
 
-async function loadSimpleFaceData(): Promise<Dataset<DataFormat.Raw["image"]>> {
+async function loadSimpleFaceData(userIdx: number, totalClient: number): Promise<Dataset<DataFormat.Raw["image"]>> {
   const folder = path.join("..", "datasets", "simple_face");
 
   const [adults, childs]: Dataset<[Image, string]>[] = [
@@ -17,10 +18,12 @@ async function loadSimpleFaceData(): Promise<Dataset<DataFormat.Raw["image"]>> {
     (await loadImagesInDir(path.join(folder, "child"))).zip(Repeat("child")),
   ];
 
-  return adults.chain(childs);
+  const combinded = adults.chain(childs);
+
+  return combinded.filter((_, i) => i % totalClient === userIdx);
 }
 
-async function loadLusCovidData(): Promise<Dataset<DataFormat.Raw["image"]>> {
+async function loadLusCovidData(userIdx: number, totalClient: number): Promise<Dataset<DataFormat.Raw["image"]>> {
   const folder = path.join("..", "datasets", "lus_covid");
 
   const [positive, negative]: Dataset<[Image, string]>[] = [
@@ -32,7 +35,11 @@ async function loadLusCovidData(): Promise<Dataset<DataFormat.Raw["image"]>> {
     ),
   ];
 
-  return positive.chain(negative);
+  const combined: Dataset<[Image, string]> = positive.chain(negative);
+
+  const sharded = combined.filter((_, i) => i % totalClient === userIdx);
+
+  return sharded;
 }
 
 function loadTinderDogData(split: number): Dataset<DataFormat.Raw["image"]> {
@@ -59,25 +66,89 @@ function loadTinderDogData(split: number): Dataset<DataFormat.Raw["image"]> {
     });
 }
 
+async function loadExtCifar10(userIdx: number): Promise<Dataset<[Image, string]>> {
+  const CIFAR10_LABELS = Array.from(await defaultTasks.cifar10.getTask().then(t => t.trainingInformation.LABEL_LIST));
+  const folder = path.join("..", "datasets", "extended_cifar10");
+  const clientFolder = path.join(folder, `client_${userIdx}`);
+
+  return new Dataset(async function*(){
+    const entries = await fs.readdir(clientFolder, {withFileTypes: true});
+
+    const items = entries
+        .flatMap((e) => {
+          const m = e.name.match(
+            /^image_(\d+)_label_(\d+)\.png$/i
+          );
+          if (m === null) return [];
+          const labelIdx = Number.parseInt(m[2], 10);
+
+          if(labelIdx >= CIFAR10_LABELS.length)
+            throw new Error(`${e.name}: too big label index`);
+
+          return {
+            name: e.name,
+            label: CIFAR10_LABELS[labelIdx],
+          };
+        })
+        .filter((x) => x !== null)
+
+      for (const {name, label} of items){
+        const filePath = path.join(clientFolder, name);
+        const image = await loadImage(filePath);
+        yield [image, label] as const;
+      }
+  })
+}
+
+function loadMnistData(split: number): Dataset<DataFormat.Raw["image"]>{
+  const folder = path.join("..", "datasets", "mnist", `${split + 1}`);
+  return loadCSV(path.join(folder, "labels.csv"))
+    .map(
+      (row) =>
+        [
+          processing.extractColumn(row, "filename"),
+          processing.extractColumn(row, "label"),
+        ] as const,
+    )
+    .map(async ([filename, label]) => {
+      try {
+        const image = await Promise.any(
+          ["png", "jpg", "jpeg"].map((ext) =>
+            loadImage(path.join(folder, `${filename}.${ext}`)),
+          ),
+        );
+        return [image, label];
+      } catch {
+        throw Error(`${filename} not found in ${folder}`);
+      }
+    });
+}
+
 export async function getTaskData<D extends DataType>(
 	taskID: Task.ID,
 	userIdx: number,
+  totalClient: number
 ): Promise<Dataset<DataFormat.Raw[D]>> {
   switch (taskID) {
     case "simple_face":
-      return (await loadSimpleFaceData()) as Dataset<DataFormat.Raw[D]>;
+      return (await loadSimpleFaceData(userIdx, totalClient)) as Dataset<DataFormat.Raw[D]>;
     case "titanic":
-      return loadCSV(
+      const titanicData = loadCSV(
         path.join("..", "datasets", "titanic_train.csv"),
       ) as Dataset<DataFormat.Raw[D]>;
+      return titanicData.filter((_, i) => i % totalClient === userIdx);
     case "cifar10":
       return (
         await loadImagesInDir(path.join("..", "datasets", "CIFAR10"))
       ).zip(Repeat("cat")) as Dataset<DataFormat.Raw[D]>;
     case "lus_covid":
-      return (await loadLusCovidData()) as Dataset<DataFormat.Raw[D]>;
+      return (await loadLusCovidData(userIdx, totalClient)) as Dataset<DataFormat.Raw[D]>;
     case "tinder_dog":
       return loadTinderDogData(userIdx) as Dataset<DataFormat.Raw[D]>;
+    case "extended_cifar10":
+      return (await loadExtCifar10(userIdx)) as Dataset<DataFormat.Raw[D]>;    
+    case "mnist":
+      return loadMnistData(userIdx) as Dataset<DataFormat.Raw[D]>;
     default:
       throw new Error(`Data loader for ${taskID} not implemented.`);
   }
--- a/discojs/src/dataset/dataset.ts
+++ b/discojs/src/dataset/dataset.ts
@@ -237,6 +237,60 @@ export class Dataset<T> implements AsyncIterable<T> {
   cached(): Dataset<T> {
     return new CachingDataset(this.#content);
   }
+
+  /** Shuffles the Dataset instance within certain window size */
+  shuffle(windowSize: number){
+    if (!Number.isInteger(windowSize) || windowSize < 1){
+      throw new Error("Shuffle window size should be a positive integer");
+    }
+
+    return new Dataset(
+      async function*(this: Dataset<T>){
+        const iter = this[Symbol.asyncIterator]();
+        const buffer: T[] = [];
+
+        // 1. Construct the initial buffer
+        while (buffer.length < windowSize){
+          const n = await iter.next();
+          if (n.done) break;
+          buffer.push(n.value);
+        }
+
+        // 2. Shuffle
+        while (buffer.length > 0){
+          const pick = Math.floor(Math.random() * buffer.length);
+          const chosen = buffer[pick];
+
+          const n = await iter.next();
+
+          if (n.done){
+            // move the last element to the pick position
+            buffer[pick] = buffer.pop() as T;
+          }else{
+            buffer[pick] = n.value;
+          }
+
+          yield chosen;
+        }
+      }.bind(this)
+    );
+  }
+
+  /** filter the indices according to the splitting condition */
+  filter(
+    condition: (value: T, index: number) => boolean | Promise<boolean>
+  ): Dataset<T>{
+    return new Dataset<T>(async function* (this: Dataset<T>): AsyncGenerator<T, void, unknown>{
+      let i = 0;
+      for await(const v of this){
+        if (await condition(v, i)){
+          yield v;
+        }
+        i += 1
+      }
+    }.bind(this));
+  }
+
 }
 
 /**
--- a/discojs/src/privacy.ts
+++ b/discojs/src/privacy.ts
@@ -1,39 +1,85 @@
 import * as tf from "@tensorflow/tfjs";
 
-import type { WeightsContainer } from "./index.js";
+import { WeightsContainer } from "./index.js";
 
-async function frobeniusNorm(weights: WeightsContainer): Promise<number> {
-  const squared = await weights
-    .map((w) => w.square().sum())
-    .reduce((a, b) => a.add(b))
-    .data();
-  if (squared.length !== 1) throw new Error("unexpected weights shape");
+import type { WeightNormHistory } from "./training/trainer.js";
 
-  return Math.sqrt(squared[0]);
+/** Computes the Frobenius norm of the given weights. */
+export async function frobeniusNorm(weights: tf.Tensor): Promise<number> {
+	const squared = await weights.square().sum().data();
+	if (squared.length !== 1) throw new Error("unexpected weights shape");
+	return Math.sqrt(squared[0]);
 }
 
-/** Scramble weights */
-export function addNoise(
-  weights: WeightsContainer,
-  deviation: number,
-): WeightsContainer {
-  const variance = Math.pow(deviation, 2);
-  return weights.map((w) => w.add(tf.randomNormal(w.shape, 0, variance)));
+/** ALDP-FL implementation */
+// Conditions need to be added for the first three epochs -> get the avg update from all of the available previous updates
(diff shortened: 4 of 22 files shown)


cli/src/args.ts
```diff
@@ -77,10 +85,34 @@ export const args: BenchmarkArguments = {
...
+
+      if (
+        dpDefaultClippingRadius !== undefined ||
+        noiseScale !== undefined ||
+        delta !== undefined
+      ){
+        if(
+          task.trainingInformation.scheme === "decentralized" ||
+          task.trainingInformation.scheme === "federated"
+        ){
```
tharvik: this will silently discard user privacy settings if other schemes, better to throw

```suggestion
        if (task.trainingInformation.scheme === "local")
          throw new Error("can't have differential privacy for local training");
```

cli/src/args.ts
```diff
@@ -77,10 +85,34 @@ export const args: BenchmarkArguments = {
...
+
+          if (dpDefaultClippingRadius !== undefined){
+            privacy.dpDefaultClippingRadius = dpDefaultClippingRadius;
+          }
+          if (noiseScale !== undefined){
+            privacy.noiseScale = noiseScale;
+          }
+          if (delta !== undefined){
+            privacy.delta = delta;
+          }
```
tharvik: setting a field to undefined is akin to not setting it at all (at least in our handling of it)

```suggestion
          task.trainingInformation.privacy = {
            dpDefaultClippingRadius,
            noiseScale,
            delta,
          };
```

models/cifar10/model.json
tharvik, line 1: theses /models files are not needed in the repo, they were probably generated by a server instance

discojs/src/dataset/dataset.ts
```diff
@@ -237,6 +238,60 @@ export class Dataset<T> implements AsyncIterable<T> {
   cached(): Dataset<T> {
     return new CachingDataset(this.#content);
   }
+
+  /** Shuffles the Dataset instance within certain window size */
+  shuffle(windowSize: number){
+    if (windowSize < 1){
+      throw new Error("Shuffle window size should be positive");
+    }
```
tharvik:
```suggestion
    if (!Number.isInteger(windowSize) || windowSize < 1)
      throw new Error("Shuffle window size should be a positive positive");
```

discojs/src/dataset/dataset.ts
```diff
@@ -237,6 +238,60 @@ export class Dataset<T> implements AsyncIterable<T> {
...
+          const chosen = buffer[pick];
+
+          const n = await iter.next();
+
+          if (n.done){
+            // swap the picked slot with the last slot in the window
+            // remove the last slot in the window
+            const last = buffer.length - 1;
+            [buffer[pick], buffer[last]] = [buffer[last], buffer[pick]];
+            buffer.pop();
```
tharvik: as we `pop` the last element anyway, we can simplify it as (`pop` returns the last element or undef if empty)
```suggestion
            buffer[pick] = buffer.pop() as T; // cast as buffer can't be empty
```

cli/src/data.ts
```diff
@@ -59,25 +66,79 @@ function loadTinderDogData(split: number): Dataset<DataFormat.Raw["image"]> {
     });
 }
 
+async function readLabels(labelsPath:string): Promise<string[]>{
+  const raw = await fs.readFile(labelsPath, "utf8");
+  return raw.split(/\r?\n/).map(s => s.trim()).filter(s => s.length > 0);
+}
```
tharvik: unused
```suggestion
```

cli/src/data.ts
```diff
@@ -59,25 +66,79 @@ function loadTinderDogData(split: number): Dataset<DataFormat.Raw["image"]> {
...
+
+  return new Dataset<[Image, string]>(async function*(){
+    const entries = await fs.readdir(clientFolder, {withFileTypes: true});
+
+    const items = entries
+        .map((e) => {
+          const m = e.name.match(
+            /^image_(\d+)_label_(\d+)\.png$/i
+          );
+          if (!m) return null;
```
tharvik: this `map(…).filter(…)` is quite common and has, in most languages, a nicer way to achieved it (and reduce the runtime cost of double trasversal). in js, it is `flatMap` where you can return an array of valid values

```suggestion
        .flatMap((e) => {
          const m = e.name.match(
            /^image_(\d+)_label_(\d+)\.png$/i
          );
          if (m === null) return [];
```

cli/src/data.ts
```diff
@@ -59,25 +66,79 @@ function loadTinderDogData(split: number): Dataset<DataFormat.Raw["image"]> {
...
+
+          const imgIdx = Number(m[1]);
+          const labelIdx = Number(m[2]);
```
tharvik: better to use `Number.parseInt` as to avoid the weird coercions rules of js

```suggestion
          const imgIdx = Number.parseInt(m[1], 10);
          const labelIdx = Number.parseInt(m[2], 10);
```

cli/src/data.ts
```diff
@@ -59,25 +66,79 @@ function loadTinderDogData(split: number): Dataset<DataFormat.Raw["image"]> {
...
+
+          if(
+            !Number.isInteger(labelIdx) ||
+            labelIdx < 0 ||
+            labelIdx >= CIFAR10_LABELS.length
+          ){
+            return null;
```
tharvik: I would throw instead as to avoid silently skipping images

cli/src/data.ts
```diff
@@ -59,25 +66,79 @@ function loadTinderDogData(split: number): Dataset<DataFormat.Raw["image"]> {
...
+          }
+
+          return {
+            idx: imgIdx,
```
tharvik: unused
```suggestion
```

discojs/src/dataset/dataset.ts
```diff
@@ -237,6 +238,60 @@ export class Dataset<T> implements AsyncIterable<T> {
...
+        const buffer: T[] = [];
+
+        while (buffer.length < windowSize){
+          const n = await iter.next();
+          if (n.done) break;
+          buffer.push(n.value);
+        }
+
+        while (buffer.length > 0){
+          const pick = Math.floor(Math.random() * windowSize);
```
tharvik: haa yes, very nice, I though of another algorithm to implement it but yours allows for more mixing between elements, well done!

discojs/src/dataset/dataset.ts
```diff
@@ -237,6 +238,60 @@ export class Dataset<T> implements AsyncIterable<T> {
...
+  }
+
+  /** filter the indices according to the splitting condition */
+  filter(
+    condition: (value: T, index: number) => boolean | Promise<boolean>
+  ): Dataset<T>{
+    return new Dataset<T>(async function* (this: Dataset<T>): AsyncGenerator<T, void, unknown>{
+      let i = 0;
+      for await(const v of this){
+        if (await condition(v, i++)){
```
tharvik: yeah, better to split assignement from evaluation, `i++` is somewhat confusing (and varies per language), better to push it at the end of the loop

cli/src/data.ts
```diff
@@ -59,25 +66,79 @@ function loadTinderDogData(split: number): Dataset<DataFormat.Raw["image"]> {
...
+            name: e.name,
+            labelIdx,
```
tharvik: as you already checked that it was within range of `CIFAR10_LABELS`, you can directly return the label

cli/src/data.ts
```diff
@@ -59,25 +66,79 @@ function loadTinderDogData(split: number): Dataset<DataFormat.Raw["image"]> {
     });
 }
 
+async function readLabels(labelsPath:string): Promise<string[]>{
+  const raw = await fs.readFile(labelsPath, "utf8");
+  return raw.split(/\r?\n/).map(s => s.trim()).filter(s => s.length > 0);
+}
+
+function loadExtCifar10(userIdx: number){
+  const CIFAR10_LABELS = ["airplane", "automobile", "bird", "cat", "deer", "dog", "frog", "horse", "ship", "truck"];
```
tharvik: this is taken from `defaultTasks.cifar10.getTask().then(t => t.trainingInformation.LABEL_LIST)`, better to use this one directly

discojs/src/privacy.ts
```diff
@@ -12,13 +14,85 @@ async function frobeniusNorm(weights: WeightsContainer): Promise<number> {
   return Math.sqrt(squared[0]);
 }
 
-/** Scramble weights */
-export function addNoise(
+/** Compute the frobenius norm of weight for one layer */
+export async function layerWiseFrobeniusNorm(weight: tf.Tensor): Promise<number>{
```
tharvik: IMO, this one can because the only `frobeniusNorm` as we can kind of define the other via this one, and the other is only used once

discojs/src/privacy.ts
```diff
@@ -12,13 +14,85 @@ async function frobeniusNorm(weights: WeightsContainer): Promise<number> {
...
+  const squared = await weight.square().sum().data();
+  return Math.sqrt(squared[0]);
+}
+
+/** ALDP-FL implementation */
+// Conditions need to be added for the first three epochs -> get the avg update from all of the available previous updates
+export function getClippingRadius(weightHistory: WeightHistory, defaultRadius:number|undefined = 20, windowSize: number=3, minRadius=1e-12): number | number[]{
```
tharvik: using [default params in js](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters) is the same as allowing undef and then setting the variable to the default. so no need to it ourself.

also, the type can be easily infered via the default, no need to set explicitly. and I see that `defaultRadius` & `windowSize` are never set in code so we can move it as constant at the start of the function (I find it a bit more readable and easy to change back if wanted).

```suggestion
export function getClippingRadius(
	weightHistory: WeightHistory,
	defaultRadius = 20,
): number | number[] {
  const WINDOW_SIZE = 3;
  const MINIMUM_RADIUS = 1e-12;
```

you can then avoid the `?? 20` below

discojs/src/privacy.ts
```diff
@@ -12,13 +14,85 @@ async function frobeniusNorm(weights: WeightsContainer): Promise<number> {
...
+      const avg = recent.reduce((sum, n) => sum+n, 0) / recent.length;
+
+      if (avg > defaultRadius)
+        return defaultRadius
+
+      return Math.max(minRadius, avg);
+    }
+  });
+
+  console.log('Print the clipping radius', radii);
```
tharvik: debug?
```suggestion
```

discojs/src/privacy.ts
```diff
@@ -12,13 +14,85 @@ async function frobeniusNorm(weights: WeightsContainer): Promise<number> {
...
+}
+
+/** ALDP-FL implementation */
+// Conditions need to be added for the first three epochs -> get the avg update from all of the available previous updates
+export function getClippingRadius(weightHistory: WeightHistory, defaultRadius:number|undefined = 20, windowSize: number=3, minRadius=1e-12): number | number[]{
+  // When there is no update in weightHistory yet
+  if (weightHistory.size === 0) return defaultRadius ?? 20;
+
+  const radii = Array.from(weightHistory.entries()).map(([_, norms]) => {
+    if (!norms || norms.length === 0) return defaultRadius ?? 20;
```
tharvik: can't be false

```suggestion
    if (norms.length === 0) return defaultRadius ?? 20;
```

discojs/src/privacy.ts
```diff
@@ -12,13 +14,85 @@ async function frobeniusNorm(weights: WeightsContainer): Promise<number> {
...
+      return Math.max(minRadius, avg);
+    } else{
+      const recent = norms.slice(-windowSize);
+      const avg = recent.reduce((sum, n) => sum+n, 0) / recent.length;
+
+      if (avg > defaultRadius)
+        return defaultRadius
+
+      return Math.max(minRadius, avg);
+    }
```
tharvik: both branches end in the same way and pretty much do the same thing, we can simplify it (and the branching logic also)
```suggestion
    const recent = norms.slice(-windowSize);
    const avg = recent.reduce((sum, n) => sum+n, 0) / recent.length;

    return Math.max(minRadius, Math.min(avg, defaultRadius));
```

discojs/src/privacy.ts
```diff
@@ -12,13 +14,85 @@ async function frobeniusNorm(weights: WeightsContainer): Promise<number> {
...
+  });
+
+  console.log('Print the clipping radius', radii);
+  return radii.length === 1 ? radii[0] : radii;
+}
+
+
+/** Add Gaussian noise per layer */
+export async function addGaussianNoise(
```
tharvik: unused?

discojs/src/privacy.ts
```diff
@@ -12,13 +14,85 @@ async function frobeniusNorm(weights: WeightsContainer): Promise<number> {
...
+  return clippedWeights.map((w) => w.add(tf.randomNormal(w.shape, 0, sigma)))
+}
+
+/** Optimized Gaussian noise using a clipping radius calculation of ALDP-FL for adaptive local differential privacy in federated learning,
+ *  https://www.nature.com/articles/s41598-025-12575-6 */
+/** Implementation of historical moving average based clipping radius calculation */
+export async function addNoise(
+  weights: WeightsContainer,
+  epsilon: number,
+  delta: number = 1e-6,
```
tharvik: optional param followed by required ones are weird to use, and as it is set everywhere, we can remove its default IMO

```suggestion
  delta: number,
```

discojs/src/privacy.ts
```diff
@@ -12,13 +14,85 @@ async function frobeniusNorm(weights: WeightsContainer): Promise<number> {
...
+
+      if (avg > defaultRadius)
+        return defaultRadius
+
+      return Math.max(minRadius, avg);
+    }
+  });
+
+  console.log('Print the clipping radius', radii);
+  return radii.length === 1 ? radii[0] : radii;
```
tharvik: is it somewhat akward to work with OR types and in here, we can avoid that by simply returning the array.
that will also simplify the implementation of `addNoise` & `clipNorm` as theses would only need to work with `number[]` instead of `number | number[]`

discojs/src/privacy.ts
```diff
@@ -12,13 +14,85 @@ async function frobeniusNorm(weights: WeightsContainer): Promise<number> {
...
+/** Optimized Gaussian noise using a clipping radius calculation of ALDP-FL for adaptive local differential privacy in federated learning,
+ *  https://www.nature.com/articles/s41598-025-12575-6 */
+/** Implementation of historical moving average based clipping radius calculation */
+export async function addNoise(
+  weights: WeightsContainer,
+  epsilon: number,
+  delta: number = 1e-6,
+  clippingRadius: number[] | number,  
+): Promise<WeightsContainer> {
+  if (typeof clippingRadius === "number"){
```
tharvik: this path is the same as the `else` with an array of a single element, you can merge both

discojs/src/privacy.ts
```diff
@@ -28,12 +102,29 @@ export function addNoise(
  **/
 export async function clipNorm(
```
tharvik, line 58: can also be simplified with similar comments as before: path with `number` is kind of the same as having a single element array

discojs/src/privacy.ts
```diff
@@ -28,12 +102,29 @@ export function addNoise(
...
+
+    return weights.map((w) => w.div(scaling));
+  }else{
+    /** Apply different clipping radius to each layer in the WeightsContainer */
+    const layers = weights.weights;
+
+    const clipped = await Promise.all(
+      layers.map(async (l, i) => {
+        const norm = await layerWiseFrobeniusNorm(l);
+        const scaling = Math.max(1, norm / radius[i]);
```
tharvik: we don't actually check for `radius[i]` is valid as done for the `number` path, better to throw than generating invalid value (`x / 0`  w/ `x > 0` gives `Infinity`)

discojs/src/privacy.ts
```diff
@@ -12,13 +14,85 @@ async function frobeniusNorm(weights: WeightsContainer): Promise<number> {
...
+  return Math.sqrt(squared[0]);
+}
+
+/** ALDP-FL implementation */
+// Conditions need to be added for the first three epochs -> get the avg update from all of the available previous updates
+export function getClippingRadius(weightHistory: WeightHistory, defaultRadius:number|undefined = 20, windowSize: number=3, minRadius=1e-12): number | number[]{
+  // When there is no update in weightHistory yet
+  if (weightHistory.size === 0) return defaultRadius ?? 20;
+
+  const radii = Array.from(weightHistory.entries()).map(([_, norms]) => {
```
tharvik: we are only interrested in the values, no need for the full entries

```suggestion
  const radii = Array.from(weightHistory.values()).map((norms) => {
```

discojs/src/training/trainer.ts
```diff
@@ -22,6 +22,20 @@ export interface RoundLogs {
   participants: number;
 }
 
+/** Key: layer index, Value: array of weight update norms */
+export type WeightHistory = Map<number, number[]>;
```
tharvik: I'm striving to use [immutable's datastructures](https://immutable-js.com/) to avoid unintended changes of data-only structures. that does require a bit of changes as now everytime one add an element to a datastructure, it will create a new collection rather than updating to original one.

while reading a bit more on it, it seems that we don't really need the keys, it is akin to a list position so I guess we can even drop the `Map` and use an array/`List` directly.

also, it isn't really a "history of weights" but a "history of norms"

```suggestion
export type WeightHistory = List<List<number>>;
```

discojs/src/training/trainer.ts
```diff
@@ -38,6 +52,9 @@ export class Trainer<D extends DataType, N extends Network> {
     AsyncGenerator<AsyncGenerator<BatchLogs, EpochLogs>, RoundLogs>,
     void
   >;
+  // Map of weight Index and weight update
+  weightHistory : WeightHistory = new Map<number, number[]>();
+  previousEpochWeights: WeightsContainer | undefined = undefined;
```
tharvik: class variables defined this way make theses public, better to use the [somewhat-new-but-very-useful private elements](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_elements) to keep theses private to the class.

also, you don't need to set `undefined`, it has this value by default

```suggestion
  #weightHistory : WeightHistory = new Map<number, number[]>();
  #previousEpochWeights: WeightsContainer | undefined;
```

discojs/src/training/trainer.ts
```diff
@@ -94,24 +111,25 @@ export class Trainer<D extends DataType, N extends Network> {
...
       yield this.#runRound(dataset, validationDataset);
 
-      let localWeights = this.model.weights;
-      if (this.#privacy !== undefined)
-        localWeights = await applyPrivacy(
-          previousRoundWeights,
-          localWeights,
-          this.#privacy,
-        );
+      const localWeights = this.model.weights;
```
tharvik: assigned once and used on the next line, you can inline it

discojs/src/training/trainer.ts
```diff
@@ -94,24 +111,25 @@ export class Trainer<D extends DataType, N extends Network> {
...
 
+      // Get the updated weights
       const networkWeights =
-        await this.#client.onRoundEndCommunication(localWeights);
+      await this.#client.onRoundEndCommunication(localWeights);
```
tharvik: confusing to have it at the same identation, it looks like a statement rather than an assignement
```suggestion
        await this.#client.onRoundEndCommunication(localWeights);
```

discojs/src/training/trainer.ts
```diff
@@ -94,24 +111,25 @@ export class Trainer<D extends DataType, N extends Network> {
...
+      
+      // Update the local weights
+      this.model.weights = networkWeights;
 
-      this.model.weights = previousRoundWeights = networkWeights;
+      // Update the previous weights after aggregation
+      this.previousEpochWeights = new WeightsContainer(networkWeights.weights);
```
tharvik: you can directly set it as `WeightsContainer` are immutable

```suggestion
      this.previousEpochWeights = networkWeights;
```

discojs/src/training/trainer.ts
```diff
@@ -125,8 +143,40 @@ export class Trainer<D extends DataType, N extends Network> {
...
+        if (this.previousEpochWeights){
+          weightsProgress = epochWeights.sub(this.previousEpochWeights);
+          updateNorm = await Promise.all(weightsProgress.weights.map((w)=>(privacy.layerWiseFrobeniusNorm(w))));
+          appendWeightHistory(this.weightHistory, updateNorm);
+        }
+        else{
+          /** For the first epoch */
+          updateNorm = await Promise.all(epochWeights.weights.map((w)=>(privacy.layerWiseFrobeniusNorm(w))));
+          appendWeightHistory(this.weightHistory, updateNorm);
+        }
```
tharvik: kind of the same ending in both paths, the "first epoch" one can be expressed by having a `previousEpochWeights` with only zeros

discojs/src/training/trainer.ts
```diff
@@ -125,8 +143,40 @@ export class Trainer<D extends DataType, N extends Network> {
...
+          epochWeights,
+          this.#privacy,
+          this.weightHistory,
+          this.#epochs,
+        );
+
+        this.model.weights = epochWeights;
+      }
+      
+      this.previousEpochWeights = new WeightsContainer(epochWeights.weights);
```
tharvik:
```suggestion
      this.previousEpochWeights = epochWeights;
```

discojs/src/training/trainer.ts
```diff
@@ -158,8 +206,27 @@ async function applyPrivacy(
     );
   }
 
-  if (options.noiseScale !== undefined)
-    ret = privacy.addNoise(ret, options.noiseScale);
+  // Adding Gaussian noise for DP
+  if (options.noiseScale !== undefined && options.delta !== undefined){
```
tharvik: theses have to be both defined for DP to work, right? we can enforce that by having `Task.trainingInformation.privacy` with an object like "differentialPrivacy" with requires both to be set. so smth like 

```ts
// reduce training accuracy and improve privacy.
privacy: z
	.object({
		// maximum weights difference between each round
		clippingRadius: z.number().optional(),

		differentialPrivacy: z.object({
			// maximum weights difference between each epoch, used for differential privacy
			dpDefaultClippingRadius: z.number().positive().optional(),
			// privacy budget, used to compute the variance of Gaussian noise
			noiseScale: z.number().positive(),
			// small probability that the privacy guarantee may not hold
			delta: z.number().positive(),
		}).optional(),
	})
```

discojs/src/training/trainer.ts
```diff
@@ -158,8 +206,27 @@ async function applyPrivacy(
...
+    const dpDefaultRadius = options.dpDefaultClippingRadius; // options.dpDefaultClippingRadius should be a number
```
tharvik: there are two type of radius, one for BFT and one for DP, consider adding a specific object for `BFT` in the privacy's schema and rename `dpDefaultClippingRadius` to `clippingRadius`

you can also set defaults directly in the schema with the `default` method

discojs/src/training/trainer.ts
```diff
@@ -158,8 +206,27 @@ async function applyPrivacy(
...
+    /** Need to use tighter clipping radius for noise calibration */
+    if (options.clippingRadius !== undefined){
+      /** When clipping radius for BFT is smaller than the improved clipping radius */
+      if (Array.isArray(dpClippingRadius)&&options.clippingRadius < Math.min(...dpClippingRadius))
+        ret = current.add(await privacy.addNoise(weightsProgress, epsilon, delta, options.clippingRadius));
+      else if(typeof dpClippingRadius == "number" && options.clippingRadius < dpClippingRadius)
+        ret = current.add(await privacy.addNoise(weightsProgress, epsilon, delta, options.clippingRadius));
+    }else{
+      ret = current.add(await privacy.addNoise(weightsProgress, epsilon, delta, dpClippingRadius));
+    }
```
tharvik: theses three code branches are quite similar, surely there is a way to commonize logic

webapp/src/components/task_creation_form/TaskCreationForm.vue
```diff
@@ -364,12 +364,38 @@
...
+                >
+                  <FormField
+                    name="trainingInformation.privacy.delta"
+                    placeholder="1e-5"
+                    as="input"
+                    type="number"
+                  />
+                </FormLabel>
+              </div>
+              <div v-show="differentialPrivacy" class="flex flex-col">
```
tharvik: you can keep the same div block for the whole option, it's simpler and ~links the elements tgth

```suggestion

                <FormLabel
                  label="Error parameter (δ): Small probability that the privacy guarantee may not hold"
                  type="required"
                >
                  <FormField
                    name="trainingInformation.privacy.delta"
                    placeholder="1e-5"
                    as="input"
                    type="number"
                  />
                </FormLabel>

```

webapp/src/components/task_creation_form/TaskCreationForm.vue
```diff
@@ -364,12 +364,38 @@
...
+                <FormLabel
+                  label="Default clipping norm: Initial limit on weight update size for differential privacy"
+                  type="required"
```
tharvik: here is it required but not in the schema, is it wanted?

webapp/src/components/task_creation_form/TaskCreationForm.vue
```diff
@@ -364,12 +364,38 @@
...
+                  />
+                </FormLabel>
+              </div>
+              <div v-show="differentialPrivacy" class="flex flex-col">
+                <FormLabel
+                  label="Error parameter (δ): Small probability that the privacy guarantee may not hold"
+                  type="required"
+                >
+                  <FormField
+                    name="trainingInformation.privacy.delta"
```
tharvik: we have to readd (yeah, that's bothering but I wasn't able to find a way to map over zod schema to transform theses) in the `nonLocalNetwork` schema below. that let the form manager know what is a valid value and warn the user about it. it has to have the same shape as the base schema otherwise it will get rejected by the server.

tharvik (changes requested):
thanks for the great work! I added a few comments on how to improve parts of it but the core feature is there :tada:

btw, I don't remember if we talked about adding a test for the feature. for an integration test, you can add one in `server/tests/e2e/federated.spec.ts` trying out the feature and if you have some hardening/units tests in mind, you can also some in `discojs/src/privacy.spec.ts` to ensure that it works in many conditions.

tharvik (approved):
updated it a bit with the latest changes for byzantine impl, thanks for the huge contributions!

PR #1052, opened 2026-02-18
Add new percentile aggregator and tests for comparing it with Byzantine

mina5rovic:
Implementation of new percentile aggregator based on the previously implemented Byzantine (in previous version of disco). This one work faster and in some cases better than the Byzantine. Also implemented tests for comparing those two. 

Changing Byzantine aggregator based on errors found in implementation while comparing them.


discojs/src/aggregator/percentile_clipping.spec.ts
```diff
@@ -0,0 +1,423 @@
...
+
+    // --- TEST 2: Old Aggregator (Percentile Clipping) ---
+    const oldAgg = new PercentileClippingAggregator(0, 3, "absolute", 0.1);
+    oldAgg.setNodes(Set(allPeers));
+
+    const timingOld = await measureAggregation(oldAgg, "Old PercentileClipping", peersWithValues);
+
+    // Both should produce a value closer to 1.0 than 100.0
+    expect(timingNew.result).toBeLessThan(50);
+    expect(timingOld.result).toBeLessThan(50);
```
JulienVig: Both aggregators result in exactly 1, is this expected? Shouldn't it be at least slightly influenced by the Byzantine outlier?

```
=== Timing Comparison: Simple Outlier Rejection ===
  New ByzantineRobust    | 14.655ms | result: 1.0000
  Old PercentileClipping | 1.197ms | result: 1.0000
```

discojs/src/aggregator/percentile_clipping.spec.ts
```diff
@@ -0,0 +1,423 @@
...
+
+    // Print timing comparison
+    console.log("\n=== Timing Comparison: Simple Outlier Rejection ===");
+    console.log(formatTiming([timingNew, timingOld]));
+  });
+
+  it("old aggregator with different percentiles", async () => {
```
JulienVig: So for this test I'm getting these results:
```
=== Timing Comparison: Old Aggregator with Different Percentiles ===
  tau=0.05 | 2.350ms | result: 1.0000
  tau=0.1  | 1.645ms | result: 1.0000
  tau=0.2  | 2.607ms | result: 1.0000
  tau=0.5  | 1.826ms | result: 1.0000

=== Timing Comparison: New Aggregator with Different Clipping Radii ===
  radius=0.5 | 2.293ms | result: 0.5000
  radius=1   | 0.949ms | result: 1.0000
  radius=2   | 0.922ms | result: 1.4000
  radius=5   | 0.749ms | result: 2.6000
```

If I understand correctly, the old aggregator is more robust than the new (but slower) in this case? Why aren't the results changing with different tau for the old aggregator?

discojs/src/aggregator/percentile_clipping.spec.ts
```diff
@@ -0,0 +1,423 @@
...
+    ];
+
+    const timingRound2 = await measureAggregation(agg, "Round 2", peersRound2);
+
+    // With centering on previous (5.0), updates to 10.0 should result in something close to 10.0
+    expect(timingRound2.result).toBeGreaterThan(5.0);
+
+    console.log("\n=== Timing Comparison: State Preservation Across Rounds ===");
+    console.log(formatTiming([timingRound1, timingRound2]));
+  });
```
JulienVig: I'm getting these:
```
=== Timing Comparison: State Preservation Across Rounds ===
  Round 1 | 0.290ms | result: 5.0000
  Round 2 | 0.305ms | result: 10.0000
```

Shouldn't it be a bit lower than 10 because of the previous round? Seems that round 1's state has been entirely discarded

discojs/src/aggregator/percentile_clipping.spec.ts
```diff
@@ -0,0 +1,423 @@
...
+
+    for (const iter of iterations) {
+      const agg = new ByzantineRobustAggregator(0, 5, "absolute", 1.0, iter, 0);
+      agg.setNodes(Set(allPeers));
+
+      const timing = await measureAggregation(agg, `iterations=${iter}`, peersWithValues);
+      timings.push(timing);
+    }
+
+    console.log("\n=== Performance Impact of Iterative Refinement ===");
```
JulienVig:
```
=== Performance Impact of Iterative Refinement ===
  iterations=1  | 0.762ms | result: 1.0000
  iterations=2  | 1.233ms | result: 1.4000
  iterations=5  | 5.417ms | result: 1.6496
  iterations=10 | 7.004ms | result: 1.6665
```
Intuitively I expected that more iterations would improve the result while it seems that it strays further from the optimal result, is this behavior correct?

discojs/src/aggregator/percentile_clipping.spec.ts
```diff
@@ -0,0 +1,423 @@
...
+    const timingNew = await measureAggregation(newAgg, "New (40% Byzantine)", peersWithValues);
+
+    const oldAgg = new PercentileClippingAggregator(0, allPeers.length, "absolute", 0.1);
+    oldAgg.setNodes(Set(allPeers));
+    const timingOld = await measureAggregation(oldAgg, "Old (40% Byzantine)", peersWithValues);
+
+    console.log("\n=== Byzantine Robustness: High Ratio Attack (4/10 = 40% malicious) ===");
+    console.log(formatTiming([timingNew, timingOld]));
+    console.log(`  Result gap: new=${timingNew.result.toFixed(4)}, old=${timingOld.result.toFixed(4)}`);
+    console.log(`  Winner: ${timingNew.result < timingOld.result ? "NEW (closer to honest 1.0)" : "OLD (closer to honest 1.0)"}`);
```
JulienVig: It's missing the case where both results are equal and it's a tie

discojs/src/aggregator/percentile_clipping.spec.ts
```diff
@@ -0,0 +1,423 @@
...
+
+    const oldAgg = new PercentileClippingAggregator(0, allPeers.length, "absolute", 0.2);
+    oldAgg.setNodes(Set(allPeers));
+    const timingOld = await measureAggregation(oldAgg, "Old (tau=0.2)", peersWithValues);
+
+    console.log("\n=== Gradient Poisoning Attack (coordinated Byzantine values) ===");
+    console.log(formatTiming([timingNew, timingOld]));
+    console.log(`  Result gap: new=${timingNew.result.toFixed(4)}, old=${timingOld.result.toFixed(4)}`);
+    console.log(`  Expected honest value: 1.0000`);
+    console.log(`  Winner: ${Math.abs(timingNew.result - 1.0) < Math.abs(timingOld.result - 1.0) ? "NEW (closer to honest)" : "OLD (closer to honest)"}`);
```
JulienVig: Address tie here too

discojs/src/aggregator/percentile_clipping.spec.ts
```diff
@@ -0,0 +1,423 @@
...
+    const timingNew = await measureAggregation(newAgg, "New (5 iterations)", peersWithValues);
+
+    const oldAgg = new PercentileClippingAggregator(0, allPeers.length, "absolute", 0.2);
+    oldAgg.setNodes(Set(allPeers));
+    const timingOld = await measureAggregation(oldAgg, "Old (tau=0.2)", peersWithValues);
+
+    console.log("\n=== Gradient Poisoning Attack (coordinated Byzantine values) ===");
+    console.log(formatTiming([timingNew, timingOld]));
+    console.log(`  Result gap: new=${timingNew.result.toFixed(4)}, old=${timingOld.result.toFixed(4)}`);
+    console.log(`  Expected honest value: 1.0000`);
```
JulienVig:
```
=== Gradient Poisoning Attack (coordinated Byzantine values) ===
  New (5 iterations) | 8.113ms | result: 2.3094
  Old (tau=0.2)      | 0.473ms | result: 1.0000
  Result gap: new=2.3094, old=1.0000
  Expected honest value: 1.0000
  Winner: OLD (closer to honest)
```
So the old aggregator is both faster and more robust? Doesn't this go against the theory for the robustness?

discojs/src/aggregator/percentile_clipping.spec.ts
```diff
@@ -0,0 +1,423 @@
...
+    const timing2Old = await measureAggregation(oldAgg, "Old Round 2 (attack)", round2Values);
+
+    console.log("\n=== Adaptive Multi-Round Attack ===");
+    console.log("Round 1 (cooperation):");
+    console.log(formatTiming([timing1New, timing1Old]));
+    console.log("\nRound 2 (adaptive Byzantine attack):");
+    console.log(formatTiming([timing2New, timing2Old]));
+    console.log(`  New result: ${timing2New.result.toFixed(4)} (expected ~10.0)`);
+    console.log(`  Old result: ${timing2Old.result.toFixed(4)} (expected ~10.0)`);
+    console.log(`  Winner: ${Math.abs(timing2New.result - 10.0) < Math.abs(timing2Old.result - 10.0) ? "NEW (better rejects attack)" : "OLD (better rejects attack)"}`);
```
JulienVig: Address potential tie

discojs/src/aggregator/percentile_clipping.spec.ts
```diff
@@ -0,0 +1,423 @@
...
+
+    // Round 2: Byzantine launches adaptive attack
+    const round2Values = [
+      ...honestPeers.map(id => ({ id, value: 10.0 })), // Honest update
+      ...byzantinePeers.map(id => ({ id, value: 50.0 })), // Byzantine aggressive attack in round 2
+    ];
+
+    const timing2New = await measureAggregation(newAgg, "New Round 2 (attack)", round2Values);
+    const timing2Old = await measureAggregation(oldAgg, "Old Round 2 (attack)", round2Values);
+
```
JulienVig:
```
=== Adaptive Multi-Round Attack ===
Round 1 (cooperation):
  New Round 1 | 4.476ms | result: 5.0000
  Old Round 1 | 2.309ms | result: 5.0000

Round 2 (adaptive Byzantine attack):
  New Round 2 (attack) | 32.874ms | result: 10.0000
  Old Round 2 (attack) | 17.731ms | result: 10.0000
  New result: 10.0000 (expected ~10.0)
  Old result: 10.0000 (expected ~10.0)
  Winner: OLD (better rejects attack)
```

Smells fishy that the results are exactly the expected value no?

discojs/src/aggregator/percentile_clipping.spec.ts
```diff
@@ -0,0 +1,423 @@
...
+    console.log(`\nOld aggregator (${timeOld.toFixed(2)}ms):`);
+    console.log(`  Layer 1: ${arrOld[0][0].toFixed(2)} (expected ~100)`);
+    console.log(`  Layer 2: ${arrOld[0][1].toFixed(2)} (expected ~10)`);
+    console.log(`  Layer 3: ${arrOld[0][2].toFixed(2)} (expected ~1)`);
+
+    // Check relative error
+    const newError = Math.abs((arrNew[0][0] - 100) / 100) + Math.abs((arrNew[0][1] - 10) / 10) + Math.abs(arrNew[0][2] - 1);
+    const oldError = Math.abs((arrOld[0][0] - 100) / 100) + Math.abs((arrOld[0][1] - 10) / 10) + Math.abs(arrOld[0][2] - 1);
+    console.log(`\nTotal relative error: new=${newError.toFixed(3)}, old=${oldError.toFixed(3)}`);
+    console.log(`Winner: ${newError < oldError ? "NEW (better handles multi-scale)" : "OLD (better handles multi-scale)"}`);
```
JulienVig: Adress tie

discojs/src/aggregator/percentile_clipping.spec.ts
```diff
@@ -0,0 +1,423 @@
...
+    console.log("Honest peers send: [100, 10, 1]");
+    console.log("Byzantine sends: [10000, 1000, 100]");
+    console.log(`\nNew aggregator (${timeNew.toFixed(2)}ms):`);
+    console.log(`  Layer 1: ${arrNew[0][0].toFixed(2)} (expected ~100)`);
+    console.log(`  Layer 2: ${arrNew[0][1].toFixed(2)} (expected ~10)`);
+    console.log(`  Layer 3: ${arrNew[0][2].toFixed(2)} (expected ~1)`);
+    console.log(`\nOld aggregator (${timeOld.toFixed(2)}ms):`);
+    console.log(`  Layer 1: ${arrOld[0][0].toFixed(2)} (expected ~100)`);
+    console.log(`  Layer 2: ${arrOld[0][1].toFixed(2)} (expected ~10)`);
+    console.log(`  Layer 3: ${arrOld[0][2].toFixed(2)} (expected ~1)`);
```
JulienVig:
```
=== Heterogeneous Gradients: Multi-Tensor Federated Model ===
Gradient structure: [layer1=100×value, layer2=10×value, layer3=value]
Honest peers send: [100, 10, 1]
Byzantine sends: [10000, 1000, 100]

New aggregator (1.98ms):
  Layer 1: 14.92 (expected ~100)
  Layer 2: 1.49 (expected ~10)
  Layer 3: 0.15 (expected ~1)

Old aggregator (0.45ms):
  Layer 1: 100.00 (expected ~100)
  Layer 2: 10.00 (expected ~10)
  Layer 3: 1.00 (expected ~1)

Total relative error: new=2.552, old=0.000
Winner: OLD (better handles multi-scale)
```

The new aggregator is quite far off the expected values, potentially there is an implementation issue to fix there

JulienVig (changes requested):
The comparison is very useful, thanks!

I have a bunch of questions, mostly due to my superficial knowledge of the algorithms. Overall, we're often getting exact values like 1.0000 which seems a bit fishy. Also, the new aggregator seems worse than the old both time- and robustness-wise. I believe it should at least be more robust so there are maybe some bugs that slipped into the implementation?

Finally, could you add a memory test case to ensure that the implementation is currently doing garbage disposal of unused TF tensors? You can do so by running multiple rounds/aggregations and keeping track of the number of tensors allocated in memory ([tf.memory()](https://js.tensorflow.org/api/latest/#memory)), which ideally should stay constant throughout the training. You can read more on memory management [here](https://www.tensorflow.org/js/guide/tensors_operations#memory).

discojs/src/aggregator/percentile_clipping.ts
```diff
@@ -0,0 +1,140 @@
...
+}
+
+function frobeniusNorm(w: WeightsContainer): number {
+  // Computes the Frobenius (L2) norm of all tensors in a WeightsContainer
+  return tf.tidy(() => {
+    const total = w.weights
+      .map(t => tf.sum(tf.square(t)))
+      .reduce((a, b) => tf.add(a, b), tf.scalar(0));
+
+    return tf.sqrt(total).dataSync()[0];
```
JulienVig, line 138: Ideally we'd only use async computation to avoid blocking other operations. I guess here that'd required refactoring how the aggregate method into an async method is that right?

discojs/src/aggregator/percentile_clipping.ts
```diff
@@ -0,0 +1,140 @@
...
+    // Step 3: Compute Frobenius norms for each centered weight
+    const normArray = centeredWeights.map(w => frobeniusNorm(w));
+
+    // Step 4: Compute tau as the percentile of the norm array
+    const tau = this.computePercentile(normArray, this.tauPercentile);
+
+    // Step 5: Clip weights based on tau
+    // Each peer gets one scale factor based on their Frobenius norm
+    const clippedWeights = centeredWeights.map((w, peerIdx) => {
+      //const scaleFactor = Math.min(1, tau / normArray[peerIdx]);
```
JulienVig, line 86:
```suggestion
```
Commented code

discojs/src/aggregator/byzantine.ts
```diff
@@ -77,34 +102,55 @@ export class ByzantineRobustAggregator extends MultiRoundAggregator {
...
+
+      const oldV = v;
       v = newV;
+      oldV.dispose();
     }
-    // Step 3: Update momentum history
+
+    eps.dispose();
+    one.dispose();
+    radius.dispose();
```
JulienVig, line 152:
```suggestion
    tf.dispose([eps, one, radius]);
```
fyi you can also dispose of multiple variables in one line

JulienVig (approved):
Thanks Mina, all good for me!

PR #1053, opened 2026-02-22
Fix issue #885

### LINKED ISSUE #885: Allow sharing DISCOllborative links
Currently, trying to directly load https://discolab.ai/#/lus_covid or any of the DISCOllaboratives redirects to https://discolab.ai/#/list.
We would like to allow sharing DISCOllaborative links and remove this redirection.

This redirection has been introduced as a temporary fix to 404 errors occurring when trying to load tasks directly. The redirection happens in webapp/Training (now TrainingSteps) and has been merged in #727.

What makes the issue non-trivial is the fact that tasks are provided asynchronously from the server to the webapp, so loading the DISCOllaborative URL directly first requires querying the server for the task and waiting for the answer. Currently, the Google server instance can take up to 10s to boot up so there needs to be a loading indicator implemented to provide user feedback.

Optionally, look into whether we can get rid of the hashtag currently introduced by the [Vue router history mode](https://router.vuejs.org/guide/essentials/history-mode#Hash-Mode).

mina5rovic:
Fixed https://github.com/epfml/disco/issues/885. 

Removed # from URL content.

No more redirecting when loading tasks.

### DIFF
--- a/webapp/cypress/e2e/datasetInput.cy.ts
+++ b/webapp/cypress/e2e/datasetInput.cy.ts
@@ -4,7 +4,7 @@ import { basicTask, setupServerWith } from "../support/e2e";
 // upstream doesn't yet allow that vuejs/test-utils#2468
 
 function goToDatasetInputStep() {
-  cy.visit("/#/list");
+  cy.visit("/list");
   cy.get(".driver-popover-close-btn").click();
   cy.get("button").contains("participate").click();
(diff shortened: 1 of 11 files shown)


JulienVig (approved):
Thanks for the fix! It's working great.

Could you hide the Previous/Next button when the task is loading? So that users can't move around in the steps before the task loads.
Image

PR #1079, opened 2026-03-18
*: bump deps

tharvik:
* bump most deps
* wrtc: also drop node-datachannel in favor of @roamhq/wrtc as the latter work better w/ simple-peer
* do not bump jsdom to latest as its types aren't yet available
* do not bump vee-validate to latest beta: beta.1's `<Field />` is not converting number input to numbers anymore, maybe in the next iteration it will work again
* examples: drop eslint as it drifted too much from the one we use, and that it's not really our role to provide a linter in an example IMO


JulienVig (approved):
Thanks Valérian!!

PR #1080, opened 2026-03-20
Draft: Add automated test runner and visualization script for stress testing

ahzero7d1:
- Add extra arguments for running full test configurations
- Separate test configuration and execution script
- Add visualization script


cli/src/python/visualize_logs.py
```diff
@@ -0,0 +1,133 @@
+import argparse
+import json
+from pathlib import Path
+from typing import Iterable
+
+import pandas as pd
+import matplotlib.pyplot as plt
```
JulienVig, line 6: Could you add a `requirements.txt` with the libraries contributors should install?

cli/src/python/visualize_logs.py
```diff
@@ -0,0 +1,133 @@
...
+    plt.legend()
+    plt.tight_layout()
+    plt.savefig(output_path)
+    plt.close()
+
+# Plot mean value of a metric with std
+def plot_mean_std(df, metric, output_path, title, ylabel):
+    summary = df.groupby("step")[metric].agg(["mean", "std"]).reset_index() 
+
+    summary["std"] = summary["std"].fillna(0)
```
JulienVig, line 68: Why were there NaN stds?

cli/src/scripts/run_experiments.py
```diff
@@ -0,0 +1,60 @@
...
+    if "beta" in cfg:
+        cmd += ["--beta", str(cfg["beta"])]
+    if "maxShareValue" in cfg:
+        cmd += ["--maxShareValue", str(cfg["maxShareValue"])]
+    
+    return cmd
+
+def main():
+    # path of configuration file for experiments
+    config_path = Path(sys.argv[1])
```
JulienVig: Could you set the default value to the path to basic_tests.json?

cli/src/python/experiments/basic_tests.json
```diff
@@ -0,0 +1,225 @@
...
+    "epochs": 50,
+    "roundDuration": 2,
+    "validationSplit": 0.2,
+    "save": true,
+    "host": "http://localhost:8080"
+  },
+  "experiments": [
+    {
+      "testID": "mnist_fed_mean_cnn3_p3_d600_e50_r2",
+      "task": "mnist_federated",
```
JulienVig: I'm getting this error:
```
cli/dist/args.js:46
    throw Error(`${unsafeArgs.task} not implemented.`);
          ^
Error: mnist_federated not implemented
```
Did you create new default tasks (`mnist_federated`, `titanic_decentralized`, etc.)?

cli/src/cli.ts
```diff
@@ -17,49 +19,103 @@ import { Disco, aggregator as aggregators, client as clients } from '@epfml/disc
...
+  const streamPath = path.join(dir, `client${userIndex}_local_log.ndjson`);
+
+  const finalLog: SummaryLogs[] = [];
+  // create a write stream that saves learning logs during the train
+  let ndjsonStream: ReturnType<typeof createWriteStream> | null = null;
+
+  if (args.save){
+    ndjsonStream = createWriteStream(streamPath, {flags: "w"});
+  }
+
```
JulienVig: Did you choose the "ndjson" name over jsonl for a particular reason? I had only ever seen jsonl until now so unless you particularly prefer ndjson I would change it to jsonl

discojs/src/models/model.ts
```diff
@@ -39,6 +39,17 @@ export abstract class Model<D extends DataType> implements Disposable {
...
+  /**
+   * Return validation metrics
+   * 
+   * TODO: currently only works for TFJS, gpt models
+   */
+  evaluate(
+    _validationDataset?: Dataset<Batched<DataFormat.ModelEncoded[D]>>
+  ): Promise<ValidationMetrics>{
+    throw new Error("Evaluation not supported for this model");
+  }
```
JulienVig:
```suggestion
  /**
   * Return validation metrics
   */
  abstract evaluate(
    _validationDataset?: Dataset<Batched<DataFormat.ModelEncoded[D]>>
  ): Promise<ValidationMetrics>;
```
I think we can make this method abstract rather than throwing

JulienVig (changes requested):
Thanks Ahyoung, the scripts are very handy! I've left a few comments and questions.

* Can you rename `scripts` to something like `python` to clarify the difference with the other typescript scripts also in `cli/src`?
* Can you also add a short `README.md` in the `scripts` (`python`) folder with instructions and example to use the two scripts?

cli/src/python/visualize_logs.py
ahzero7d1: NaN std can occur when running the function with only one client. While this is not a typical case, I think it is nice to keep this line to handle edge cases.

JulienVig: Sounds good. Could you add a small comment for this in the code?

cli/src/python/experiments/basic_tests.json
ahzero7d1: Yes, I didn’t push them since they add multiple extra default tasks. 
For now, we should define a new default task based on the `scheme` and `minNbOfParticipants`, and we can adjust the rest of learning parameters in the experiment setting json file.

cli/src/cli.ts
ahzero7d1: There was no special reason for that. I will replace ndjson to jsonl

discojs/src/models/model.ts
ahzero7d1: I didn’t make the method abstract since `ONNXModel` class does not implement `evaluate`. Do you think it would be better to make it abstract and throw an error in that class for now?

JulienVig: Hmm yes I think making it abstract and throwing for `ONNXModel` would make more sense, we're already throwing for `train`

cli/src/python/experiments/basic_tests.json
JulienVig: Okay then in this case can you modify `basic_tests.json` to only include the pre-existing default tasks? So that we can use `run_experiments.py` without having to create new tasks first. Then contributors can extend the config file as they want.

JulienVig (approved):
Thanks Ahyoung! I just have a small comment for `basic_tests.json` but other than that all good for me.

PR #1087, opened 2026-03-26
Fix secure aggregator for decentralized learning

ahzero7d1:
Fix an issue where the secure aggregator was incorrectly defined under federated task training information instead of decentralized tasks.


JulienVig (approved):
Thanks!

PR #1104, opened 2026-04-11
Security: Potential path traversal when saving models to disk

tuanaiseo:
## Problem

`saveModelToDisk` builds the output path via string interpolation (`${modelFolder}/${modelFileName}`) without sanitizing `modelFileName`. If user-controlled, attackers can inject `../` sequences or absolute paths to overwrite unintended files.

**Severity**: `high`
**File**: `discojs-node/src/model_loader.ts`

## Solution

Use `path.resolve`/`path.join` and validate that the resolved path stays inside an allowed base directory. Reject path separators in `modelFileName` or enforce a safe filename regex.

## Changes

- `discojs-node/src/model_loader.ts` (modified)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed
- [ ] No new warnings/errors introduced

### DIFF
--- a/discojs-node/src/model_loader.ts
+++ b/discojs-node/src/model_loader.ts
@@ -1,4 +1,5 @@
 import fs from "node:fs/promises";
+import path from "node:path";
 
 import type { models, DataType } from "@epfml/discojs";
 import { serialization } from "@epfml/discojs";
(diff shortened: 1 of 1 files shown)


hleepfl:
random llm change

PR #1110, opened 2026-04-15
Fix decentralized clients connection issue

### LINKED ISSUE #1109: Decentralized learning stalls during training
https://github.com/epfml/disco/blob/c2c21118abc85730092e7814a943a50d777ab87f/discojs/src/client/decentralized/decentralized_client.ts#L153-L156

Decentralized learning sometimes stalls after the first few rounds.
This appears to be caused by differences in how quickly nodes establish connections with their peers in each round.

Currently, once a node finishes establishing connections via `establishPeerConnections()`, it immediately proceeds to `exchangeWeightUpdates()` and starts sending weight updates to other peers. However, if some peers are slower in completing `establishPeerConnections()`, they may not yet be ready to receive these updates.

As a result, slower nodes can miss incoming weight updates and eventually timeout while waiting for them, causing the entire training process to stall.

ahzero7d1:
Closes: #1109 

This pull request tackles decentralized learning peer connection stability and synchronization issues. We are mainly solving 3 main issues.

### Connection Ready Check and Signaling Weight Sharing
`ConnectionsRetry`, `StartWeightSharing` messages are newly added to prevent faster peers proceed to weight sharing while slower peers are still establishing connections. Detailed process is as follows:
1. After completing peer connections, each client sends a `ConnectionsReady` message to the server.
2. The server counts the number of received `ConnectionsReady` messages.
3. Once the number of `ConnectionsReady` messages matches the number of expected participants for the round, the server sends a `StartWeightSharing` message to all round participants.
4. Clients only start sharing model updates after receiving `StartWeightSharing`.

### Connection Retries and Failed Client Disconnection
`RetryPeerConnections`, `ConnectionFail` messages are newly added. In addition, webapp shows error message when a client is disconnected after repetitive failure. `maxConnectionRetry` parameter is added to `trainingInformation`, and added to the webapp so that users can specify the value during task creation.
Connection  failure handling process is as follows:
1. If the number of retries is still below `maxConnectionRetry`, the server sends a `RetryPeerConnection` message to all peers in the current round.
2. When clients receive `RetryPeerConnection`, they clean up their peer pool and aggregator nodes, then rerun the peer connection phase.
3. If the connection setup still fails after more than `maxConnectionRetry` attempts, the server removes the failed peers from the round peer list.
4. The server sends a `ConnectionFail` message to the failed peers.
5. When a client receives `ConnectionFail`, it disconnects from the server.
6. The remaining clients receive `RetryPeerConnection` and retry the peer connection phase without the failed peers.

### Model Syncing for Participants Joining in the Middle of Training
Model syncing between peers is implemented for participants who are joining in the middle of training. `ModelSyncRequest`, `SignalModelProvider`, `SignalNewPeer`, `SharedModel` messages are newly added. In addition, `ModelWeightAccess` interface was added in `disco.ts` so that client object can adjust model weights when it receives the latest model from the provider peer.
Model syncing process is as follows:
1. When a new participant joins in the middle of training, the server marks the participant as having joined mid-training in `NewDecentralizedNodeInfo`.
2. After receiving `NewDecentralizedNodeInfo`, the new client sets a local flag indicating that it needs model syncing.
3. When training begins, if this flag is set, the new client sends a `ModelSyncRequest` message to the server.
4. After receiving `ModelSyncRequest`, the server sends messages as step 5 and 6, using selected model provider information from previous training round.
5. The server sends `SignalModelProvider` to the new participant with information about the provider peer.
6. The server sends `SignalNewPeer` to the provider peer with information about the newly joined peer.
7. Using this signaling information, the new participant and provider peer establish a peer connection.
8. The provider waits until the current aggregation round has finished, then sends the latest model to the new participant using a `SharedModel` message.
9. The new participant receives the model and updates its local model weights.
10. After syncing, the new participant can join subsequent decentralized training rounds.

### DIFF
--- a/discojs/src/client/decentralized/decentralized_client.ts
+++ b/discojs/src/client/decentralized/decentralized_client.ts
@@ -44,6 +44,24 @@ export class DecentralizedClient extends Client<"decentralized"> {
     // Emits the `participants` event
     this.nbOfParticipants = this.aggregator.nodes.size === 0 ? 1 : this.aggregator.nodes.size
   }
+
+  // Used by model provider peer during model syncing
+  private async handleSignalNewPeer(event: any): Promise<void> {
+    if (this.#pool === undefined) throw new Error('received signal about new peer but peer pool is undefined')
+      const roundFinishedPromise = this.#roundFinishedPromise
+      const syncConnection = await this.#pool.getPeers(Set([event.newNode]), this.server, ()=>{})
+
+      const newcomerConn = syncConnection.get(event.newNode)
+
+      if (newcomerConn === undefined){
+        // if connection with newly joining client fails, print debug message
+        // and return
+        debug(`Cannot connect to newly joined client [${event.newNode}]`)
+        return
+      }
+
+      await this.sendModel(newcomerConn, roundFinishedPromise)
+  }
   
   /**
    * Public method called by disco.ts when starting training. This method sends
@@ -80,20 +98,8 @@ export class DecentralizedClient extends Client<"decentralized"> {
     // Listen if the client is selected as a model provider node for a newly joining client.
     // Upon receiving the signal, this client establishes a connection with the newcomer
     // and sends the latest model weights.
-    this.server.on(type.SignalNewPeer, async (event) => {
-      if (this.#pool === undefined) throw new Error('received signal about new peer but peer pool is undefined')
-      const roundFinishedPromise = this.#roundFinishedPromise
-      const syncConnection = await this.#pool.getPeers(Set([event.newNode]), this.server, ()=>{})
-
-      const newcomerConn = syncConnection.get(event.newNode)
-
-      if (newcomerConn === undefined){
-        // if connection with newly joining client fails, print debug message
-        // and return
-        debug(`Cannot connect to newly joined client [${event.newNode}]`)
-        return
-      }
-      await this.sendModel(newcomerConn, roundFinishedPromise)
+    this.server.on(type.SignalNewPeer, (event) => {
+      void this.handleSignalNewPeer(event)
     })
 
     // c.f. setupServerCallbacks doc for explanation
--- a/discojs/src/client/decentralized/decentralized_client.ts
+++ b/discojs/src/client/decentralized/decentralized_client.ts
@@ -1,7 +1,7 @@
 import createDebug from "debug";
 import { Map, Set } from 'immutable'
 
-import type { DataType, Model, WeightsContainer } from "../../index.js";
+import { DataType, Model, WeightsContainer } from "../../index.js";
 import { serialization } from "../../index.js";
 import { Client,  shortenId } from '../client.js'
 import { type NodeID } from '../index.js'
@@ -26,6 +26,10 @@ export class DecentralizedClient extends Client<"decentralized"> {
   #pool?: PeerPool
   #connections?: Map<NodeID, PeerConnection>
 
+  // Store the latest model
+  // This is used when the client becomes a model provider for a newcomer
+  #latestModel?: WeightsContainer
+
   // Flag if this model requires model synchronization 
   #modelSyncNeeded?: boolean
 
@@ -45,12 +49,20 @@ export class DecentralizedClient extends Client<"decentralized"> {
     this.nbOfParticipants = this.aggregator.nodes.size === 0 ? 1 : this.aggregator.nodes.size
   }
 
+  private cloneWeights(weights: WeightsContainer): WeightsContainer {
+    return new WeightsContainer(weights.weights.map(t => t.clone()))
+  }
+
   // Used by model provider peer during model syncing
   private async handleSignalNewPeer(event: NarrowMessage<type.SignalNewPeer>): Promise<void> {
       if (this.#pool === undefined){
         throw new Error('received signal about new peer but peer pool is undefined')
       }
       const roundFinishedPromise = this.#roundFinishedPromise
+
+      // Note:
+      // getPeers() keeps the temporary connection with model provider even after model synchronization. 
+      // This connection is not added to #connections so it is not used for aggregation.
       const syncConnection = await this.#pool.getPeers(Set([event.newNode]), this.server, ()=>{})
 
       const newcomerConn = syncConnection.get(event.newNode)
@@ -174,7 +186,7 @@ export class DecentralizedClient extends Client<"decentralized"> {
       this.server.send({ type: type.ModelSyncRequest })
 
       // 2. Get the provider information from the server
-      const providerInfo = await waitMessageWithTimeout(this.server, type.SignalModelProvider, 30_000, "Timeout while waiting for the latest model provider")
+      const providerInfo = await waitMessageWithTimeout(this.server, type.SignalModelProvider, this.task.trainingInformation.maxModelSyncTime, "Timeout while waiting for the latest model provider")
       
       if (this.#pool === undefined) {
         throw new Error('peer pool is undefined, make sure to call `client.connect()` first')
@@ -193,9 +205,10 @@ export class DecentralizedClient extends Client<"decentralized"> {
       }
 
       const latestModel = await this.receiveModel(providerConn)
-      this.modelWeightAccess?.setModelWeight(latestModel)
 
-      this.emit("modelSynced", this.modelWeightAccess?.getModelWeight())
+      this.#latestModel = this.cloneWeights(latestModel)
+
+      this.emit("modelSynced", this.cloneWeights(latestModel))
       this.#modelSyncNeeded = false
     }
 
@@ -237,6 +250,9 @@ export class DecentralizedClient extends Client<"decentralized"> {
       // (3) After multiple retires, if the connection is still unsuccessful, the server starts excluding nodes from the round
       // and sends a ConnectionFail message to those nodes
       // (4) Upon receiving ConnectionFail, the client disconnects from the server
+      // TODO: Promise.race() does not close the waitMessage listeners that lost the race.
+      // Therefore, unsolved listeners may accumulate across rounds.
+      // We can add listener resolving if this becomes a problem later.
       const msg = await Promise.race([
         waitMessage(this.server, type.StartWeightSharing),
         waitMessage(this.server, type.RetryPeerConnections),
@@ -427,7 +443,7 @@ export class DecentralizedClient extends Client<"decentralized"> {
    * Receive model from the model provider.
    */
   private async receiveModel(providerConn: PeerConnection): Promise<WeightsContainer>{
-    const message = await waitMessageWithTimeout(providerConn, type.SharedModel, 30_000, "Timeout while waiting for the latest model")
+    const message = await waitMessageWithTimeout(providerConn, type.SharedModel, this.task.trainingInformation.maxModelSyncTime, "Timeout while waiting for the latest model")
     
     const decoded = serialization.weights.decode(message.model)
     return decoded
@@ -444,13 +460,13 @@ export class DecentralizedClient extends Client<"decentralized"> {
       await roundFinishedPromise
     }
 
-    const model = this.modelWeightAccess?.getModelWeight()
+    const model = this.#latestModel
 
     if (model === undefined){
       debug("Failed to get the latest model from model provider client")
       return
     }
-    const encoded = await serialization.weights.encode(model)
+    const encoded = await serialization.weights.encode(this.cloneWeights(model))
 
     const message: messages.SharedModel = {
         type: type.SharedModel,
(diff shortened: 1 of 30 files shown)


server/src/controllers/decentralized_controller.ts
```diff
@@ -98,13 +155,19 @@ export class DecentralizedController<
       // Remove the participant when the websocket is closed
       this.connections = this.connections.delete(peerId)
       this.#roundPeers = this.#roundPeers.delete(peerId)
+      this.#connectFinishedNodes = this.#connectFinishedNodes.delete(peerId)
       debug("client [%s] left", shortId)
 
+      // If this participant was a latest model provider node,
+      // replace the provider node to another node
+      if (this.#providerNode === peerId) {
+        this.#providerNode = this.connections.keySeq().first()
```
JulienVig: Maybe ensure that the new `providerNode` is not already in `#synchingNodes` here

discojs/src/client/decentralized/decentralized_client.ts
```diff
@@ -129,14 +163,50 @@ export class DecentralizedClient extends Client<"decentralized"> {
...
+      if (this.#pool === undefined) {
+        throw new Error('peer pool is undefined, make sure to call `client.connect()` first')
+      }
+
+      // 3. Connect with model provider client and get the latest model
+      const syncConnection = await this.#pool.getPeers(
+        Set([providerInfo.providerNode]),
+        this.server,
+        ()=>{}
+      )
```
JulienVig, line 203: Calling `getPeers` adds the `providerNode` to the peer pool ([this line](https://github.com/epfml/disco/blob/dfa3e45f48506bacaf3df950b4ad65860d67710f/discojs/src/client/decentralized/peer_pool.ts#L63)) which is then never removed and unused in aggregation either. 
I don't think that's a big deal as I assume this model sync won't happen that often but worth adding a comment to help debug if it becomes problematic

discojs/src/client/decentralized/decentralized_client.ts
```diff
@@ -149,11 +219,55 @@ export class DecentralizedClient extends Client<"decentralized"> {
...
+        continue
+      } else if (msg.type === type.ConnectionFail){
+        debug(`[${shortenId(this.ownId)}] disconnect from the server`)
+        await this.disconnect()
+        throw new Error("Client disconnected after connection failure")
+      }
+    }
     // Exchange weight updates with peers and return aggregated weights
-    return await this.exchangeWeightUpdates(weights)
+    const aggregatedWeight = await this.exchangeWeightUpdates(weights)
```
JulienVig, line 290: Is this scenario ever covered in the unit tests?

discojs/src/client/decentralized/messages.ts
```diff
@@ -38,6 +39,54 @@ export interface PeersForRound {
...
+
+// Server signals each peer to start weight update sharing
+export interface StartWeightSharing {
+  type: type.StartWeightSharing
+}
+
+// Server signals peers to reestablish peer connections
+export interface RetryPeerConnections {
+  type: type.RetryPeerConnections
+  aggregationRound: number
```
JulienVig: I don't think `aggregationRound` is ever used with `RetryPeerConnections`, is it on purpose?

discojs/src/client/decentralized/messages.ts
```diff
@@ -75,11 +133,17 @@ export function isMessageFromServer (o: unknown): o is MessageFromServer {
         typeof o.waitForMoreParticipants === 'boolean'
     case type.SignalForPeer:
       return 'peer' in o && isNodeID(o.peer) &&
-        'signal' in o // TODO check signal content?
+        'signal' in o
     case type.PeersForRound:
       return 'peers' in o && Array.isArray(o.peers) && o.peers.every(isNodeID)
+    case type.SignalNewPeer:
+      return 'newNode' in o && isNodeID(o.newNode)
```
JulienVig: If I'm not mistaken `SignalNewPeer` is always sent by the server to a client, if so there's no need to check its properties, you can return true directly (as you're doing for `SignalModelProvider`)

discojs/src/client/decentralized/decentralized_client.ts
```diff
@@ -36,6 +44,24 @@ export class DecentralizedClient extends Client<"decentralized"> {
     // Emits the `participants` event
     this.nbOfParticipants = this.aggregator.nodes.size === 0 ? 1 : this.aggregator.nodes.size
   }
+
+  // Used by model provider peer during model syncing
+  private async handleSignalNewPeer(event: NarrowMessage<type.SignalNewPeer>): Promise<void> {
+    if (this.#pool === undefined) throw new Error('received signal about new peer but peer pool is undefined')
```
JulienVig: There's a weird indentation in this function. Maybe put the throw on a new line to make it clearer

server/src/controllers/decentralized_controller.ts
```diff
@@ -145,9 +214,165 @@ export class DecentralizedController<
...
+        return [id, encoded] as [client.NodeID, Buffer]
+      })
+      .map(([id, encoded]) => {
+        const conn = this.connections.get(id)
+        if (conn === undefined) {
+          throw new Error(`peer ${id} marked as ready but not connection to it`)
+        }
+        return [conn, encoded] as [WebSocket, Buffer]
+      })
+      .forEach(([conn, encoded]) => {conn.send(encoded)})
```
JulienVig: This snippet looks very similar to the one below line 355, can you create a function that broadcasts a message to a set of node to avoid code duplication? Also the same logic l307 and l241 in `signalWeightSharing`

server/src/controllers/decentralized_controller.ts
```diff
@@ -145,9 +214,165 @@ export class DecentralizedController<
...
+   */
+  private handleTimeout(): void {
+    this.clearTimeout()
+    debug(`Connection setup timeout for round ${this.#aggregationRound}, Retrying with same peers`)
+    // Increment the connection retry count
+    this.#connectionRetry += 1;
+
+    // If the number of retries exceeds the threshold, exclude the failed peers from the round
+    // and retry peer connection only with the remaining peers
+    if (this.#connectionRetry >= this.task.trainingInformation.maxConnectionRetry){
```
JulienVig: With the current logic, peers are excluded after two retries if maxConnectionRetry = 3 while I would expect the logic to try 3 times and exclude on the fourth timeout.

server/src/controllers/decentralized_controller.ts
```diff
@@ -145,9 +214,165 @@ export class DecentralizedController<
...
     // empty the list of peers for the next round
     this.#roundPeers = Map()
+    this.#connectFinishedNodes = Map()
     this.#aggregationRound++
   }
+
+  /**
+   * Set a timeout to check peer connections establishment
+   */
+  private startTimeout(maxTime: number = 60_000): void {
```
JulienVig: I wonder if we should make this timeout a task parameter task like `maxConnectionRetry`, what do you think?

discojs/src/client/decentralized/decentralized_client.ts
```diff
@@ -303,4 +420,48 @@ export class DecentralizedClient extends Client<"decentralized"> {
     }
     return await this.aggregationResult
   }
+
+  /**
+   * Receive model from the model provider.
+   */
+  private async receiveModel(providerConn: PeerConnection): Promise<WeightsContainer>{
+    const message = await waitMessageWithTimeout(providerConn, type.SharedModel, 30_000, "Timeout while waiting for the latest model")
```
JulienVig: I think we should make this timeout value a task parameter, depending on the model size it can make sense to increase the timeout

discojs/src/training/disco.ts
```diff
@@ -127,9 +138,19 @@ export class Disco<D extends DataType, N extends Network> extends EventEmitter<{
...
+        return new WeightsContainer(this.trainer.model.weights.weights.map(t => t.clone()));
+      },
+      setModelWeight: (weights) => {
+        this.trainer.model.weights = weights;
+      }
+    });
     // Simply propagate the training status events emitted by the client
     this.#client.on("status", (status) => this.emit("status", status));
     this.#client.on("participants", (nbParticipants) => this.emit("participants", nbParticipants));
+    this.#client.on("modelSynced", (latestWeights) => this.emit("modelSynced", latestWeights));
```
JulienVig: Maybe you can avoid this new `ModelWeightAccess` interface (that is only useful for decentralized and needs to be created after initialization) if you do this:
```
this.#client.on("modelSynced", (latestWeights) => {
    this.trainer.model.weights = weights;
    this.emit("modelSynced", latestWeights)
});
```
And to access the model from `decentralized_client.ts`, you can store a copy of the model like `latestModel` as an attribute of decentralized client that you can set in `finishRound` by having the trainer add the model as argument l144 of `trainer.ts`:
```
this.#client.finishRound(networkWeights);
```
Let me know what you think or if my description is too vague.

discojs/src/client/decentralized/decentralized_client.ts
```diff
@@ -149,11 +219,55 @@ export class DecentralizedClient extends Client<"decentralized"> {
...
+      // (1) If the client receives a StartWeightSharing message, it proceeds to weight update exchange
+      // (2) If it receives a RetryPeerConnections message, it retries peer connection establishment
+      // (3) After multiple retires, if the connection is still unsuccessful, the server starts excluding nodes from the round
+      // and sends a ConnectionFail message to those nodes
+      // (4) Upon receiving ConnectionFail, the client disconnects from the server
+      const msg = await Promise.race([
+        waitMessage(this.server, type.StartWeightSharing),
+        waitMessage(this.server, type.RetryPeerConnections),
+        waitMessage(this.server, type.ConnectionFail),
+      ])
```
JulienVig, line 264: This creates two event listeners that are not resolved per round and accumulates throughout the rounds. I don't think it's a huge deal but can you add a comment/TODO to note this?

JulienVig (changes requested):
Great work! I have a few refactoring comments but nothing blocking. I'll approve the merge once the next demo is passed :)

server/src/controllers/decentralized_controller.ts
ahzero7d1: I updated to select `providerNode` that is not already in `#synchingNodes` by adding condition.

[corresponding line](https://github.com/epfml/disco/blob/decentralized-connection-fix/server/src/controllers/decentralized_controller.ts#L172)

discojs/src/client/decentralized/decentralized_client.ts
ahzero7d1: I added the comment for the future: [link to the line](https://github.com/epfml/disco/blob/decentralized-connection-fix/discojs/src/client/decentralized/decentralized_client.ts#L64)

discojs/src/client/decentralized/decentralized_client.ts
ahzero7d1: New unit tests are added here: [link to the unit test](https://github.com/epfml/disco/blob/decentralized-connection-fix/server/tests/e2e/decentralized_controller.spec.ts)

discojs/src/client/decentralized/messages.ts
ahzero7d1: I think I forgot to remove that part! It is now removed.

server/src/controllers/decentralized_controller.ts
ahzero7d1: Comparison logic is now modified: [link to the line](https://github.com/epfml/disco/blob/decentralized-connection-fix/server/src/controllers/decentralized_controller.ts#L327C1-L328C1)

server/src/controllers/decentralized_controller.ts
ahzero7d1: I added `maxPeerConnectionTime` parameter to training information.

discojs/src/client/decentralized/decentralized_client.ts
ahzero7d1: `maxModelSyncTime` parameter is added to training information.

[link to the line](https://github.com/epfml/disco/blob/decentralized-connection-fix/discojs/src/task/training_information.ts#L112)

discojs/src/client/decentralized/decentralized_client.ts
ahzero7d1: I added comment for this!: [link to the line](https://github.com/epfml/disco/blob/decentralized-connection-fix/discojs/src/client/decentralized/decentralized_client.ts#L253)

discojs/src/training/disco.ts
ahzero7d1: `ModelWeightAccess` interface is removed and the latest model is now sent as an argument.

- [emit the synchronized model weight](https://github.com/epfml/disco/blob/decentralized-connection-fix/discojs/src/training/disco.ts#L137)
- [finishRound method](https://github.com/epfml/disco/blob/decentralized-connection-fix/discojs/src/client/decentralized/decentralized_client.ts#L479)
- [Updating the local weight with latest model](https://github.com/epfml/disco/blob/decentralized-connection-fix/discojs/src/training/trainer.ts#L144)

server/src/controllers/decentralized_controller.ts
ahzero7d1: `broadcastMessageToNodes` method is implemented and used in the relevant parts: [link to the method](https://github.com/epfml/disco/blob/decentralized-connection-fix/server/src/controllers/decentralized_controller.ts#L213)

PR #1111, opened 2026-04-16
GPT-2 training with Goldfish loss

mina5rovic:


cli/src/evaluate_finetuned_gpt2.ts
```diff
@@ -0,0 +1,225 @@
...
+    continuation: string
+): Promise<number> {
+    const promptTokens = tokenizer.tokenize(prompt).toArray();
+    const fullTokens = tokenizer.tokenize(prompt + continuation).toArray();
+
+    const inputTokens = fullTokens.slice(0, -1);
+    const inputTensor = tf.tensor2d([inputTokens], [1, inputTokens.length], "int32");
+
+    const logits = tfModel.predict(inputTensor) as tf.Tensor;
+    const logProbs = tf.logSoftmax(logits, -1);
```
JulienVig: You're running inference on the MCQ question for each option but the question is always the same. You could run inference only once and then retrieve the logits of each option and save a lot of time. see the code snippet in the next comment

Edit: this is assuming that there is only one continuation token to evaluate, I see that there's a lot over multiple tokens lines 79-83 so maybe my comment is not valid

cli/src/evaluate_finetuned_gpt2.ts
```diff
@@ -0,0 +1,225 @@
...
+    const arr = await logProbs.array() as number[][][];
```
JulienVig: You're computing the softmax for every position but you only need the last one and line 74 materializes the whole array while you only need 4. You could rewrite this logic such that you only work with the last position, for example:
```
const optionLogProbs = tf.tidy(() => {
        const logits = tfModel.predict(inputTensor) as tf.Tensor3D; // [1, seqLen, vocab]
        const lastLogits = logits
            .slice([0, promptTokens.length - 1, 0], [1, 1, -1])      // final position only
            .reshape([-1]);                                          // [vocab]
        const logProbs = tf.logSoftmax(lastLogits);                  // [vocab]
        return tf.gather(logProbs, continuationTokenIDs);                  // continuationTokenIDs is an array of the 4 continuations' tokenID
    });
const scores = await optionLogProbs.array(); // just 4 values
```

cli/src/evaluate_finetuned_gpt2.ts
```diff
@@ -0,0 +1,225 @@
...
+        for (let i = 1; i < scores.length; i++) {
+            if (scores[i] > scores[bestIdx]) bestIdx = i;
+        }
+
+        const predicted = options[bestIdx];
+
+        if (predicted === answer) correct++;
+        total++;
+
+        if (confusion[answer]) {
```
JulienVig: You already checked that answer was contained in options line 128 so you can skip this check or only for throw an error if something unexpected happens :)

cli/src/evaluate_finetuned_gpt2.ts
```diff
@@ -0,0 +1,225 @@
...
+    const arr = await logProbs.array() as number[][][];
+
+    let score = 0;
+    let count = 0;
+
+    for (let targetPos = promptTokens.length; targetPos < fullTokens.length; targetPos++) {
+        const targetToken = fullTokens[targetPos];
+        score += arr[0][targetPos - 1][targetToken];
+        count++;
+    }
```
JulienVig: I think there should only be the last token to evaluate right? Is this loop over multiple tokens necessary?

cli/src/measure_memorization_gpt2.ts
```diff
@@ -0,0 +1,443 @@
...
+  };
+}
+
+async function main() {
+  const args = parse<Args>(
+    {
+      modelPath: { type: String, description: "Path to a saved Disco GPT model.json" },
+      dataPath: { type: String, description: "Path to records/canaries text file" },
+      maxRecords: { type: Number, description: "Maximum records to evaluate; -1 for all", defaultValue: 100 },
+      promptLengths: { type: String, description: "Comma-separated prompt lengths", defaultValue: "10,50,100,200,500" },
```
JulienVig: It's probably not useful to test prompt length 500 if the context length is 256 or 512

cli/src/measure_memorization_gpt2.ts
```diff
@@ -0,0 +1,443 @@
...
+
+    const logits = tf.tidy(() => {
+      const output = tfModel.predict(input);
+      if (Array.isArray(output)) {
+        return output[0] as tf.Tensor;
+      }
+      return output as tf.Tensor;
+    });
+
+    console.log("logits shape:", logits.shape);
```
JulienVig: Make sure to remove the debug prints before merging the PR

JulienVig (commented):
The logic of the benchmark and the memorization looks good to me! Just had some performance comments

datasets/.gitignore
```diff
@@ -7,6 +7,7 @@
 /simple_face-example.png
 /titanic*
 /mnist*
+/medicalMCQtxtNoExplanation
```
JulienVig: To remove from main PR

cli/src/args.ts
```diff
@@ -17,8 +17,18 @@ export interface BenchmarkArguments {
   numberOfUsers: number;
   epochs: number;
   roundDuration: number;
+  roundIterations?: number;
   batchSize: number;
   validationSplit: number;
+  validationFrequency?: number;
+  datasetPath?: string;
+  validationDatasetPath?: string;
```
JulienVig, line 57: Can you document how all the validation parameters interact with each others? For example what happens if you specify both validationSplit and validationDatasetPath?

discojs/src/training/trainer.ts
```diff
@@ -71,10 +87,38 @@ export class Trainer<D extends DataType, N extends Network> {
...
+      throw new Error("roundIterations must be a positive integer");
+
+    if (
+      this.#validationFrequency !== undefined &&
+      (!Number.isInteger(this.#validationFrequency) ||
+        this.#validationFrequency < 0)
+    )
+      throw new Error("validationFrequency must be a non-negative integer");
+
+    // if (!Number.isInteger(this.#epochs / this.#roundDuration))
```
JulienVig:
```suggestion
```

discojs/src/client/federated/federated_client.ts
```diff
@@ -138,28 +153,58 @@ export class FederatedClient extends Client<"federated"> {
...
-    // Need to await the resulting global model right after sending our local contribution
-    // to make sure we don't miss it
+    const round = this.aggregator.round;
+    {
+      debugProcessMemory(
+        `[${shortenId(this.ownId)}] round ${round} before encode`,
+      );
+      const payload = await serialization.weights.encode(payloadToServer);
+      debugProcessMemory(
+        `[${shortenId(this.ownId)}] round ${round} after encode`,
```
JulienVig: is it useful to log the before after?

discojs/src/client/federated/messages.ts
```diff
@@ -22,7 +22,7 @@ export interface NewFederatedNodeInfo {
   type: type.NewFederatedNodeInfo;
   id: NodeID;
   waitForMoreParticipants: boolean;
-  payload: serialization.Encoded;
+  payload?: serialization.Encoded | null;
```
JulienVig, line 24: why allow null?

discojs/src/serialization/model.ts
```diff
@@ -16,11 +20,13 @@ export async function encode(model: Model<DataType>): Promise<Encoded> {
   switch (true) {
     case model instanceof models.TFJS: {
       const serialized = await model.serialize();
+      debug("TFJS model serialized");
```
JulienVig: I think you can limit the number of debug statements in this file

discojs/src/training/disco.ts
```diff
@@ -90,6 +113,7 @@ export class Disco<D extends DataType, N extends Network> extends EventEmitter<{
   readonly #logger: Logger;
   readonly #task: Task<D, N>;
   readonly #preprocessOnce: boolean;
+  readonly #debugLabel?: string;
```
JulienVig, line 78: could you include a comment to describe this attribute?

discojs/src/training/trainer.ts
```diff
@@ -71,10 +87,38 @@ export class Trainer<D extends DataType, N extends Network> {
     this.#client = client;
     this.#roundDuration = task.trainingInformation.roundDuration;
     this.#epochs = task.trainingInformation.epochs;
+    this.#roundIterations = task.trainingInformation.roundIterations;
+    this.#validationFrequency = task.trainingInformation.validationFrequency;
+    this.#shouldValidateAfterAggregation =
```
JulienVig: Could you make it a task parameter? so we can use if we want validation before aggregation, after aggregation or both

discojs/src/training/trainer.ts
```diff
@@ -91,13 +135,18 @@ export class Trainer<D extends DataType, N extends Network> {
...
   > {
+    debug("Start train");
     if (this.#training !== undefined)
       throw new Error(
         "training already running, stop it before launching a new one",
       );
 
     try {
-      this.#training = this.#runRounds(dataset, validationDataset);
+      // this.#training = this.#runRounds(dataset, validationDataset);
```
JulienVig:
```suggestion
```

discojs/src/models/implementations/gpt/model.ts
```diff
@@ -125,39 +172,48 @@ export class GPTModel extends tf.LayersModel {
...
           const gradsClipped = clipByGlobalNormObj(grads, 1);
           this.optimizer.applyGradients(gradsClipped);
+          tf.dispose(Object.values(gradsClipped));
           return lossTensor;
         });
+        goldfishMask?.dispose();
 
         const loss = await lossTensor.array();
         averageLoss += loss;
         weightUpdateTime = performance.now() - weightUpdateTime;
```
JulienVig, line 206: Why not dispose of the tensors here? Doesn't this introduce a memory leak?

server/src/controllers/federated_controller.ts
```diff
@@ -39,12 +50,66 @@ export class FederatedController<D extends DataType> extends TrainingController<
...
+          `round ${aggregator.round} before encoding aggregate`,
+        );
+        const payload = await serialization.weights.encode(weightUpdate);
+        debugProcessMemory(
+          `round ${aggregator.round} after encoding aggregate`,
+        );
+        debug(
+          "round %o aggregate payload byteLength=%d",
+          aggregator.round,
+          payload.byteLength,
```
JulienVig: I think you can remove a lot of debug statements, this is valid for the whole PR

server/src/controllers/federated_controller.ts
```diff
@@ -17,15 +17,26 @@ import FederatedMessages = client.federated.messages;
...
+    externalGB: m.external / 1024 / 1024 / 1024,
+    arrayBuffersGB: m.arrayBuffers / 1024 / 1024 / 1024,
+  });
+}
+
 export class FederatedController<D extends DataType> extends TrainingController<
   D,
   "federated"
 > {
+  #pendingUpdateRecipients = new Map<client.NodeID, WebSocket>();
```
JulienVig: Can you add a comment to describe what this data structure is used for?

server/src/controllers/federated_controller.ts
```diff
@@ -39,12 +50,66 @@ export class FederatedController<D extends DataType> extends TrainingController<
...
+            aggregator.round,
+            recipientId.slice(0, 4),
+          );
+          recipientWs.send(encodedMsg);
+          debug(
+            "Aggregated payload sent to client [%s] for round %o",
+            recipientId.slice(0, 4),
+            aggregator.round,
+          );
+        });
```
JulienVig: What motivated you to refactor the federated controller this way? Before we would create a "once" callback on the aggregator when receiving a peer payload and now there is a unique "on" callback that loops through the peers. What does this new implementation achieve?

server/src/controllers/federated_controller.ts
```diff
@@ -39,12 +50,66 @@ export class FederatedController<D extends DataType> extends TrainingController<
...
-    this.#aggregator.on("aggregation", async (weightUpdate) => {
-      this.#latestGlobalWeights =
-        await serialization.weights.encode(weightUpdate);
+        recipients.forEach((recipientWs, recipientId) => {
+          debug(
+            "Sending global weights for round %o to client [%s]",
+            aggregator.round,
+            recipientId.slice(0, 4),
+          );
+          recipientWs.send(encodedMsg);
```
JulienVig: Maybe add a try catch here in case the send fails for one participant so that it continues sending to others

server/src/controllers/federated_controller.ts
```diff
@@ -107,30 +179,32 @@ export class FederatedController<D extends DataType> extends TrainingController<
...
+              this.#pendingUpdateRecipients.set(clientId, ws);
+              this.#aggregator.add(clientId, weights, round);
+              added = true;
+              debug(
+                `Successfully added contribution from client [%s] for round ${round}`,
+                shortId,
+              );
+            } finally {
+              weights.dispose();
+              if (!added) this.#pendingUpdateRecipients.delete(clientId);
```
JulienVig, line 205: Why the try finally without catch? we check line 181 that the contribution is valid, so adding the contribution shouldn't throw any error

server/src/controllers/federated_controller.ts
```diff
@@ -39,12 +50,66 @@ export class FederatedController<D extends DataType> extends TrainingController<
   ) {
     super(task);
     this.#latestGlobalWeights = this.initialWeights;
+  }
+
+  #makeAggregator(): aggregators.MeanAggregator {
```
JulienVig: Can you add a description of what this method does?

discojs/src/aggregator/mean.ts
```diff
@@ -35,8 +41,23 @@ export class MeanAggregator extends MultiRoundAggregator {
...
+        const next = summed.add(contribution);
+        summed.dispose();
+        summed = next;
+      }
+
+      return summed.map((weight) => weight.div(contributions.length));
+    } finally {
+      summed.dispose();
+      contributions.forEach((contribution) => contribution.dispose());
+    }
```
JulienVig: If you re-implemented the average computation to prevent memory leaks, can you use this implementation in aggregation.ts `avg`?
Also is the try-finally actually useful? I don't see how an error could occur. And if an error happens we wouldn't want to silently ignore it

discojs/src/aggregator/mean.ts
JulienVig, line 1: Not exactly related to this PR but why is the mean aggregator a subclass of MultiRoundAggregator if only does one round?

server/src/controllers/federated_controller.ts
```diff
@@ -92,7 +161,10 @@ export class FederatedController<D extends DataType> extends TrainingController<
             id: clientId,
             waitForMoreParticipants:
               this.connections.size < minNbOfParticipants,
-            payload: this.#latestGlobalWeights,
+            payload:
+              this.#aggregator.round === 0
+                ? undefined
+                : this.#latestGlobalWeights,
```
JulienVig, line 166: Why are you sending `undefined` instead of the base model when new clients are joining?

server/src/controllers/federated_controller.ts
JulienVig: If it can throw an error then it should be handled in a catch

server/src/routes/training_router.ts
```diff
@@ -55,10 +55,14 @@ export class TrainingRouter<N extends Exclude<Network, "local">> {
...
-        (await model).weights,
-      );
+      const model = await serialization.model.decode(encodedModel);
+      const weights = model.weights;
+      let encodedWeights: serialization.Encoded;
+      try {
+        encodedWeights = await serialization.weights.encode(weights);
+      } finally {
+        model[Symbol.dispose]();
+      }
```
JulienVig, line 71: Can you add a catch to handle any error?

discojs/src/training/trainer.ts
```diff
@@ -120,35 +172,85 @@ export class Trainer<D extends DataType, N extends Network> {
...
+      throw new Error("roundIterations was not set");
+
+    const totalRound =
+      this.#privacy?.differentialPrivacy === undefined
+        ? Number.MAX_SAFE_INTEGER
+        : Math.max(
+            1,
+            Math.ceil((await dataset.size()) / this.#roundIterations) *
+              this.#epochs,
+          );
```
JulienVig, line 203: What has DP have to do with the total number of rounds?

discojs/src/training/trainer.ts
```diff
@@ -91,13 +135,18 @@ export class Trainer<D extends DataType, N extends Network> {
...
         "training already running, stop it before launching a new one",
       );
 
     try {
-      this.#training = this.#runRounds(dataset, validationDataset);
+      // this.#training = this.#runRounds(dataset, validationDataset);
+      this.#training =
+        this.#roundIterations === undefined
+          ? this.#runRounds(dataset, validationDataset)
+          : this.#runIterationRounds(dataset, validationDataset);
```
JulienVig: I think these method names are not explicit enough, what about `#runRoundsByEpoch` and `#runRoundsByIteration`? Can you add documentation for each method to clearly differentiate their intended use and behaviors?

JulienVig: Same comment for runRound and runIterationRound

discojs/src/training/trainer.ts
```diff
@@ -120,35 +172,85 @@ export class Trainer<D extends DataType, N extends Network> {
...
+          );
+
+    let round = 0;
+    for (let epoch = 0; epoch < this.#epochs; epoch++) {
+      const trainingIterator = dataset[Symbol.asyncIterator]();
+      let next = await trainingIterator.next();
+      let pendingBatch: Batched<DataFormat.ModelEncoded[D]> | undefined =
+        next.done === true ? undefined : next.value;
+
+      while (pendingBatch !== undefined) {
```
JulienVig:
```suggestion
      while (next.done !== true) {
```
I think you can simplify the implementation and drop pendingBatch if you directly use next.done and next.value instead

discojs/src/training/trainer.ts
```diff
@@ -179,6 +283,107 @@ export class Trainer<D extends DataType, N extends Network> {
...
+        validationDataset as
+          | Dataset<Batched<DataFormat.ModelEncoded["text"]>>
+          | undefined,
+        setDone,
+      ),
+    );
+
+    yield gen;
+    const epochLogs = await result;
+    epochsLogs = epochsLogs.push(epochLogs);
```
JulienVig:
```suggestion
    const [gen, epochLogs] = async_iterator.split(
      model.trainNextBatches(
        datasetIterator as AsyncIterator<
          Batched<DataFormat.ModelEncoded["text"]>
        >,
        maxBatchCount,
        validationDataset as
          | Dataset<Batched<DataFormat.ModelEncoded["text"]>>
          | undefined,
        setDone,
      ),
    );

    yield gen;
    epochsLogs = epochsLogs.push(await epochLogs);
```

discojs/src/models/implementations/gpt/model.ts
```diff
@@ -184,4 +240,68 @@ export class GPTModel extends tf.LayersModel {
...
+      hash = Math.imul(hash, 0x01000193);
+      hash ^= (token >>> 24) & 0xff;
+      hash = Math.imul(hash, 0x01000193);
+      hash ^= 0xff;
+      hash = Math.imul(hash, 0x01000193);
+    }
+
+    return hash >>> 0;
+  }
 }
```
JulienVig, line 324: Can you add documentation to this method?

discojs/src/training/trainer.ts
```diff
@@ -179,6 +283,107 @@ export class Trainer<D extends DataType, N extends Network> {
...
+
+    const validation =
+      validationDataset !== undefined
+        ? await this.model.evaluate(validationDataset)
+        : undefined;
+
+    const model = this.model as unknown as IterationTrainableTextModel;
+    if (typeof model.trainNextBatches !== "function")
+      throw new Error("model does not support iteration-based training");
+
```
JulienVig: I find this part really inconvenient because it's not generic, it only addresses on particular case. Here's an idea to work around this, let me know what you think
1. Implement a new dataset method that splits the dataset after a specified number of steps and returns 1) one dataset with only the next specified number of batches and 2) the remainder of the dataset after these batches.  e.g., const [nextBatches, datasetRemainder] = dataset.subset(maxBatchCount)
2. From runIterationRounds, I think this allows you to do reuse the existing #runRound, e.g. this.#runRound(nextBatches, roundValidationDataset)
3. With this method I don't think you need to deal with all the iterators

Let me know what you think and whether this is feasible

JulienVig (commented):
1st part of my review, I'll finish at a different time, feel free to start addressing these comments

discojs/src/client/federated/messages.ts
mina5rovic: Because payload is null only at round 0, when no global weights exist yet. The server sets it to undefined, but MessagePack decodes undefined as null, so the interface must allow null. It was one of the silent error, took some time to catch.

discojs/src/client/federated/federated_client.ts
mina5rovic: It was useful when debugging, can remove it now.

discojs/src/client/federated/messages.ts
JulienVig: Shouldn't the first payload sent be the base model that the clients use to train on their local data? I saw that your PR changed this to send null on the first round but I don't see why

mina5rovic: The client already downloads model via super.connect() before opening the federated WebSocket (before round 0). On develop, the same weights are then sent again in NewFederatedNodeInfo, so my change avoids that duplicate transfer, which is costly for GPT-2. I can put it back but I'm not sure it is optimal.

discojs/src/training/trainer.ts
mina5rovic: Added it, but please check it out in the newest commit to see if it is what you mean.

discojs/src/models/implementations/gpt/model.ts
mina5rovic: you are right, I missed it, thanks

server/src/controllers/federated_controller.ts
mina5rovic: The main motivation was to avoid serializing the same aggregated weights once per client. Previously, every contribution registered a `once` callback, so all callbacks serialized the same model concurrently when aggregation completed. This caused large memory and CPU overhead for large models.

The new implementation serializes and encodes the result once, then sends the same message to every client that contributed to that round. The recipient map preserves the previous delivery behavior while also handling failed contributions and disconnected clients more safely.

server/src/controllers/federated_controller.ts
mina5rovic: Explained this above. Because the base model is sent earlier. No need to repeat communication twice

discojs/src/aggregator/mean.ts
mina5rovic: To be honest, no idea. It was refactored before. Name really is missleading

discojs/src/training/trainer.ts
mina5rovic: When using dp, we need concrete number of rounds (approximated). But when we want to go through whole dataset, we need to put it to max integer. So it ruins the computation for DP.

mina5rovic: Maybe there is a better way to handle it. But this was my fix at the time.

discojs/src/training/trainer.ts
mina5rovic: I agree that the GPT-specific cast and `trainNextBatches` are not ideal, and moving this logic out of the model would make it more generic.

My only concern with `dataset.subset(maxBatchCount)` is that `Dataset` is currently immutable and can be iterated multiple times, while returning a remainder would make it stateful. Using `take` and `drop` could also become expensive because we would keep iterating over the already consumed batches.

Maybe we can keep the iterator in the trainer, wrap the next batches in a temporary dataset, and pass it to the existing `model.train(...)`. This would remove the GPT-specific method without changing the current `Dataset` behavior. We would still need to be careful when reusing `#runRoundByEpoch`, since it trains for `roundDuration` full passes while the iteration-based version should consume each batch only once.

discojs/src/aggregator/mean.ts
mina5rovic: The try/finally remains only to make sure in any case we dispose the aggregator-owned contributions. It does not silently ignore errors, exceptions still propagate after cleanup.

server/src/controllers/federated_controller.ts
mina5rovic: `finally` is there just to make sure we do dispose for sure in any case.

discojs/src/training/trainer.ts
JulienVig: yes perfect!

martinjaggi:
1111, that's a lot of PRs, 🎉

PR #1136, opened 2026-06-05
Migrate npm publishing to Trusted Publisher instead of access tokens

JulienVig:
NPM recommends avoid access tokens to publish new packages in our CD pipeline. In addition, the max lifetime of a token is 90 days so we regularly have to renew the token manually. 
Trusted publishing directly connects our repository with the npm packages and creates short-lived tokens per publishing and also avoids any manual renewal.
[NPM documentation](https://docs.npmjs.com/trusted-publishers)

### DIFF
--- a/.github/workflows/publish-deploy.yml
+++ b/.github/workflows/publish-deploy.yml
@@ -8,6 +8,7 @@ jobs:
   publish-npm:
     runs-on: ubuntu-latest
     permissions:
+      id-token: write
       contents: read
     steps:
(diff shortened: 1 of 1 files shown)


hleepfl (approved):

hleepfl:
Since we might be deprecating this anyways, what we can do is
- Merge to test if it works
- If it doesn't revert change and leave it as is until we deprecate
- If it does, great and we'll probably deprecate later anyways

PR #1144, opened 2026-06-16
Setup knip (part 1)

hleepfl:
First part of knip setup. Install and cleanup of webapp.
Next steps are cleanup of other subprojects, then adding knip to the ci pipeline


webapp/src/components/task_creation_form/TaskCreationForm.vue
```diff
@@ -818,7 +818,7 @@ const TFJSModelSchema = {
       switch (true) {
         case fileOrSet instanceof File:
           return fileOrSet;
-        case immutable.isSet(fileOrSet): {
+        case isSet(fileOrSet): {
```
tharvik: well, as there is `Set` from std and `Set` from immutable, that's less readable than the previous version

webapp/package.json
tharvik, line 1: use major only version specifier and let the lock file take care of the specific version. that helps when bumping minor version as it only modifies the lock and I find it cleaner.

package.json
```diff
@@ -29,6 +29,7 @@
 		"eslint": "10",
 		"eslint-plugin-cypress": "6",
 		"eslint-plugin-vue": "10",
+		"knip": "^6.17.1",
```
tharvik: blabla major blabla lock blabla

```suggestion
		"knip": "6",
```

tharvik (approved):

webapp/src/components/task_creation_form/TaskCreationForm.vue
hleepfl: What about named import like `isImmutableSet` ? I really don't like the star import

tharvik: sounds good!

PR #1153, opened 2026-06-19
clean up CI

hleepfl:
Overall update of the CI workflows. I recommend reviewing by commit.
Main highlights :
1. Repetitive actions have been extracted to custom actions
2. Main CI workflow was broken up into 1 main file and 3 sub files. It now runs in order : static code analysis > build > tests, with each step requiring the previous one. It's clean but a little bit slower, as we only start the tests after building everything. I believe that the previous version was incorrect in not waiting for webapp to be built before running though.
3. Dependency graph between build jobs is simpler to read
4. TSConfig references have been fixed, which allows linting to run without having to build anything. This does add a little bit of code duplication. sad.
5. Test jobs have been split to make it clearer which are failing or flaky and allow more targeted re-runs. No gains in speed, and we do end up running the setup actions more in the end, but I believe the added clarity is worth it
6. Dataset downloads were using cache to do both the regular cache job, but also artifacts (passing data between jobs inside a workflow). Concerns are now split, and the hash key is the download script (I assumed target files are static, but either way the previous hash key did not do anything either I think)
7. Build artifacts are now transferred between builds and tests. This adds a lot of ugly code and gives only a ~10s gain on test jobs. Should we just remove it ? It will probably be more useful once we have Rust code 
8. Checks will not be run anymore on Markdown-only changes. It should not block merging but I haven't tested it.

Conclusion : CI is more readable (imo) and modular. But there are close to zero efficiency gains apart from the markdown filter thing. Thanks for reading.


.github/actions/download-datasets/action.yml
```diff
@@ -0,0 +1,16 @@
...
+  using: composite
+  steps:
+    - id: cache
+      uses: actions/cache@v5
+      with:
+        path: datasets
+        key: datasets-${{ hashFiles('datasets/populate') }} # Only invalidate cache when download script changes
+    - if: steps.cache.outputs.cache-hit != 'true'
+      run: datasets/populate
+      shell: bash
```
tharvik, line 12: no need to specify a shell, it can directly be executed
```suggestion
```

.github/actions/setup-node/action.yml
```diff
@@ -0,0 +1,15 @@
...
+      with:
+        node-version-file: .nvmrc
+    - uses: actions/cache@v5
+      with:
+        path: |
+          ~/.npm
+          ~/.cache/Cypress
+        key: npm-${{ runner.os }}-${{ hashFiles('package-lock.json') }}
+    - run: npm ci
+      shell: bash
```
tharvik, line 15: no need for a shell
```suggestion
```

.github/actions/pull-dataset-artifacts/action.yml
tharvik, line 1: why is that separated from download-datasets?

.github/actions/download-datasets/action.yml
```diff
@@ -0,0 +1,16 @@
+name: Download datasets
+runs:
+  using: composite
+  steps:
+    - id: cache
+      uses: actions/cache@v5
+      with:
+        path: datasets
+        key: datasets-${{ hashFiles('datasets/populate') }} # Only invalidate cache when download script changes
+    - if: steps.cache.outputs.cache-hit != 'true'
```
tharvik:
```suggestion
    - if: steps.cache.outputs.cache-hit == 'false'
```

.github/workflows/verify.yml
```diff
@@ -0,0 +1,30 @@
...
+  static-analysis:
+    uses: ./.github/workflows/_static-analysis.yml
+
+  build:
+    needs: [download-datasets, static-analysis]
+    uses: ./.github/workflows/_build.yml
+
+  test:
+    needs: [build]
+    uses: ./.github/workflows/_test.yml
```
tharvik, line 30: sadly, this chunk by chunk approach actually increase the CI time by ~40%. I personally don't care much for CI time but I heard some people complaining when it gets longer

.github/workflows/_build.yml
tharvik, line 1: it doesn't reuse the built artifact between build steps so derived cache might be depending on a different built version. better to reuse it or to [cache multiple paths](https://github.com/actions/upload-artifact#upload-using-multiple-paths-and-exclusions) (also allow to extract a single archive and get all the needed deps thus simplifying the download artifacts step)

.github/workflows/verify.yml
```diff
@@ -0,0 +1,30 @@
+name: verify
+on:
+  push:
+    branches: [develop]
+    paths-ignore:
+      - "**/*.md"
+  pull_request:
+    paths-ignore:
+      - "**/*.md"
```
tharvik, line 9: doesn't check formating of markdown files then?

server/tsconfig.lib.json
```diff
@@ -1,5 +1,13 @@
 {
   "extends": "../tsconfig.base.json",
   "compilerOptions": { "outDir": "dist" },
+  "references": [
+    {
+      "path": "../discojs"
+    },
+    {
+      "path": "../discojs-node"
+    }
```
tharvik, line 10:
```suggestion
    { "path": "../discojs" },
    { "path": "../discojs-node" }
```

and some other

tharvik (approved):
small comments but nothing important, thanks for the rework!

* well done for the tsconfig references, weird that it doesn't merge it though when extending
* for passing built files around, that's not so much for built speed but to ensure that what we build is what we test is what we publish

.github/actions/download-datasets/action.yml
hleepfl: Not in composite actions, I've had pipelines crash when omitting it.
The [docs](https://docs.github.com/en/actions/reference/workflows-and-actions/metadata-syntax#runsstepsshell) seem to confirm it as well

.github/actions/pull-dataset-artifacts/action.yml
hleepfl: They are two different things. `download-datasets` downloads the files from an external server (and checks the github cache beforehand) and is executed only by the first job of the pipeline. `download-dataset-artifacts` is for passing around those downloaded files between jobs (using the same actions as for `dist` folders after compiling typescript code).

Maybe I could improve naming and/or add some comments though, I can see how it could be confusing

.github/workflows/verify.yml
hleepfl: Yes I've seen that =(
Three remarks :
- I think the previous plugging was wrong, and let the webapp tests start before the webapp code was built, so it was a bit too fast for wrong reasons
- We gained about a minute of CI time by not compiling code when linting it, which partly compensates
- I think the added clarity and organization is more than worth it

That being said, I'm open to finding some way to reduce the gap, for example by letting long-running tests start earlier. Any opinions @JulienVig ?

.github/workflows/_build.yml
hleepfl: I missed the multiple paths approach ! I assume this can also be extended to downloads, which get very ugly right now

.github/workflows/verify.yml
hleepfl: Oops you're right I have to factor that in

server/tsconfig.lib.json
hleepfl: Looks like Prettier is not opinionated enough...
I'm interested in Biome but not before it supports yaml and ideally markdown

In the meantime I'll find a prettier plugin to close those gaps, I don't want to be thinking of formatting ever.

tharvik: that's alright if prettier forces you to do that, no need to plug it too much

hleepfl: My problem was that Prettier accepts both
But after trying to force either multi-line or single-line everywhere, it looks like Prettier was kinda right, they both lead to unpleasant results in one place or the other. So for now it will stay up to personal preference I guess

PR #1167, opened 2026-06-23
Migrate to pnpm

hleepfl:
Switching over from npm to pnpm. Reasons are : it's faster, uses less disk space, forces stricter definition of dependencies and makes monorepo management easier (catalog feature for example)
Again I recommend reviewing by commit

Notes
- I had to pin exact dependency versions. pnpm doesn't like major version only, and some minor version changes were breaking things anyways
- Shared dependencies which need to be the same version can be defined in `catalog`
- I had to remove the polyfill lib in webapp and polyfill `Buffer` by hand instead. The library had bugs that hadn't been fixed since like 8 minor versions.

Also I haven't been able to test npm publishing yet. I'm like 50% confident


package.json
```diff
@@ -6,34 +6,23 @@
...
+    "eslint": "10.0.3",
+    "eslint-plugin-cypress": "6.2.0",
+    "eslint-plugin-vue": "10.8.0",
+    "knip": "6.17.1",
+    "prettier": "3.6.2",
+    "typescript": "5.9.3",
+    "typescript-eslint": "8.58.0",
+    "vitest": "catalog:"
+  },
+  "packageManager": "pnpm@11.8.0+sha512.c1f5e7c4cb241c8f174b743851d82f42b802324afc8b0f116b96adb15aa06664948dde36960a3ba1079ba5b4b29dd0140135b94b5b5f5263592249d68e555f26"
```
tharvik, line 27: that's very precise, probably too much to be commited. I mean, I don't want to be force to use a specific minor version of pnpm

Dockerfile
```diff
@@ -1,36 +1,44 @@
 # TODO freeze to 22 until tfjs#8425 is merged
 FROM node:22 AS builder
 
+RUN corepack enable
```
tharvik, line 4: [corepack is to be removed soon](https://github.com/nodejs/TSC/pull/1697#issuecomment-2737093616)

DEV.md
```diff
@@ -71,24 +71,33 @@ Use it to install the version we use in DISCO.
 nvm install # it reads `.nvmrc` to select the correct version
 ```
 
-`nvm` manages your different Node.js versions while `npm` handles your different Node.js project packages within one version.
+To use pnpm, you can use corepack, which usually comes bundled with Node.js
+
+```
+corepack enable
+corepack use pnpm
+```
```
tharvik, line 79: corepack is dying

.prettierignore
```diff
@@ -1,5 +1,6 @@
 # Prettier also ignores everything from .gitignore
 package-lock.json
```
tharvik: no more need to avoid this one as it shouldn't ever appear

```suggestion
```

Dockerfile
```diff
@@ -1,36 +1,44 @@
...
+
+COPY --link --from=builder /disco/package.json /disco/pnpm*.yaml /disco/
 COPY --link --from=builder /disco/isomorphic-wrtc/package.json isomorphic-wrtc/
 COPY --link --from=builder /disco/discojs/package.json discojs/
 COPY --link --from=builder /disco/discojs-node/package.json discojs-node/
 COPY --link --from=builder /disco/server/package.json server/
-RUN npm --omit=dev ci
 
+# We need to copy the file dependency before installing
```
tharvik, line 38: `isomorphic-wrtc` is already partially copied before, consider merging the two calls

cli/package.json
```diff
@@ -4,23 +4,27 @@
   "type": "module",
   "main": "dist/cli.js",
   "scripts": {
-    "watch": "nodemon --ext ts --ignore dist --watch ../discojs-node/dist --watch ../server/dist --watch . --exec npm run",
-    "start": "npm run build && node dist/cli.js",
-    "benchmark_gpt": "npm run build && node dist/benchmark_gpt.js",
-    "train_gpt": "npm run build && node dist/train_gpt.js",
-    "hellaswag_gpt": "npm run build && node dist/hellaswag_gpt.js",
+    "watch": "nodemon --ext ts --ignore dist --watch ../discojs-node/dist --watch ../server/dist --watch . --exec pnpm run",
+    "start": "node dist/cli.js",
```
tharvik, line 8: don't want to build it before? we do it for the other calls so why not this one?

discojs-node/tsconfig.lib.json
tharvik, line 1: why specifying the root dir? it shouldn't impact anything

package.json
```diff
@@ -6,34 +6,23 @@
...
-    "webapp",
-    "onnx-converter"
-  ],
   "dependencies": {
-    "@epfml/isomorphic-wrtc": "file:isomorphic-wrtc",
-    "debug": "4",
-    "immutable": "5",
-    "vitest": "4",
-    "zod": "4"
+    "@tensorflow/tfjs-node": "catalog:"
```
tharvik, line 10: part of the catalog now
```suggestion
```

docs/examples/tsconfig.json
tharvik, line 1: historically, we didn't wanted the example to be linked to the rest of the repo. that way, the users can simply copy the directory and have a working project.

that design choice can be revised for sure, but then we can do more links btw the two code bases

tharvik (approved):

package.json
hleepfl: I'll address that along with the corepack stuff since they go together in the next PR, which already fixes the Dockerfile comments.

Dockerfile
hleepfl: Thanks for letting me know though ! As said above I'll address it in the next PR

cli/package.json
hleepfl: Server and cli were doing it, but not webapp.
Since our example commands and tutorials always specify to build before running, I aligned on webapp rather than the others and kept the documentation as is. If I realize I missed some contradictory docs I'll fix them.
If there is a reason I didn't think of for building in `start` I'm open to aligning on the other side instead

discojs-node/tsconfig.lib.json
hleepfl: Required by newer versions on typescript (next PR), but I did it in this one because building examples (which are in a nested directory) was breaking

docs/examples/tsconfig.json
hleepfl: Was it ever the case ?
It uses `server` which we do not publish through npm, and didn't declare `immutable` (which iirc needs to be a specific version to be compatible anyways) but used it instead through the root npm deps (I assume)

If we take this route, what else were you thinking of in terms of linking code bases ? I just saw that the devDependencies can be removed but idk about the rest.

Opinions @JulienVig ? I can make a later PR to try and fully decouple them

Dockerfile
hleepfl: Solved in later PR

package.json
hleepfl: Still needed for vitest import (in `testSetupImportTFJSNode.ts`)

docs/examples/tsconfig.json
JulienVig: IMO examples should be fully decoupled and could even be a separate repository. 
Now, from the previous discussions we've had about the Disco use cases, using Disco via Node.js was not really part of them so I'm not sure whether this is useful to maintain (similarly for npm packages). Once we have Burn.ai integrated and potential GPU support then maybe having more code examples will be useful.

PR #1168, opened 2026-06-25
Typescript 6

hleepfl:
builds on #1167, review the other one before

Bumps typescript to 6 with small tsconfig adjustments
Bumps typescript-eslint to 8.62, `no-unnecessary-type-assertion` became more restrictive but buggy. Autofixed except in one case in `discojs` and everywhere in `webapp`.


discojs/src/client/decentralized/peer.ts
tharvik, line 1: the casting was to make it obvious what we were writing.

discojs/src/processing/index.ts
tharvik, line 1: better to inline the various now useless renaming variables

tharvik (approved):

PR #1175, opened 2026-07-01
Small pnpm adjustments

hleepfl:
Last devops PR (until next time =)

Change the static file dependency on `isomorphic-wrtc` to a workspace dependency. Works the same way but makes the Dockerfile slightly simpler and more maintainable, and allows dropping the `syncInjectedDepsAfterScripts`.

Also add a useless `supports-color` dependency to work against a pnpm bug and allow deduplicating dependencies. Hopefully we can drop it in future pnpm versions.

Finally, drop corepack usage, as it's getting phased out.

The numerous commits is due to me messing up rebases, don't mind it.


package.json
JulienVig, line 1: Is there a place to document why we're using supports-color for future confused others?

DEV.md
```diff
@@ -71,11 +71,14 @@ Use it to install the version we use in DISCO.
 nvm install # it reads `.nvmrc` to select the correct version
 ```
 
-To use pnpm, you can use corepack, which usually comes bundled with Node.js
+To use pnpm, you can either use their install script, or install it through npm.
 
 ```
-corepack enable
-corepack use pnpm
+curl -fsSL https://get.pnpm.io/install.sh | sh -
```
JulienVig: Maybe we should simply point to their [installation instructions](https://pnpm.io/installation) as we did for nvm, this command only runs for POSIX but also doesn't work for Mac Intel

JulienVig (approved):

PR #1179, opened 2026-07-10
Clarify data not uploaded

### LINKED ISSUE #1137: Make it extra-clear to users that data is never uploaded
Users are still getting confused about whether their confidential data is being uploaded online. We should make it unambiguous and intuitive that data always stay local and is only read by the browser when training data locally. Some ideas:
- Instant feedback: The moment files are selected, show parsed metadata immediately: row counts, column names, file sizes, a preview. Anything that appears faster than an upload could possibly complete: "Read 48,210 rows locally in 0.3s."
- Live data transfer counter: "Data uploaded: 0 bytes." Then, when training rounds send gradients, the counter shows exactly what left: "Model updates sent: 2.1 MB (weights only — no raw data)."
- data stays local icon with secure icon
- suggest to users to test it by turning off their wifi when connecting the data (though it doesn't prove that data is not uploaded later during training) or checking the devtools network tab

Mock up:

Image

hleepfl:
# Changes
- Move lock icon around and make it more visible
- Remove as many visual elements as possible to give focus to the padlock
- Add a folder icon for local file selection. Probably doesn't do much ? idk
- List files when uploaded

(will fix e2e once design is finalized)

# Screenshots
Before
image
image
After
image
image

Hopefully closes #1137

### DIFF
--- a/webapp/src/components/dataset_input/FileSelection.vue
+++ b/webapp/src/components/dataset_input/FileSelection.vue
@@ -8,7 +8,7 @@
   >
     <section>
       <!-- Hide the file input field when already submitted-->
-      <div v-if="noUpload" class="flex justify-center items-center mt-1">
+      <div v-if="noUpload" class="flex justify-center items-center my-1">
         <span
           v-tippy="{
             content: 'Data always stays on your device and is never shared.',
--- a/webapp/src/components/dataset_input/FileSelection.vue
+++ b/webapp/src/components/dataset_input/FileSelection.vue
@@ -36,7 +36,7 @@
         </p>
         <label class="mb-6">
           <span
-            class="px-4 py-2 min-w-32 text-lg uppercase text-white bg-disco-cyan rounded-sm duration-200 hover:bg-white dark:hover:bg-slate-950 hover:outline-solid hover:outline-disco-cyan hover:outline-2 hover:text-disco-cyan hover:cursor-pointer"
+            class="px-4 py-2 min-w-32 text-lg capitalize text-white bg-disco-cyan font-disco rounded-full duration-200 hover:bg-transparent dark:hover:bg-transparent hover:outline-solid hover:outline-2 hover:outline-disco-cyan dark:hover:outline-disco-light-cyan hover:text-disco-cyan dark:hover:text-disco-light-cyan"
           >
             select
           </span>
--- a/webapp/src/components/dataset_input/FileSelection.vue
+++ b/webapp/src/components/dataset_input/FileSelection.vue
@@ -21,13 +21,13 @@
         <p
           class="p-4 text-lg text-disco-blue dark:text-disco-light-blue flex-wrap justify-center"
         >
-          <span>Drag and drop the {{ fileType }} or</span>
+          <span>Drop {{ fileType }} here or</span>
         </p>
         <label class="mb-6">
           <span
             class="px-4 py-2 min-w-32 text-lg uppercase text-white bg-disco-cyan rounded-sm duration-200 hover:bg-white dark:hover:bg-slate-950 hover:outline-solid hover:outline-disco-cyan hover:outline-2 hover:text-disco-cyan hover:cursor-pointer"
           >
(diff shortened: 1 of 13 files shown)


webapp/src/components/dataset_input/FileSelection.vue
```diff
@@ -8,6 +8,17 @@
   >
     <section>
       <!-- Hide the file input field when already submitted-->
+      <div v-if="noUpload" class="flex justify-center items-center my-1">
+        <span
+          v-tippy="{
+            content: 'Data always stays on your device and is never shared.',
+          }"
+          class="bg-green-200 text-green-500 rounded-full hover:cursor-pointer px-2 py-1"
```
JulienVig: I think the colors are a bit off:
- Mallo recommended [this website](https://color.review/) to check the legibility and the current combination fails to pass
- It looks too bright in the dark theme, the bg and text colors should be swapped I believe
- I find the colors in the mockups of #1137 look quite nice

webapp/src/components/dataset_input/FileSelection.vue
```diff
@@ -8,6 +8,17 @@
   >
     <section>
       <!-- Hide the file input field when already submitted-->
+      <div v-if="noUpload" class="flex justify-center items-center my-1">
+        <span
+          v-tippy="{
+            content: 'Data always stays on your device and is never shared.',
```
JulienVig: I think this tooltip doesn't add much to the button text, what do you think of the mockup text: "Your files are read locally in your browser. They are never to our servers or anyone else." ?

webapp/src/components/dataset_input/FileSelection.vue
```diff
@@ -21,13 +32,14 @@
         <p
           class="p-4 text-lg text-disco-blue dark:text-disco-light-blue flex-wrap justify-center"
```
JulienVig:
```suggestion
          class="p-4 text-lg text-disco-blue dark:text-white flex-wrap justify-center"
```
While we're on this part of the code, let's make the dark mode text white like the rest of the page. The light blue isn't very legible

webapp/src/components/dataset_input/FileSelection.vue
```diff
@@ -8,6 +8,17 @@
   >
     <section>
       <!-- Hide the file input field when already submitted-->
+      <div v-if="noUpload" class="flex justify-center items-center my-1">
```
JulienVig:
```suggestion
      <div v-if="noUpload" class="flex justify-center items-center my-2">
```
Looks a bit too close to me

JulienVig (approved):
A few suggestions but otherwise all good for me! Most important one is the bg and text color for the secure icon text.
Are we giving up the "try with wifi off" idea?

webapp/src/components/dataset_input/FileSelection.vue
hleepfl: Thanks for the link ! Colors are more sane now

webapp/src/components/dataset_input/FileSelection.vue
hleepfl: How about
"Your files are read locally in your browser, who can only read them if you manually select them first. They are never sent to our servers or anyone else\'s. To convince yourself, you can start a training session (without collaboration) while turning off your machine\'s wifi !"
?

JulienVig: I find the "manually select" part potentially confusing, what do you think of "Your files are read locally in your browser, they are never sent to our servers or anyone else's. To convince yourself, you can start a local training while turning off your machine's Wi-Fi!"

webapp/src/components/dataset_input/FileSelection.vue
JulienVig: Looks great!

webapp/src/components/dataset_input/FileSelection.vue
hleepfl: Sold

PR #1218, opened 2026-09-18
Test webapp local training in parallel

ale18V:
## What & why

This PR adds dedicated end-to-end coverage for local training tasks and runs the specs independently in CI. It also fixes trained-model ownership so the model remains available to the completion screen after the training session is cleaned up. The changes cover Titanic, LUS COVID, and Wikitext workflows, with task-specific failure screenshots and contributor documentation.

## Technical plan

The CI workflow runs the local-training specs through a separate task matrix while excluding them from the general E2E job. Shared Cypress setup and small fixtures keep each scenario focused and reproducible. The training dashboard serializes and decodes a copy of the trained model before disposing the training session, while the steps view disposes replaced and unmounted models.

## Todo

Increase local-training workloads after measuring CI timings.


webapp/src/components/training/TrainerDashboard.vue
```diff
@@ -289,8 +289,16 @@ async function startTraining(): Promise<void> {
...
     debug("while training: %o", e);
   } finally {
-    emit("model", disco.trainer.model);
-    await cleanupTrainingSession();
+    try {
+      // close() disposes the trainer's model. Give the save/evaluation screen
+      // its own copy before releasing the training session.
+      const trainedModel = await modelDecode(
+        await modelEncode(disco.trainer.model),
+      );
```
ale18V: Before the model was being emitted and the reference cleaned up immediately with cleanupTrainingSession.

So now we copy it and emit a copy. I suggest adding an explicit function `modelClone` to `discojs/src/serialization/model.ts` 

@JulienVig what do you think?

webapp/cypress/fixtures/local_training/wikitext.txt
```diff
@@ -0,0 +1 @@
+The small town had a school near the river. Every morning the children walked along the road to learn and play. Their teacher read a book about the world and asked them to write a story of their own. After class they went home and told their parents what they had learned that day. It was a good day for everyone.
```
ale18V: @JulienVig for the test runs I created those small training samples to keep CI runtime small. I think for LUS and Titanic I can try to use the whole dataset but wikitext is ~500MB so it's not feasible to use it all.

Any feedback on this?

webapp/src/components/training/TrainerDashboard.vue
JulienVig: Can you rebase your PR once #1234 is merged? It makes sense to use a shallowRef so I added it to my PR but no need to do a deep copy of the model we can just transfer the ownership

webapp/cypress/fixtures/local_training/wikitext.txt
JulienVig: For text tasks we usually use a max number of iterations rather than epochs. But training multiple GPT-2 at once will be too much anyway so let's skip wikitext for now

webapp/src/components/training/TrainerDashboard.vue
JulienVig: #1234 is merged now

Issue #1219, opened 2026-09-20
Federated learning staleness bugs


ale18V:
## Summary

This tracking issue groups confirmed defects in federated stale-contribution recovery and aggregation membership handling.

## Reproduction branch

The characterization suite is available on [`federated-learning-staleness-bugs`](https://github.com/epfml/disco/tree/federated-learning-staleness-bugs), in [`server/tests/e2e/federated_stale_reproduction.spec.ts`](https://github.com/epfml/disco/blob/federated-learning-staleness-bugs/server/tests/e2e/federated_stale_reproduction.spec.ts).

Run it from the repository root:

```bash
node_modules/.bin/vitest run --project=server server/tests/e2e/federated_stale_reproduction.spec.ts
```

The suite is a prototype that characterizes current behavior. It intentionally lives only on the reproduction branch and is not proposed for merge into `main`.

## Sub-issues

The sub-issues below track the stale-recovery round number, disconnected-contributor aggregation, and first-round recovery behavior.

PR #1223, opened 2026-09-21
fix: discard contributions from departed nodes

### LINKED ISSUE #1221: Remove departed contributors from pending federated aggregation
## Problem

When a participant disconnects after submitting, `removeNode` removes it from membership but leaves its buffered contribution. The fullness check then compares the buffered count with the reduced membership count and can aggregate before a still-connected participant submits. That participant becomes stale.

## Reproduce

Use the [reproduction branch](https://github.com/epfml/disco/tree/federated-learning-staleness-bugs):

```bash
node_modules/.bin/vitest run --project=server server/tests/e2e/federated_stale_reproduction.spec.ts -t "counts a departed contributor"
```

The scenario is: A submits and disconnects; B and C submit; D remains connected but has not submitted. The server aggregates A, B, and C, advances to round 1, and rejects D’s later round-0 contribution.

## Expected outcome

Define membership for an in-flight round and make the contribution buffer follow that policy. Under current-membership semantics, a departed participant’s buffered contribution must be removed and disposed, and aggregation must await D. Add a regression test for the selected policy.

ale18V:
## What & why

Fixes #1221 and contributes to #1219. A federated participant that disconnected after submitting remained in the pending contribution buffer, allowing its update to count in place of a still-connected participant. This change discards and disposes pending contributions when their node leaves, so aggregation waits for the remaining active participants. Regression coverage verifies both the controller flow and the aggregator buffer behavior.

## Technical plan

Keep active-node membership and buffered contributions synchronized when a node is removed. Exercise the disconnect path with the federated controller and assert that aggregation includes only the remaining contributors.

## Deviations from plan

None.

## Todo

None.


ale18V:
Closing as @leonard-amsler is taking care of it.

Issue #1222, opened 2026-09-20
Define first-round stale contribution recovery


ale18V:
## Problem

The controller initializes `#latestGlobalWeights` with the initial model. Consequently, rejecting a pre-round stale contribution returns those initial weights labeled `round: -1`. This contradicts the `docs/FEDERATED.md` diagram, which says no response is sent when no global model exists on the first round.

## Reproduce

Use the [reproduction branch](https://github.com/epfml/disco/tree/federated-learning-staleness-bugs):

```bash
node_modules/.bin/vitest run --project=server server/tests/e2e/federated_stale_reproduction.spec.ts -t "returns initial weights"
```

The test deliberately submits `round: -1` before any aggregation and observes initial weight `10` returned with `round: -1`.

## Expected outcome

Choose and document the intended first-round recovery behavior. If initial weights are returned, label them with the next valid submission round (`0`); otherwise omit the response. Update `docs/FEDERATED.md` and add a regression assertion for the selected behavior.

PR #1224, opened 2026-09-21
fix: recover stale federated clients at current round

### LINKED ISSUE #1220: Fix stale contribution recovery round number
## Problem

After the server has advanced, a stale contribution is rejected and the server returns the cached global weights with `round: serverRound - 1`. `FederatedClient` adopts that value with `aggregator.setRound`, so it remains stale and its next contribution is rejected again.

## Reproduce

Use the [reproduction branch](https://github.com/epfml/disco/tree/federated-learning-staleness-bugs):

```bash
node_modules/.bin/vitest run --project=server server/tests/e2e/federated_stale_reproduction.spec.ts -t "labels the same cached weights"
```

The test shows a successful round-0 aggregation returning round 1, followed by three stale recoveries all returning the same cached weights labeled round 0.

## Expected outcome

A stale-recovery response must identify the next round in which the client may submit, consistently with a normal aggregation response. Add a regression test showing that the next contribution after recovery is accepted.

ale18V:
## What & why

Fixes #1220 and contributes to #1219. A rejected stale contribution received cached global weights labeled with the prior round, so the client remained behind and was rejected again. This change records the next valid submission round alongside every cached global model and returns that round during recovery. The regression suite verifies that a recovered client can submit an accepted update in the next round.

## Technical plan

Capture the completed aggregation round before serializing its model, then treat the payload and round as one cached recovery state. Use the same state for normal broadcasts and stale-contribution responses.

## Deviations from plan

None.

## Todo

The first-round recovery documentation and explicit round-0 assertion are handled by the stacked draft for #1222.


ale18V:
Closing as @leonard-amsler is taking care of it.

